PR #2, opened 2020-10-31
Crate skeleton

str4d:


book/src/design.md
```diff
@@ -1 +1,18 @@
...
+byte encodings to implement them. This resulted in subtly-different consensus rules being
+deployed than were intended, for example where a particular type was not round-trip
+encodable.
+
+In Pollard, we avoid this by defining the consensus rules in terms of the byte encodings
+of all variables, and being explicit about any types that are not round-trip encodable.
+This makes consensus compatibility between strongly-typed implementations (such as this
+crate) and byte-oriented implementations easier to achieve.
+
+- TODO: Do we want to take this approach, or try again for "everything non-canonical"?
```
daira, line 18: I disagree with this approach.

In Sapling we failed to implement some of what the spec said &mdash; for example, that abst<sub>J</sub> should return ⊥ for any encoding not in the image of repr<sub>J</sub>. I don't think that the approach of defining an abstract protocol involving high-level types was the issue here.

There were three contributing factors:
* we just didn't check the implementation against the spec thoroughly enough;
* the spec included non-constructive definitions (such as the one above);
* the consensus rules weren't named in a way that could be linked to their implementations.

We can fix these problems without losing the ability to define consensus rules on high-level types.

src/bundle.rs
```diff
@@ -0,0 +1,89 @@
+use crate::{
+    circuit::Proof,
+    note::{EncryptedNote, NoteCommitment, Nullifier},
+    primitives::tweedlesig::{self, Binding, SpendAuth},
```
daira, line 4: These are binding and spend auth signatures associated with Pollard, not with Tweedle. IMHO we should try to share as much code as possible between Pollard and Sapling. So, we could define `PollardBinding`, `PollardSpendAuth`, `SaplingBinding`, etc. structs that capture the differences in parameterization, and then we'd be able to use types like `reddsa::PublicKey<PollardBinding>`, etc.

src/primitives/tweedlesig.rs
```diff
@@ -0,0 +1,36 @@
+//! TODO
+//!
+//! - If we're using the other curve, then this won't be an instantiation of RedDSA
+//! - let's use "Verification" and "Signing" key terminology
```
daira, line 4:
```suggestion
//! - let's use "Validation" and "Signing" key terminology
```

src/primitives/tweedlesig.rs
```diff
@@ -0,0 +1,36 @@
+//! TODO
+//!
+//! - If we're using the other curve, then this won't be an instantiation of RedDSA
```
daira, line 3: I'm confused. Yes it will. (RedDSA supports cofactor 1 as a special case, and it's not dependent on curve type.) We should definitely generalize the `redjubjub` crate rather than writing completely new code.

src/primitives/tweedlesig.rs
```diff
@@ -0,0 +1,36 @@
...
+impl SigType for SpendAuth {}
+
+pub enum Binding {}
+impl SigType for Binding {}
+
+pub struct SigningKey<T: SigType> {
+    _t: PhantomData<T>,
+}
+
+pub struct VerificationKey<T: SigType> {
```
daira, line 20:
```suggestion
pub struct ValidationKey<T: SigType> {
```

src/bundle.rs
str4d: We can't reuse RedDSA code unless we use an embedded Edwards curve. Are we doing that, rather than using one of the cycle curves?

src/primitives/tweedlesig.rs
str4d: Oh, is that so? Then yes, we can generalise.

book/src/design.md
nuttycom: I tend to agree here. By defining consensus rules relative to high-level types, those types represent explicit sets of values, rather than encodings of those values. I think this is *more* compatible with strongly-typed implementations than defining rules in terms of wire encodings would be. There are two "layers" of consensus, as I see it: representational consensus (in terms of the ability to parse from a particular byte representation) and semantic consensus, which is better addressed by defining directly the membership of sets of valid values (the high-level types) and then the consensus rules that relate the interaction of those types.

src/bundle.rs
daira: RedDSA itself doesn't depend on any properties of the curve, other than having a fairly small cofactor (which can be 1). The same code could be generic over any represented group, e.g. any `CofactorGroup` from the `group` crate.

str4d:
I've extracted the book changes to add Pollard design notes into a separate PR: #5.

book/src/design.md
daira: I'm now convinced that we can get the best of both worlds: explicitly define the conversions between transaction fields represented as byte sequences, and also be able to analyse the protocol in terms of high-level types. So I consider this resolved.

src/primitives/tweedlesig.rs
str4d: Hmm, that's inconsistent with pretty much every signature library I can think of. I've seen `PublicKey` (which I dislike) and (variants of) `VerificationKey`, but never `ValidatingKey`.

src/keys.rs
```diff
@@ -0,0 +1,123 @@
+//! Key structures for Orchard.
+//!
+//! TODO: Should we have the concept of a Network here? Or make these standalone without
```
str4d, line 3: After writing this comment, I ran into a situation where I needed this, so I added a `Chain` trait to represent this.

src/keys.rs
```diff
@@ -0,0 +1,123 @@
...
+//! defined string encodings, and use newtypes in Zcash?
```
nuttycom, line 4: IMO it's useful to keep serialization concerns separate.

src/lib.rs
```diff
@@ -6,3 +6,24 @@
...
+pub mod value;
+
+pub use address::Address;
+pub use note::{EncryptedNote, Note, NoteCommitment, Nullifier};
+
+/// Chain-specific constants and constraints for Orchard.
+///
+/// The purpose of this trait is to encapsulate things like the human-readable prefixes
+/// for encoded addresses, or the range of allowable values for notes.
+pub trait Chain {
```
nuttycom, line 26: I'm not 100% sold on this trait. It has the flavor of the kind of constraints that are ultimately proving relatively trivial invariants in the type system, and which fall apart at serialization boundaries. It currently adds a fair bit of noise to the crate, and just strikes me as the sort of thing I've regretted doing in the past. Will have to consider further. It particularly seems a little bit superfluous given that Rust doesn't have GADTs, so none of the static invariants that this would cause to hold can be accessed dynamically without throwing away the type safety this would give you anyway (ala https://github.com/zcash/librustzcash/blob/master/zcash_primitives/src/consensus.rs#L273).

One alternative that occurs to me: instead of defining concrete data structures that are parameterized on `C: Chain` one could instead define a set of traits that represent the capabilities that those data structures represent. Then functions that accept such values as input can be constrained with those traits, and we can get some of the same benefits without needing all the phantoms and such?

str4d:
I've removed the `Chain` and `value::Constraint` traits. My rationale for including them was to make it simpler to reason about what is responsible for enforcing chain-specific constraints, and to reduce duplication (by enabling the wrapping chain implementation to use type definitions and leverage all built-in behaviour, instead of newtypes and needing to add a bunch of wrapping logic and boilerplate, some of which would encode chain-specific logic).

Instead, we'll try working within the requirement that this crate enforces minimal base constraints and hard-codes any constants, and then have the wrapping chain provide encoding prefixes and additional value constraints where necessary. I'm wary of the complexity this will add to understanding the consensus rules (in that chains will be applying constraints that constrict / replace ones defined in this crate, rather than having the constraints defined either here or in the chain code), but maybe it will work out fine.

src/keys.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/// A key that provides the capability to view incoming and outgoing transactions.
+///
+/// This key is useful in situations where you only need the capability to detect inbound
+/// payments, such as merchant terminals.
+///
+/// This key is not suitable for use in a wallet, as it cannot maintain accurate balance.
+/// You should use a [`FullViewingKey`] instead.
+///
```
therealyingtong, line 60:
```suggestion
```

src/keys.rs
```diff
@@ -0,0 +1,123 @@
...
+/// A key that provides the capability to detect and decrypt incoming notes from the block
+/// chain, without being able to spend the notes or detect when they are spent.
+///
+/// This key is useful in situations where you only need the capability to detect inbound
+/// payments, such as merchant terminals.
+///
+/// This key is not suitable for use in a wallet, as it cannot maintain accurate balance.
+/// You should use a [`FullViewingKey`] instead.
+#[derive(Debug)]
+pub struct IncomingViewingKey<C: Chain>(PhantomData<C>);
```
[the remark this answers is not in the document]
ebfull: (retract this objection)

src/keys.rs
ebfull: Why remove this?

therealyingtong: This is repeated on line 90

str4d: This was on FVK, not IVK. Copy-paste error on my part.

PR #6, opened 2020-12-17
First pass at an explicit requirements document.

nathan-at-least:


book/src/requirements.md
```diff
@@ -0,0 +1,170 @@
...
+
+### TSR2. Differences from Sapling Design Document
+
+**Requirement:** Every observable difference from Sapling from the perspective of a Mobile Wallet SDK, a zcashd RPC developer, or any user of any known services built with those components, must be documented in a "Differences from Sapling" document in this repository. Any change that impacts a product-level requirement must exist in this document as an explicit requirement.
+
+**Rationale - Technical Strategy:** Any deviation from Sapling needs to be clearly communicated to users and developers and reviewed by protocol engineers, product owners, and ecosystem stakeholders.
+
+### TSR3. Halo 2 Deployment
+
+**Requirement:** Pollard must use Halo 2 for zeroknowledge proof validation.
```
rex4539, line 146:
```suggestion
**Requirement:** Pollard must use Halo 2 for zero knowledge proof validation.
```

nathan-at-least:
Note on CI - I reviewed these failing CI tests and read each highlighted error output:

- https://github.com/zcash/pollard/pull/6/checks?check_run_id=1645368403
- https://github.com/zcash/pollard/pull/6/checks?check_run_id=1645368750
- https://github.com/zcash/pollard/pull/6/checks?check_run_id=1645368704
- https://github.com/zcash/pollard/pull/6/checks?check_run_id=1645369069

Based on that cursory skim, I don't believe this PR is the cause of any failure. *However*, it seems that rust build errors in [Intra-doc links](https://github.com/zcash/pollard/pull/6/checks?check_run_id=1645369069) might be masking any potential doc errors in `requirements.md` of this PR, but I'm not sure.

book/src/requirements.md
```diff
@@ -0,0 +1,170 @@
+# Requirements
+
+The design of Pollard aimed to be mostly identical to the Sapling circuit in
+functionality. We used this opportunity to clarify the requirements for Pollard at
+the product and engineering levels.
+
+## Protocol Product Requirements
+
+### ProtoPR1. Pollard Addresses
```
steven-ecc, line 9: Shouldn't the first Protocol Product Requirement be "Eliminate the Need for a Trusted Setup"?

defuse:
tl;dr of my suggestions (see my internal ECC slack message for more detail):

- Add requirement that it's no longer possible to create Sprout outputs by the time pollard activates,
- Add requirement that we publicly set expectations for how and when the Pollard pool will be deprecated before Pollard activates.
- Add requirement that at least one ECC employee who has not significantly contributed to Pollard's design feels confident in their ability to attack it.
- Add requirement that if the external security audits turn up many unexpected issues, we internally re-evaluate the safety of deploying Pollard rather than assuming the auditors found all of the problems and that fixing just the problems mentioned in the audit report is good enough.
- Add requirement that there is enough wallet/core engineering bandwidth to fix the privacy problems in the light client protocol (provided that I finish suggesting what to do soon enough).

book/src/requirements.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Rationale - Engineering:** Wallet vendors who rely on the SDK can add Pollard features prior to Mainnet activation.
+
+**Rationale - UX:** Wallet vendors can generate Pollard addresses without requiring users to back up any new secrets.
+
+**Rationale - UX:** Wallet vendors can support importing compatible secrets in a new wallet which will have access to pre-Pollard-activation funds and also support Pollard addresses.
+
+### SDKPR2. Mobile Wallet SDK supports sending to Pollard addresses.
+
+**Requirement:** The ECC Mobile Wallet SDK supports sending funds to a Pollard address from a T-Address or a Pollard address upon Pollard activation.
```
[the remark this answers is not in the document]
steven-ecc: There's no new transparent address support being added. I think the wording could be clarified.

nathan-at-least:
Much more work is needed on requirements. I'm just going to begin jotting down things in comments and incorporate them later:

"we intend to retain the property with Pollard that a break of soundness (binding) will not result in a break of privacy"

nathan-at-least:
Note: I use $CODENAME rather than Pollard, because we may rename the feature.

> * Add requirement that it's no longer possible to create Sprout outputs by the time pollard activates

I believe this would be a significant complication, but it may be worth it.

A design/implementation complication: the current set of requirements keeps the changes fairly well encapsulated in a new circuit / txn field with familiar turnstile rules as a key interface. That encapsulation isn't written as an explicit requirement itself. ;-) So maybe it's okay to broaden the scope, but I'm wary of broadening the scope in any way, since I believe the key to shipping things safely and quickly is safe bite-sized chunks.

The next complication is that this _MAY_ require a circuit change for Sprout. Does any domain expert know if this is true or false? If so, that will require both circuit design work and a parameter setup ceremony.

The third complication is in the UX & ecosystem layer, but fortunately it only applies to Sprout wallets (of which I'm only aware of zcashd and zecwallet) and services: how do those products communicate the restriction to users and how does it guide them in the right way.

All of these concerns, for me, push me towards doing this change in a separate follow up upgrade. However, I think we can commit our intent to do so in the next upgrade.

This line of reasoning makes me wonder about the timeline of requirements. The current PR intends for all requirements to be met at the same time ("before activation"). Maybe it's better to make target dates for different requirements variable, then to sequence them?

This might lead to a kind of funky requirements flow where we say "$CODENAME feature has activated! But the requirements are not complete because we haven't disabled Sprout outputs in the subsequent upgrade, which is still in the future." I think that could work well once we wrap our heads around it.

> * Add requirement that we publicly set expectations for how and when the Pollard pool will be deprecated before Pollard activates.

+10. I think we should do this, but it adds a bit of complexity. I still think we should do it.

A minimally acceptable version could define a progression of phases with loose lower-bound timelines. For example, here's a brainstorm:

- Phase 1 - fully active: (≥ 12 months)
- Phase 2 - sunset: transferring funds _into_ the $CODENAME pool is disabled (≥ 18 months)
- Phase 3 - dusk: any transfer of $CODENAME funds _must_ exit the pool (≥ 24 months)
- Phase 4 - night time: like dusk, but funds also are assessed a "late penalty fee" when moved that ramps up from 0 to 100% over K months (≥ 30 months)
- Phase 5 - the witching hour: the late penalty is now 100%. Therefore the pool is deleted, nodes no longer need to remember state, and all remaining pool balance immediately transfers to the penalty fee destination (whatever that is, not specified here). (≥ 36 months)

We should also guarantee maybe a 6 month gap between any phases.

A stronger version could add _upper bounds_ (eg: the witching hour must happen in ≤ 48 months).

A difficulty in all of this is soliciting requirements feedback from the ecosystem. I'm relatively certain there will be strongly held opinions on multiple fronts here.

> * Add requirement that at least one ECC employee who has not significantly contributed to Pollard's design feels confident in their ability to attack it.

I like this a lot, but it doesn't seem like what I think of as a technical requirement. This seems more like an "organizational requirement". However, it's similar to "must be audited by external firms" style requirements, so I say we err on the side of including it.

> * Add requirement that if the external security audits turn up many unexpected issues, we internally re-evaluate the safety of deploying Pollard rather than assuming the auditors found all of the problems and that fixing just the problems mentioned in the audit report is good enough.

+1. I think there may need to be a judgement call, because sometimes auditors include minor things like "the API doc string used a different parameter name from the spec, so the code can be clearer by making the name match the spec." If we consider only "severe" bugs, maybe *any* severe bugs should trigger a fix, then a new audit by a different firm? Just a heuristic rule for how to terminate the process.

> * Add requirement that there is enough wallet/core engineering bandwidth to fix the privacy problems in the light client protocol (provided that I finish suggesting what to do soon enough).

In so far as any other $CODENAME requirements impact those privacy problems, we should definitely consider killing two birds with one stone (if the scope creep is reasonable).

OTOH, as far as those concerns are decouple-able, I am strongly biased towards minimizing scope creep for $CODENAME. If these concerns are separable and separated in this way, we should carefully review the priority of the outstanding lightwallet privacy flaws.

zebambam:
Hey Nathan, thanks for the opportunity to chime in on this. I raised the following concerns on behalf of security (@defuse and I) in our internal meeting, and now I'm making them available here for both internal and external comment:

We have two layers of substantive considerations for the PRD content and then a couple of meta-concerns around the process that I'll document afterwards for completeness.

### Strategy-level concerns

- There’s no meaningful improvement to the user experience over Sapling by comparison to the user inconvenience caused by the upgrade.

I think it comes down to understanding that any upgrade to a new shielded pool comes with significant user friction (as-is) and as such if we want the update to have wide appeal - wider than the portion of users who see the benefit in improved security fix coverage and a removal of toxic waste - then there needs to be some appreciable improvement over Sapling to make it worth their while. Having worked a bit to try to get very important security features into products in the past by making _any_ changes to the user experience, and finding that very difficult, I know from those experiences that a customer-focused organization needs to highly prize the customer experience above all else. 

I'm fine with us accepting that such users simply won't have an improved user experience, so long as we understand the limiting impact that will have on attempts to grow the userbase incrementally. If our strategy is "get all the big blocks in place and then partner with someone huge" then I'm perfectly happy with that, but some of the rationale I've seen includes "delight and foster new users" but without the features necessary to actually do that.

The second layer are mainly contributions of security requirements to Orchard, along the lines of our normal role here and within our remit for helping to regulate the security of future systems. They're further divided into functional and non-functional requirements:

### Functional Requirements

When considering the question "does Orchard included all the security features necessary to be a complete product?" in the context of both the MLP - Minimum Lovable Product, and the existing requirements you've mentioned that may or may not require protocol support, the following seem important to us:

1. We don't have to run a new MPC for any security fixes in Orchard.
2. We add a 'migrate wallet' feature for when users know or suspect that their secret key has been compromised.
3. ZIP-314 is enacted to provide an all-or-nothing download strategy for memo privacy and constant bandwidth protocol.
4. [?] Meaningful increase in anonymity set - perhaps:
    a) [?] Automatic t-to-z for mobile wallets.
    b) [?] Transaction arity obfuscation.
    c) [?] Anchor selection improvements?

### Non-functional Requirements

**_We can’t defend a lot of shielded pools._**

Add requirement that it's no longer possible to create Sprout outputs by the time pollard activates,
Add the requirement that we publicly set expectations for how and when the Pollard pool will be deprecated before Pollard activates.

**_External assessments are required._**

1. Add the requirement that if the external security audits turn up many unexpected issues, we internally re-evaluate the safety of deploying Pollard rather than assuming the auditors found all of the problems and that fixing just the problems mentioned in the audit report is good enough.
2. Relevant external security assessment(s) expertly applied by our approved security vendors.
3. External Zero-knowledge experts have assessed our proving system and circuit designs and implementations.
4. External Cryptographic implementation experts have assessed our implementation.

**_External assessments aren’t enough._**

5. Add the requirement that at least one ECC employee who has not significantly contributed to Pollard's design feels confident in their ability to attack it.
6. Add the requirement that the internal security team must have confidence in the launch.

**_A mathematical proof is required._**

7. Add the requirement that a relevant cryptographic security proof has been published.

**_Other work is important_**

8. Add the requirement that there is enough wallet/core engineering bandwidth to fix the privacy problems in the light client protocol (provided that I finish suggesting what to do soon enough).

### Process-level meta-concerns

I can see concerns around scope-creep legitimately being raised. This is important to think about robustly. If we're doing a PRD to specify _only_ the protocol, there's an implication around the features that can be added to the system as a whole. In the past we have been too focussed only on the protocol and not on the user experience. Consideration of the user's perspective is critical if we plan to incrementally grow the userbase.

One thing we can do is ask whether we are in good shape yet to deliver a new shielded pool: “We’re considering committing to this very intense build period, is there anything else that needs to be dealt with before we commit to it?” Asking that question is both more complete in its approach and isolating in terms of scope for requirements discussions. Dependencies shake out far easier if you do this, without having to have the PRD be so expansive in it's scope.

Unrealistic response deadlines - we raised this and it has now been addressed, thank you.

daira:
@zebambam wrote:
> Add requirement that it's no longer possible to create Sprout outputs by the time pollard activates

The problem with that &mdash;and the reason we didn't *already* prevent that in Canopy&mdash; is that in order to hide the total value of the inputs to a JoinSplit, it's necessary to create new Sprout notes. So when migrating from Sprout to Orchard (or Sapling), there would be greater information leakage about the Sprout note values if this restriction were enforced.

book/src/requirements.md
```diff
@@ -0,0 +1,170 @@
...
+
+### ProtoPR1. Pollard Addresses
+
+**Requirement:** Pollard introduces a new address format with an unambiguous user-facing encoding.
+
+**Rationale - UX:** A new address format allows users to identify which ZKP platform their funds rely on.
+
+**Rationale - UX:** A new address format reinforces the fact that users must migrate funds to opt into Pollard.
+
+**Rationale - UX:** A new address format, in combination with the turnstile design, signals to users a boundary for privacy-guarantees.
```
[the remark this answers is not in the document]
nathan-at-least: Agreed that both are important rationales, not sure either is more salient, so I'll add the "supply integrity" rationale.

book/src/requirements.md
nathan-at-least: Yes, agreed. I believe this is strongly implied by some of the rationales, but I agree it's a foremost requirement.

book/src/requirements.md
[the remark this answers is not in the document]
nathan-at-least: Pushed 57a1937 which adds a rationale explicitly for "supply integrity". How does that sound?

book/src/requirements.md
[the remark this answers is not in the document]
nathan-at-least: I've removed all requirements on ECC products outside of specifically core consensus protocol requirements. This concern still stands for ECC's mobile wallet sdk and demo wallet, so let's resume this topic as we work on those product requirements.

book/src/requirements.md
```diff
@@ -0,0 +1,170 @@
...
+
+### SDKPR4. Mobile Wallet SDK supports turnstile migration from Sapling to Pollard addresses.
+
+**Requirement:** The ECC Mobile Wallet SDK supports migrating funds from one or more Sapling addresses to a Pollard address.
+
+**Rationale - Engineering:** Wallet vendors who rely on the SDK can add Pollard migration support prior to Mainnet activation.
+
+### SDKPR5. Mobile Wallet SDK Pollard viewing key support.
+
+**Requirement:**  ECC Mobile Wallet SDK supports exporting and importing Pollard viewing keys, and scanning private transaction data based on those viewing keys.
```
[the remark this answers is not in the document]
nathan-at-least: Non-protocol product requirements will be tracked elsewhere, and I removed them in a0d0019. Remind me of this concern as we refine those product requirements.

nathan-at-least:
Pushed two changes:

- Addressed @nuttycom's suggestion that "supply integrity" is a rationale for separate pools / addresses / turnstiles in 57a1937.
- Non-protocol product requirements will be tracked elsewhere, and I removed them in a0d0019. Remind me of this concern as we refine those product requirements.

Still a lot to address. I will address more in separate commits or separate PR comments, and I'll try to break down giant posts into separate topical comments (I find it easier to do cross-linking this way).

nathan-at-least:
Responding to https://github.com/zcash/orchard/pull/6#issuecomment-754319048 :

> tl;dr of my suggestions (see my internal ECC slack message for more detail):
> 
> * Add requirement that it's no longer possible to create Sprout outputs by the time pollard activates,
> * Add requirement that we publicly set expectations for how and when the Pollard pool will be deprecated before Pollard activates.

Defining a consistent "pool lifecycle", moving Sprout forward through that process towards retirement, and defining how it applies to Orchard are all very valuable relevant goals. However, this is a case where we've run out of time for NU5.

**However**, I would like to see some effort on a framework soon, during NU5 Feature Selection, if possible, so that even if it's not finalized, or even if there are outstanding open issues, disagreements, or snags, the Zcash dev community would at least have a common frame of reference by Feature Selection deadline.

> * Add requirement that at least one ECC employee who has not significantly contributed to Pollard's design feels confident in their ability to attack it.

I added a scoping clarification in 95a8a59 that excludes requirements that depend on specific people, orgs, or authorities. However, this is an excellent goal for ECC itself, and even one that we could publicly commit to.

The reason for that scoping restriction is to keep the requirements + spec as a standalone internally consistent artifact that arbitrary outsiders can verify in order to maintain a protocol meta-goal that arbitrary future users can potentially verify every aspect of Zcash. (Note that trusted setup violates this meta-goal.) This discussion prompted me to define an explicit requirements in 3e605ac about design verifiability.

> * Add requirement that if the external security audits turn up many unexpected issues, we internally re-evaluate the safety of deploying Pollard rather than assuming the auditors found all of the problems and that fixing just the problems mentioned in the audit report is good enough.

This is an excellent _process_ requirement proposal, and I argue it's outside the scope of this document. This feels more like a Network Upgrade Pipeline feature. I've raised this issue in the `#network-upgrade-roadmap` channel on [the zcash dev discord](https://discord.gg/ZPMdapGK).

> * Add requirement that there is enough wallet/core engineering bandwidth to fix the privacy problems in the light client protocol (provided that I finish suggesting what to do soon enough).

This is a product-specific issue so outside of scope. Yet, it's a very important issue, and it impacts the entire shielded mobile ecosystem. I don't see us resolving it by NU5. One important question: are there protocol changes that could help? If so, now is the right time to start ramping up NU6 timeline proposals.

nathan-at-least:
From https://github.com/zcash/orchard/pull/6#issuecomment-756417244 :

> Much more work is needed on requirements. I'm just going to begin jotting down things in comments and incorporate them later:
> 
> "we intend to retain the property with Pollard that a break of soundness (binding) will not result in a break of privacy"

Resolved in cc62c35, pending review of that commit.

nathan-at-least:
From https://github.com/zcash/orchard/pull/6#issuecomment-763199188 :

> Hey Nathan, thanks for the opportunity to chime in on this. I raised the following concerns on behalf of security (@defuse and I) in our internal meeting, and now I'm making them available here for both internal and external comment:
> 
> We have two layers of substantive considerations for the PRD content and then a couple of meta-concerns around the process that I'll document afterwards for completeness.
> 
> ### Strategy-level concerns
> * There’s no meaningful improvement to the user experience over Sapling by comparison to the user inconvenience caused by the upgrade.
> 
> I think it comes down to understanding that any upgrade to a new shielded pool comes with significant user friction (as-is) and as such if we want the update to have wide appeal - wider than the portion of users who see the benefit in improved security fix coverage and a removal of toxic waste - then there needs to be some appreciable improvement over Sapling to make it worth their while. Having worked a bit to try to get very important security features into products in the past by making _any_ changes to the user experience, and finding that very difficult, I know from those experiences that a customer-focused organization needs to highly prize the customer experience above all else.

I hear this, and I've heard similar concerns elsewhere. For this requirements doc, I've aiming to reduce the scope substantially to exclude usability issues that aren't directly associated with the protocol (ex: address encodings would be in scope, yet general UX outside of scope).

Meanwhile this requirements PR seems to be the natural place for this kind of feedback to arise from the current Orchard R&D process. So **takeaway 1** is to find a better home for this kind of feedback, and **takeaway 2** is to gather this kind of feedback earlier in R&D cycles in the future.

Drilling into the specifics of this feedback: the UX under substantial active development within ECC and elsewhere. So this may affect the calculus of the trade-off, both in terms of UX benefits to upgrading and UX cost to upgrading.

In particular, I strongly believe in the future _after Zcash gains 10x or 100x adoption_ there will be needs to migrate to new address formats (for example, we see new address formats arising in Bitcoin over a decade after mainnet launch), so whatever pain there exists today for migration, we need to minimize that pain by 10x or 100x in the coming years.

> 
> I'm fine with us accepting that such users simply won't have an improved user experience, so long as we understand the limiting impact that will have on attempts to grow the userbase incrementally. If our strategy is "get all the big blocks in place and then partner with someone huge" then I'm perfectly happy with that, but some of the rationale I've seen includes "delight and foster new users" but without the features necessary to actually do that.

My main thoughts on delighting and fostering new users is that UX, while constrained to some degree by the protocol is largely independent of it. If we want to move the needle on delighting users, I believe there's a lot of opportunity beyond the protocol, and that we should (and are) focusing on that to a greater degree. Within ECC we're using Orchard as an opportunity to hone in on shielded wallet usability and make substantial low-hanging fruit improvements, many of which are not constrained or burdened by the Orchard requirements.

>
> The second layer are mainly contributions of security requirements to Orchard, along the lines of our normal role here and within our remit for helping to regulate the security of future systems. They're further divided into functional and non-functional requirements:
> 
> ### Functional Requirements
> When considering the question "does Orchard included all the security features necessary to be a complete product?" in the context of both the MLP - Minimum Lovable Product, and the existing requirements you've mentioned that may or may not require protocol support, the following seem important to us:
> 
> 1. We don't have to run a new MPC for any security fixes in Orchard.

✅ - Let me know if the new requirement and the rationale I just added under it in d5110c3 fully address this.

> 2. We add a 'migrate wallet' feature for when users know or suspect that their secret key has been compromised.

As of commit 7e74697, the scope was clarified to exclude non-core-protocol product considerations. However, this sounds like a very valuable request for ECC's Wallet team. cc @braddmiller 

> 3. ZIP-314 is enacted to provide an all-or-nothing download strategy for memo privacy and constant bandwidth protocol.

This is excluded from these Orchard-specific protocol reqs, and this again seems like an important request for @braddmiller. Since `lightwalletd` protocol is a backbone for the whole shielded mobile space, it deserves some substantial love moving forward.

> 4. [?] Meaningful increase in anonymity set - perhaps:
>    a) [?] Automatic t-to-z for mobile wallets.
>    b) [?] Transaction arity obfuscation.
>    c) [?] Anchor selection improvements?
> 

I argue these are outside the scope for this document. However, I would say all of these are good candidates for "wallet privacy/security standards" that could be ZIP-ified (and also prototyped with ECC's wallet prototype & SDKs prior to ZIP standardization). Again, though, I think the best next step is to get them on a product backlog and prioritize them independently of Orchard.

> ### Non-functional Requirements
> **_We can’t defend a lot of shielded pools._**
> 
> 1. Add requirement that it's no longer possible to create Sprout outputs by the time pollard activates,
> 2. Add the requirement that we publicly set expectations for how and when the Pollard pool will be deprecated before Pollard activates.
> 

I followed up on similar suggestions from @defuse in https://github.com/zcash/orchard/pull/6#issuecomment-789279899 . Recap: important protocol goal, arguably in-scope, but didn't get in in time, so I am following up on this for NU6 timeline.

> **_External assessments are required._**
> 
> 1. Add the requirement that if the external security audits turn up many unexpected issues, we internally re-evaluate the safety of deploying Pollard rather than assuming the auditors found all of the problems and that fixing just the problems mentioned in the audit report is good enough.

I suggested also in https://github.com/zcash/orchard/pull/6#issuecomment-789279899 this is out-of-scope here but *in scope* for Network Upgrade Pipeline improvements. I raised this issue in the community dev discord NU Roadmap channel.

Even if the wider Zcash dev community doesn't establish any specific criteria here, we can/should probably do so within ECC.

> 2. Relevant external security assessment(s) expertly applied by our approved security vendors.

AFAIK this is the established precedent for the NUP (except of course, ECC's approved vendors may differ from ZF or other Zcash dev participants).

> 3. External Zero-knowledge experts have assessed our proving system and circuit designs and implementations.

I would argue this should be standard / precedent for NUP and also we should document this with the broader Zcash community. cc @steven-ecc - I know you're working to get NUP timelines up at a shared resource. Would that spot accommodate documenting/tracking more criteria like the above quote?

> 4. External Cryptographic implementation experts have assessed our implementation.
> 

Same as last response, basically. This is NUP table-stakes, and I want a public Zcash-community-dev-wide resource to document these process acceptance criteria.

> **_External assessments aren’t enough._**
> 
> 1. Add the requirement that at least one ECC employee who has not significantly contributed to Pollard's design feels confident in their ability to attack it.

This is an excellent goal for ECC, though I've excluded it from the scope of this reqs doc as of commit 7e74697 because the reqs in this repo need to be "people independent" claims about the spec that it lives next to, whereas this is a people/org-dependent safety check. @steven-ecc do we have good tracking for internal "acceptance criteria" that's distinct from the spec reqs and also independent of the community-wide NUP criteria?

Or arguably, this could live in NUP criteria and either say "ECC has these extra criteria" or a stronger "each node implementer team meets this criteria" or something like that.

> 2. Add the requirement that the internal security team must have confidence in the launch.
> 

This is an ECC-internal requirement, and this should be a blocking input to the Trademark Agreement signoff (and probably a blocking input to an earlier technical safety signoff). cc @steven-ecc 

> **_A mathematical proof is required._**
> 
> 1. Add the requirement that a relevant cryptographic security proof has been published.
> 

I nominate this again as a NUP criteria. TODO: I find a home for NUP criteria (including proposed drafts) and return to this comment and pull in all of these valuable proposals.

> **_Other work is important_**
> 
> 1. Add the requirement that there is enough wallet/core engineering bandwidth to fix the privacy problems in the light client protocol (provided that I finish suggesting what to do soon enough).
> 

Repeated quote and my response is in https://github.com/zcash/orchard/pull/6#issuecomment-789279899 .

> ### Process-level meta-concerns
> I can see concerns around scope-creep legitimately being raised. This is important to think about robustly. If we're doing a PRD to specify _only_ the protocol, there's an implication around the features that can be added to the system as a whole. In the past we have been too focussed only on the protocol and not on the user experience. Consideration of the user's perspective is critical if we plan to incrementally grow the userbase.

Yes! This is something ECC needs to improve on and this is also why the first pass at these reqs and these PR comments have such a wide scope. After that early pass, I worked within ECC (with much credit to others!) to improve our UX and product-driven approach to this. That is ramping up well internally, and that is _why_ I'm able to return to these reqs and cut down the scope, because we can lob a lot of these product & UX-level questions over the fence successfully now.

It will take a bit of time for some of those efforts to reach the public, and unfortunately I think some of it won't be ready before Feature Selection target date of April 1. Let's check back in mid-April to see if we've done a better job of grappling with complexity across the board, and also check how well that ties into Zcash community developers' perspectives. 

> 
> One thing we can do is ask whether we are in good shape yet to deliver a new shielded pool: “We’re considering committing to this very intense build period, is there anything else that needs to be dealt with before we commit to it?” Asking that question is both more complete in its approach and isolating in terms of scope for requirements discussions. Dependencies shake out far easier if you do this, without having to have the PRD be so expansive in it's scope.
> 

Based on this new delineation within ECC between wider product goals versus tighter protocol requirements/spec, and the public progress on the protocol development, how well do you think we've addressed that question for Orchard on the NU5 timeline?

How can we do this moving forward for the post-NU5 product-level goals?

> Unrealistic response deadlines - we raised this and it has now been addressed, thank you.

Noted. Now it's time for another pass. Ideally we can lock down this tighter scope of requirements this week or next, so  please review the recent commits I've pushed, and then let me know how well all of your requests have been addressed. Thanks for crucial feedback!

nathan-at-least:
Responding to https://github.com/zcash/orchard/pull/6#issuecomment-763295843 :

General comment: both issues covered are about Pool Lifecycles, which we've split of into a product & protocol goal _after_ NU5, and therefore the issues are out-of-scope for the requirements doc in this PR.

> @zebambam wrote:
> 
> > Add requirement that it's no longer possible to create Sprout outputs by the time pollard activates
> 
> The problem with that —and the reason we didn't _already_ prevent that in Canopy— is that in order to hide the total value of the inputs to a JoinSplit, it's necessary to create new Sprout notes as change or as intermediates. So when migrating from Sprout to Orchard (or Sapling), there would be greater information leakage about the Sprout note values if this restriction were enforced.

While disabling Sprout output creation prevents better-privacy migration, doing so may still be a useful stage later in the Sprout lifecycle. Let's document the possible protocol-level changes like this and then select among those possibilities with a product-lens in a separate Pool Lifecycle design doc.

> 
> @nathan-at-least wrote:
> 
> > The next complication is that this MAY require a circuit change for Sprout. Does any domain expert know if this is true or false? If so, that will require both circuit design work and a parameter setup ceremony.
> 
> There's no possibility of us doing a Sprout circuit change for NU5. As it happens we could do it using Halo 2 without a setup, because the Sprout circuit doesn't depend on a particular finite field. But it would still be a large circuit (~216 rows per input) for which Halo 2 is suboptimal, and it would be an unreasonable amount of work and security assurance cost for little gain, IMHO.

Noted. So while this kind of change would have been a plausible goal to include in NU5, it was excluded for time/complexity reasons. Let's ensure we follow up on it soon, especially because Zcash Foundation has explicitly requested that the Zcash community prioritize it.

I just filed https://github.com/zcash/zips/issues/452 as a spot to document these kinds of discussions.

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+
+**Rationale - UX:** A separate shielded pool clarifies which funds rely on which underlying ZKP/cryptographic platform.
+
+**Rationale - UX:** A separate shielded pool associated with a new address type makes it easier to reason about migration.
+
+### ProtoPR4. Orchard Privacy Protections are Independent of Soundness.
+
+**Requirement:** Orchard's privacy protections are independent of proving system soundness so that a soundness compromise does not compromise any privacy protections.
+
+**Rationale - UX:** Users can rely on the full privacy protections of Orchard even in the face of an Orchard soundness compromise, such as a successful countereiting attack.
```
defuse, line 53: typo: countereiting

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+
+These requirements are scoped to the protocol design itself, which excludes important broader considerations for Orchard as a whole, such as usability goals for a wallet which aren't directly constrained by the protocol. Additionally, an ideal criteria for these requirements is that any outside reviewer with sufficient domain expertise should be able to verify that the existing protocol specification meets the requirements in this document. This excludes requirements based on specific authorities / orgs / people.
+
+Finally, there is a category of proposed requirements that could have been coherently in-scope for Orchard which were excluded due to time & complexity constraints.
+
+These scope clarifications exlude some draft requirements found in earlier revisions of this document, in pull request comments, or elsewhere. Many of those could become ZIPs in their own right.
+
+## Protocol Product Requirements
+
+### ProtoPR1. Orchard Relies on a Zeroknowledge Proving System which is Secure in The Uniform Random String Model
```
daira, line 15:
```suggestion
### ProtoPR1. Orchard Relies on a Zero-Knowledge Proving System that is Secure in The Uniform Random String Model
```

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+
+**Requirement:** The Orchard shielded transfer circuit uses a Zeroknowledge Proving System which is secure in the Uniform Random String (URS) Model, and therefore has no reliance on a trusted setup.
```
daira, line 17:
```suggestion
**Requirement:** The Orchard shielded transfer circuit uses a Zero-Knowledge Proving System that is secure in the Uniform Random String (URS) Model, and therefore has no reliance on a trusted setup.
```

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+
+**Rationale - UX:** Users who rely on Orchard are not directly vulnerable to the class of trusted parameter compromises.
+
+**Rationale - Strategic:** As Sprout and Sapling usage declines or are deprecated, the ZEC a whole will become protected from any previous trusted setup compromise henceforth.
+
+**Rationale - Strategic:** By avoiding a requirement on a trusted setup, Orchard *and future* upgrades no longer need a parameter setup MPC process, which improves protocol agility.
```
daira, line 23:
```suggestion
**Rationale - Strategic:** By avoiding a requirement on a trusted setup, Orchard *and future* upgrades no longer need a parameter setup MPC process, which improves protocol agility and the ability to respond quickly to security vulnerabilities.
```

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+
+### ProtoPR2. Orchard Addresses
+
+**Requirement:** Orchard introduces a new address format with an unambiguous user-facing encoding.
+
+**Rationale - UX:** A new address format allows users to identify which ZKP platform their funds rely on.
```
daira, line 29: Is this consistent with the proposal to switch to mainly relying on address bundles?

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+
+**Requirement:** A new shielded pool is introduced with Orchard. ZEC may only enter the Orchard pool from t-addrs or shielded migration transactions, following the same "turnstile" design as for Sprout and Sapling.
+
+**Rationale - UX:** A separate shielded pool clarifies which funds rely on which underlying ZKP/cryptographic platform.
+
+**Rationale - UX:** A separate shielded pool associated with a new address type makes it easier to reason about migration.
+
+### ProtoPR4. Orchard Privacy Protections are Independent of Soundness.
+
+**Requirement:** Orchard's privacy protections are independent of proving system soundness so that a soundness compromise does not compromise any privacy protections.
```
daira, line 51:
```suggestion
**Requirement:** Orchard's privacy protections are independent of proving system soundness (in the sense that the security arguments are disentangled from each other and can be separately verified) so that a soundness compromise does not compromise any privacy protections.
```

book/src/requirements.md
daira:
```suggestion
**Rationale - UX:** Users can rely on the full privacy protections of Orchard even in the face of an Orchard soundness compromise, such as a successful counterfeiting attack.
```

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+
+**Rationale - UX:** Users can rely on the full privacy protections of Orchard even in the face of an Orchard soundness compromise, such as a successful countereiting attack.
+
+## Technical Strategy Requirements
+
+### TSR1. Sapling-equivalent Functionality
+
+**Requirement:** Unless other product or engineering concerns dictate, the functionality of
+Orchard should match that of the Sapling circuit.
```
daira, line 60:
```suggestion
Orchard should match that of the Sapling shielded protocol.
```

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+
+**Rationale - UX:** By being functional equivalent to Sapling, the existing Sapling
+userbase can immediately carry over most conceptual knowledge and UX habits from
```
daira, line 63:
```suggestion
user base can immediately carry over most conceptual knowledge and UX habits from
```

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+Sapling to Orchard.
+
+**Rationale - Technical Strategy:** Orchard is the first step in removing the toxic waste
```
daira, line 66:
```suggestion
**Rationale - Technical Strategy:** Orchard is the first step in removing the trusted setup
```

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+
+### TSR2. Differences from Sapling Design Document
+
+**Requirement:** Every observable difference from Sapling from the perspective of a Mobile Wallet SDK, a zcashd RPC developer, or any user of any known services built with those components, must be documented in a "Differences from Sapling" document in this repository. Any change that impacts a product-level requirement must exist in this document as an explicit requirement.
+
+**Rationale - Technical Strategy:** Any deviation from Sapling needs to be clearly communicated to users and developers and reviewed by protocol engineers, product owners, and ecosystem stakeholders.
+
+### TSR3. Halo 2 Deployment
+
+**Requirement:** Orchard must use Halo 2 for zero knowledge proof validation.
```
daira, line 78:
```suggestion
**Requirement:** Orchard must use Halo 2 for zero-knowledge proof validation.
```

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+
+**Rationale - Technical Strategy:** By selecting Halo 2 as the ZKP platform for Orchard, the Orchard deployment provides a forcing function to productionize Halo 2.
```
daira, line 80:
```suggestion
**Rationale - Technical Strategy:** By selecting Halo 2 as the ZKP platform for Orchard, the Orchard deployment provides a forcing function to productionize Halo 2, thus unlocking potential future applications of recursion/proof-carrying data while minimising short-term risk.
```

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+### TSR5. Full Design Verifiability
+
+**Requirement:** Any arbitrary future reviewer should be able to verify the entire design and public state of the Zcash blockchain without relying on external authorities. Note that the Trusted Setup ceremonies for Sprout and Sapling violate this goal, and while Orchard activation itself does not achieve the goal, it can be achieved if Sprout and Sapling pools are eventually retired.
+
+**Rationale - Technical Strategy:** A permissionless system requires the possibility of full-design validation by any arbitrary participant, and the need for this grows as the technology grows in importance.
+
+## Non-Requirements
+
+This section documents potential requirements that are explicitly not required, and their implementation is up to the product and engineering teams best judgement.
+
```
daira, line 97:
```suggestion

### NonR1. Removing Sprout Functionality from the Consensus Protocol.

**Non-Requirement:** There will be no change in what is possible with Sprout funds at the consensus layer. (This leaves open the question of whether it will be possible to migrate funds directly from Sprout to Orchard within a transaction.)

**Rationale - Engineering:** This reduces the scope of consensus changes associated with NU5. Similar effects can be obtained by non-consensus restrictions (or just omission of functionality) in wallets. Consensus-level restrictions would have required changes in the migration design of [ZIP 308](https://zips.z.cash/zip-0308).

```

book/src/requirements.md
```diff
@@ -0,0 +1,102 @@
...
+### NonR2. User-Defined Asset Precursor Support
+
+**Non-Requirement:** The protocol does not require precursor support for a future User-Defined Assets feature.
+
+**Rationale - Technical Strategy:** Getting precursor support right requires certainty about a subset of UDA requirements, and blocking Orchard on clarifying future UDA requirements introduces more deployment & execution risk.
```
daira, line 102:
```suggestion
**Rationale - Technical Strategy:** Getting precursor support right requires certainty about a subset of UDA requirements, and blocking Orchard on clarifying future UDA requirements introduces more deployment & execution risk.

**Rationale - Engineering:** It appears to be possible to add UDAs in any case without explicit precursor support in this upgrade.
```

daira (changes requested):
Reviewed with suggested changes.

book/src/requirements.md
nathan-at-least: Great point. The address bundle may be ambiguous depending on some trade-offs.

For example if it's intended to be backwards compatible with t-addr wallets and conform to [ZIP-321](https://zips.z.cash/zip-0321) then maybe it will be ambiguous to users.

In fact, the word "unambiguous" is ambiguous: does it mean "users don't confuse Orchard addresses for other kinds of addresses" or does it mean "the user facing encoding of Orchard addresses is unambiguously specified (so that all conforming implementations use the same encoding)". When I wrote this, I meant the former.

I'm going to leave this as a pending comment and circle around with people on Address Bundle clarity.

book/src/requirements.md
nathan-at-least: I've added a requirement to spell this out. Please see the `ProtoPR1` in revision 92ed81b.

zebambam:
@nathan-at-least:

Remaining things from the security review that haven't been discussed after above but also haven't made it into the PR:

1) Add the requirement that a relevant cryptographic security proof has been published.

2) Add the requirement that if the external security audits turn up many unexpected issues, we internally re-evaluate the safety of deploying Pollard rather than assuming the auditors found all of the problems and that fixing just the problems mentioned in the audit report is good enough.

3) Relevant external security assessment(s) expertly applied by our approved security vendors.

4) External Zero-knowledge experts have assessed our proving system and circuit designs and implementations.

5) External Cryptographic implementation experts have assessed our implementation.

Adding these should be trivial since we're already doing them and on schedule, so we're just documenting what's actually happening. This is uncontentious.

Then the one thing that has been discussed further, but I disagree with the conclusion:

- Add the requirement that we publicly set expectations for how and when the Pollard pool will be deprecated before Pollard activates.

The argument about raising a ZIP and following that completely doesn't make sense to me, because we could simply say in a blog post that we intend at some point in the next ten years to deprecate the Orchard shielded pool in favor of a new Halo pool that has better still technology.

_Users need to know that before they start transferring their funds into this pool, according to the principle of least surprise_. Users should not experience a loss in service that they were not expecting. It doesn't have to be our final answer. It doesn't have to be complete. It doesn't have to be accurate. It just has to set the expectation of the impermanence of this pool.

zebambam:
@nathan-at-least I recommend creating a new section called "Non-functional requirements" and just shoveling the above in verbatim. I can raise a PR if it helps.

daira:
Most of my comments have been addressed. A note is needed about the potential conflict of ProtoPR2 with address bundles. I agree that points [1-5 mentioned by @bambam](https://github.com/zcash/orchard/pull/6#issuecomment-791793131) should be added.

I don't particularly see the need to say anything about eventual deprecation of the Orchard pool. We have to be *able* to deprecate the Orchard pool, but it's completely obvious that we can do that. Any pool can be deprecated, and there's nothing we could possibly do in the Orchard design that would prevent it. Users already know, by the example of Sprout, that we eventually deprecate pools. OTOH, I don't object to saying so explicitly.

zebambam:
> but it's completely obvious that we can do that

It's completely obvious to you, and to me, and to anyone else in ECC maybe. But I'm saying that we should say it _to all users_. If it's completely obvious to all our existing users, then we should add this and have a goal to acquire so many more new user that it's not completely obvious to that we need to say it to them explicitly. And then we should say it to them.

nathan-at-least:
Ok, I pushed a pass at incorporating https://github.com/zcash/orchard/pull/6#issuecomment-791793131 except I've tweaked the assessment requirements a bit. Does it still meet @zebambam 's goals?

I had a discussion with @zebambam and @defuse that helped shift my thinking on the purpose of these requirements and the value of keeping many non-functional requirements in scope.

One notable change: instead of requiring that Orchard pool retirement "how and when" are clarified up front, I changed it to just clarifying the possibility and intent to do so (because the "how and when" aren't yet sorted).

@daira I am persuaded by @zebambam and @defuse that at least clarifying this possibility and intent is so important to user safety (due to expectations) that it deserves to be a non-functional requirement.

steven-ecc (approved):
utACK

nuttycom (changes requested):

steven-ecc:
@zebambam Just wondering if you could check @nathan-at-least's changes to see if they address your comments?

book/src/requirements.md
```diff
@@ -0,0 +1,166 @@
...
+
+**Rationale - Strategic:** As Sprout and Sapling usage declines or are deprecated, the ZEC monetary base as a whole will become protected from any previous trusted setup compromise henceforth.
+
+**Rationale - Strategic:** By avoiding a requirement on a trusted setup, Orchard *and future* upgrades no longer need a parameter setup multiparty computation (MPC) process, which improves protocol agility and the ability to respond quickly to security vulnerabilities.
+
+### ProtoR2. Orchard Addresses
+
+**Requirement:** Orchard introduces a new address format with an unambiguous user-facing encoding.
+
+**Rationale - UX:** A new address format allows users to identify which zero-knowledge proof (ZKP) platform their funds rely on.
```
daira, line 29: The design in the protocol spec doesn't have this property. A user can only tell (by the `u1` prefix) that this is a unified address, not which pool will be used. Arguably, that isn't a property of addresses anyway: a user's funds depend primarily on the pool that their wallet holds them in.

book/src/requirements.md
```diff
@@ -0,0 +1,166 @@
...
+
+### ProtoR2. Orchard Addresses
+
+**Requirement:** Orchard introduces a new address format with an unambiguous user-facing encoding.
+
+**Rationale - UX:** A new address format allows users to identify which ZKP platform their funds rely on.
+
+**Rationale - UX:** A new address format reinforces the fact that users must migrate funds to opt into Orchard.
+
+**Rationale - UX:** A new address format, in combination with the turnstile design, signals to users a boundary for turnstile-based supply integrity protections.
```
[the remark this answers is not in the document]
daira: The current design signals this to users once, when they switch to Orchard addresses. It doesn't signal any distinction among pools from Orchard onward. I don't think this is a problem, but it shouldn't be part of the rationale.

book/src/requirements.md
```diff
@@ -0,0 +1,166 @@
...
+
+**Requirement:** Orchard introduces a new address format with an unambiguous user-facing encoding.
+
+**Rationale - UX:** A new address format allows users to identify which zero-knowledge proof (ZKP) platform their funds rely on.
+
+**Rationale - UX:** A new address format reinforces the fact that users must migrate funds to opt into Orchard.
+
+**Rationale - UX:** A new address format, in combination with the turnstile design, signals to users a boundary for turnstile-based supply integrity protections.
+
+**Rationale - UX:** A new address format, in combination with the turnstile design, signals to users a boundary for privacy-guarantees.
```
daira, line 35: Ditto. Addresses do not signal any distinction among pools from Orchard onward, for either supply integrity or privacy.

book/src/requirements.md
```diff
@@ -0,0 +1,166 @@
...
+
+**Rationale - UX:** A new address format allows users to identify which ZKP platform their funds rely on.
+
+**Rationale - UX:** A new address format reinforces the fact that users must migrate funds to opt into Orchard.
+
+**Rationale - UX:** A new address format, in combination with the turnstile design, signals to users a boundary for turnstile-based supply integrity protections.
+
+**Rationale - UX:** A new address format, in combination with the turnstile design, signals to users a boundary for privacy-guarantees.
+
+**Counter-rationale - UX:** This complicates the simplified narrative of "z→z has strong privacy" because migrating between pools has a privacy cost.
```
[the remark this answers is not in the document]
daira: There shouldn't be any reference to [ZIP 308](https://zips.z.cash/zip-0308) here, if that's what you meant @nuttycom, because we don't intend to use it for the Sapling -> Orchard migration.

book/src/requirements.md
```diff
@@ -0,0 +1,166 @@
...
+
+**Rationale - Engineering:** A new address format simplifies cryptographic design and performance.
```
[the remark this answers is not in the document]
daira: The design of Orchard requires switching to the Pasta cycle, which precludes sticking with Jubjub-based addresses. The overhead in circuit size of also supporting Sapling addresses in Pasta-based proofs would be at least 100x.

book/src/requirements.md
```diff
@@ -0,0 +1,166 @@
...
+
+### ProtoR3. An Orchard Shielded Pool and Turnstile
+
+**Requirement:** A new shielded pool is introduced with Orchard. ZEC may only enter the Orchard pool from t-addrs or shielded migration transactions, following the same "turnstile" design as for Sprout and Sapling.
+
+**Rationale - UX:** A separate shielded pool clarifies which funds rely on which underlying ZKP/cryptographic platform.
+
+**Rationale - UX:** A separate shielded pool associated with a new address type makes it easier to reason about migration.
```
daira, line 47: That isn't clear. Suppose for the sake of argument that we introduced a new pool after Orchard, for some reason such as better performance or to mitigate a cryptographic flaw. Suppose further that this didn't require a change to the abstract form of an address (i.e. (d, pk<sub>d</sub>)). What's the argument, in more detail, for how changing the address type would make it easier to reason about migration? I don't see it. (Remember that address types are hidden by the unified address abstraction anyway; they aren't visible to users.)

daira: Oh, maybe you mean that *if* you have a new address type, then it's easier to reason about migration when there is a new shielded pool. But that isn't clear either! We could add an address type that supports detection keys, for example, and there would be no reason to create a new pool for it.

book/src/requirements.md
```diff
@@ -0,0 +1,166 @@
...
+
+### ProtoR4. Orchard Privacy Protections are Independent of Soundness.
+
+**Requirement:** Orchard's privacy protections are independent of proving system soundness (in the sense that the security arguments are disentangled from each other and can be separately verified) so that a soundness compromise does not compromise any privacy protections.
+
+**Rationale - UX:** Users can reasonably rely on the full privacy protections of Orchard even in the face of an Orchard soundness compromise, such as a successful counterfeiting attack.
+
+## Non-functional Requirements
+
+Non-functional requirements may not be verifiable from the protocol specification itself, but never-the-less define the goals, processes, or assessments used in delivering the Orchard protocol.
```
daira, line 57:
```suggestion
Non-functional requirements may not be verifiable from the protocol specification itself, but nevertheless define the goals, processes, or assessments used in delivering the Orchard protocol.
```

book/src/requirements.md
```diff
@@ -0,0 +1,166 @@
...
+### Technical Strategy Requirements
+
+These requirements define and rationalize some high-level goal which may not be apparent from the design specification.
+
+#### TSR1. Sapling-equivalent Functionality
+
+**Requirement:** Unless other product or engineering concerns dictate, the functionality of
+Orchard should match that of the Sapling shielded protocol.
+
+**Rationale - UX:** By being functional equivalent to Sapling, the existing Sapling
```
daira, line 68:
```suggestion
**Rationale - UX:** By being functionally equivalent to Sapling, the existing Sapling
```

zebambam:
@steven-ecc taking a look now.

book/src/requirements.md
```diff
@@ -0,0 +1,166 @@
...
+
+#### TSR3. Halo 2 Deployment
+
+**Requirement:** Orchard must use Halo 2 for zero-knowledge proof validation.
+
+**Rationale - Technical Strategy:** By selecting Halo 2 as the ZKP platform for Orchard, the Orchard deployment provides a forcing function to productionize Halo 2, thus unlocking potential future applications of recursion/proof-carrying data while limiting short-term risk.
+
+#### TSR4. User Opt-in upon Activation
+
+**Requirement:** Users may individually shield funds or migrate into the Orchard Pool as soon as the feature activates.
```
defuse, line 90: To me, "opt-in" (from the title of the issue) means "the user can decide to keep using the old thing, and has to explicitly decide to switch to the new thing, it's not forced on them" but I believe wallets may start auto-sheilding to Orchard. Does TSR4 mean wallets must make this opt-in, i.e. the wallet will keep working exactly as before until the user clicks a button to switch over to Orchard? I think some clarification is needed.

defuse (changes requested):

book/src/requirements.md
defuse: Perhaps the title should be "Availability to users upon activation"

defuse:
ACK modulo my comment about "opt-in"

zebambam:
LGTM, thanks for the changes @nathan-at-least

zebambam (approved):
LGTM thanks for the changes

daira:
Some of this rationale went into ZIP 224, because it belonged as rationale for that ZIP in particular. But it's unclear what to do with this PR. We don't want to have two sources of truth for what the Orchard and/or NU5 requirements are (especially two different versions of the same wording).

nathan-at-least:
These are likely outdated and diverge from the actual implementation. :-(

PR #10, opened 2021-01-08
book: Add nullifier rationale

str4d:
This reverts to the previous nullifier design (which we've determined is actually fine).


daira (changes requested):
Added my suggested changes at https://github.com/zcash/orchard/pull/13

PR #11, opened 2021-01-09
book: New nullifier design

str4d:
While reviewing my local changes for #10, @daira and I broke the previous design 😄

I'm still working on writing up the nullifier rationale in #10 (a large part of which now needs rewriting :sigh:). In the meantime, this PR:
- Updates the Faerie Resistance assumptions for the alternatives.
- Switches the book to specify the new nullifier design (selected from the alternatives).
- Documents the main reason for rejecting the other alternatives.


str4d:
Hmm, @daira has looked at the previous design more closely, and it appears that it is not actually broken. I'll review the reasoning next week, and probably open a PR to revert to the previous design.

PR #13, opened 2021-01-21
Nullifier rationale

daira:
* Swap ak and nk in the input to ShortCommit<sup>ivk</sup>. They somehow got reversed relative to Sapling.
* Explain in more detail the argument for Balance.
* F might be Poseidon.
* Make some formatting changes so that it's correct Markdown.

### DIFF
--- a/book/src/design/nullifiers.md
+++ b/book/src/design/nullifiers.md
@@ -127,7 +127,17 @@ In the above alternatives:
 In order to satisfy the **Balance** security property, we require that the circuit must be
 able to enforce that only one nullifier is accepted for a given note. As in Sprout and
 Sapling, we achieve this by ensuring that the nullifier deterministically depends only on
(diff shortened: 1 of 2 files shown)


book/src/design/nullifiers.md
```diff
@@ -29,22 +30,26 @@ We care about several security properties for our nullifiers:
 - **Balance:** can I forge money?
 
 - **Note Privacy:** can I gain information about notes only from the public block chain?
+
```
str4d, line 33: These spaces separating different bullet-point levels are necessary for valid reStructuredText, not Markdown.

str4d (approved):

book/src/design/nullifiers.md
daira: Ah, good point. I always get confused, so I usually write in a subset that minimises the differences.

PR #14, opened 2021-01-21
CI: Update benchmarks runner

str4d:


str4d:
This fixes the benchmark runner (it is now back to failing because there are no benchmarks yet).

PR #15, opened 2021-01-22
book: Note that we use 0 for uncommitted leaves in the commitment tree

str4d:


book/src/design/commitment-tree.md
```diff
@@ -29,3 +29,33 @@ clients can assert they are valid independently of the full block.
...
+
+The fixed-depth incremental Merkle trees that we use (in Sprout and Sapling, and again in
+Orchard) require specifying an "empty" or "uncommitted" leaf - a value that will never be
+appended to the tree as a regular leaf.
+
+- For Sprout (and trees composed of the outputs of bit-twiddling hash functions), we use
+  the all-zeroes array; the probability of a real note having a colliding note commitment
+  is cryptographically negligible.
+- For Sapling (where leaves are u-coordinates of Jubjub points), we use the value $1$
+  (which is not the u-coordinate of any Jubjub point).
```
daira, line 43:
```suggestion
- For Sapling, where leaves are $u$-coordinates of Jubjub points, we use the value $1$
  which is not the u-coordinate of any Jubjub point.
```

book/src/design/commitment-tree.md
```diff
@@ -29,3 +29,33 @@ clients can assert they are valid independently of the full block.
...
+
+Orchard note commitments are the x-coordinates of Pallas points; thus we take the same
+approach as Sapling, using a value that is not the x-coordinate of any Pallas point as the
```
daira, line 46:
```suggestion
Orchard note commitments are the $x$-coordinates of Pallas points; thus we take the same
approach as Sapling, using a value that is not the $x$-coordinate of any Pallas point as the
```

daira (approved):
ACK with minor suggestions.

PR #17, opened 2021-02-08
book: Document design rationale for Orchard keys and addresses

str4d:


book/src/design/keys.md
```diff
@@ -0,0 +1,86 @@
...
+  same $\mathsf{ivk}$ but different $\mathsf{ovk}$s. Users still have control over whether
+  $\mathsf{ovk}$ is used when constructing a transaction.
+
+- All diversifiers now result in valid payment addresses, due to group hashing into Pallas
+  being specified to be infallible. This removes significant complexity from the use cases
+  for diversified addresses.
+
+Other than the above, Orchard retains the same design rationale for its keys and addresses
+as Sapling. For example, diversifiers remain at 11 bytes, so that Orchard addresses are
+the same length as Sapling addresses.
```
daira, line 35:
```suggestion
Other than the above, Orchard retains the same design rationale for its keys and addresses
as Sapling, and the same bech32 encoding. For example, diversifiers remain at 11 bytes, so
that Orchard addresses are the same length as Sapling addresses.
```

book/src/design/keys.md
```diff
@@ -0,0 +1,86 @@
+# Keys and addresses
+
+Orchard keys and payment addresses are structurally similar to Sapling. The main change is
+that Orchard keys use the Pallas curve instead of Jubjub, in order to enable the future
+use of the Pallas-Vesta curve cycle in the Orchard protocol. This involves corresponding
+changes to the key derivation process (such as using Sinsemilla for Pallas-efficient
+commitments).
```
daira, line 7:
```suggestion
use of the Pallas-Vesta curve cycle in the Orchard protocol. (We already use Vesta as
the curve on which Halo 2 proofs are computed, but this doesn't yet require a cycle.)

Using the Pallas curve and making the most efficient use of the Halo 2 proof system
involves corresponding changes to the key derivation process, such as using Sinsemilla
for Pallas-efficient commitments. We also take the opportunity to remove all uses of
expensive general-purpose hashes (such as BLAKE2s) from the circuit.
```

book/src/design/keys.md
```diff
@@ -0,0 +1,86 @@
...
+
+We also make several structural changes, building on the lessons learned from Sapling:
```
daira, line 9:
```suggestion
We make several structural changes, building on the lessons learned from Sapling:
```

book/src/design/keys.md
```diff
@@ -0,0 +1,86 @@
...
+      - (N) Normal addresses
+        - (N) Address 0
+        - (N) Address 1...
+      - (N) Change addresses
+        - (N) Change address 0
+        - (N) Change address 1...
+    - (H) Account 1...
+
+Shielded accounts do not require separating change addresses from normal addresses, due to
+addresses not being revealed in transactions. Similarly, there is also no need to generate
```
daira, line 64:
```suggestion
Shielded accounts do not require separating change addresses from normal addresses, because
addresses are not revealed in transactions. Similarly, there is also no need to generate
```

book/src/design/keys.md
```diff
@@ -0,0 +1,86 @@
...
+a fresh spending key for every transaction, and in fact this causes a linear slow-down in
+wallet scanning. But for users that do want to generate multiple addresses per account,
+they can generate the following structure, which does not use non-hardened derivation:
```
daira, line 67:
```suggestion
a fresh spending key for every transaction, and in fact this would cause a linear slow-down
in wallet scanning. But for users that do want to generate multiple addresses per account,
they can generate the following structure, which does not use non-hardened derivation:
```

book/src/design/keys.md
```diff
@@ -0,0 +1,86 @@
...
+      - Diversified address 0
+      - Diversified address 1...
+    - (H) Account 1...
+
+Non-hardened derivation is therefore only required for use-cases that require the ability
+to derive more than one child layer of addresses. However, in the years since Sapling was
+deployed, we have not seen *any* such use cases appear.
+
+Therefore, for Orchard we only define hardened derivation, and do so with a much simpler
+design that ZIP 32. All derivations produce an opaque binary spending key, from which the
```
daira, line 81:
```suggestion
design than ZIP 32. All derivations produce an opaque binary spending key, from which the
```

book/src/design/keys.md
```diff
@@ -0,0 +1,86 @@
...
+keys and addresses are then derived.
```
daira, line 82:
```suggestion
keys and addresses are then derived. As a side benefit, this makes key formats
shorter. (The formats that will actually be used in practice for Orchard will correspond
to the simpler Sapling formats in the protocol specification, rather than the longer
and more complicated "extended" ones defined by ZIP 32.)
```

daira (approved):
ACK with suggestions.

book/src/design/keys.md
str4d: This might not quite be the case (we might still want something akin to the "chain code" in ZIP 32 when actually deriving spending keys), but the spending key structure itself should end up the same. We should decide how to derive a diversifier key though (probably just from the spending key bytes, rather than as an artifact of HD derivation).

daira: We should pair on working out how the ZIP 32 derivation will work for Orchard.

PR #20, opened 2021-02-24
Minor API changes

str4d:


nuttycom (approved):
ACK modulo Clippy complaints

str4d:
The clippy complaints are in unrelated parts of the code, and are "unused" lints that will go away once things start getting implemented.

PR #21, opened 2021-02-24
Implement Sinsemilla primitives

str4d:
Test vectors will be added in a subsequent PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ publish = false
 rustdoc-args = [ "--html-in-header", "katex-header.html" ]
 
 [dependencies]
+group = "0.9"
 halo2 = { git = "https://github.com/zcash/halo2.git", branch = "main" }
 nonempty = "0.6"
 
--- a/src/primitives/sinsemilla.rs
+++ b/src/primitives/sinsemilla.rs
@@ -1,7 +1,5 @@
 //! The Sinsemilla hash function and commitment scheme.
 
-use std::iter;
-
 use group::{Curve, Group};
 use halo2::{
     arithmetic::{CurveAffine, CurveExt},
@@ -18,6 +16,52 @@ fn lebs2ip_32(bits: &[bool]) -> u32 {
     bits.iter()
         .enumerate()
         .fold(0u32, |acc, (i, b)| acc + if *b { 1 << i } else { 0 })
+}
+
+struct Pad<I: Iterator<Item = bool>> {
+    inner: I,
+    len: usize,
+    padding_left: Option<usize>,
+}
+
+impl<I: Iterator<Item = bool>> Pad<I> {
+    fn new(inner: I) -> Self {
+        Pad {
+            inner,
+            len: 0,
+            padding_left: None,
+        }
+    }
+}
+
+impl<I: Iterator<Item = bool>> Iterator for Pad<I> {
+    type Item = bool;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            if let Some(n) = self.padding_left.as_mut() {
+                if *n == 0 {
+                    break None;
+                } else {
+                    *n -= 1;
+                    break Some(false);
+                }
+            } else if let Some(ret) = self.inner.next() {
+                self.len += 1;
+                assert!(self.len <= K * C);
+                break Some(ret);
+            } else {
+                // Inner iterator just ended.
+                let rem = self.len % K;
+                if rem > 0 {
+                    self.padding_left = Some(K - rem);
+                } else {
+                    // No padding required.
+                    self.padding_left = Some(0);
+                }
+            }
+        }
+    }
 }
 
 /// Hash extractor for Pallas, from [§ 5.4.8.7].
@@ -42,13 +86,8 @@ fn Q(domain_prefix: &str) -> pallas::Point {
 ///
 /// [§ 5.4.1.9]: https://zips.z.cash/protocol/orchard.pdf#concretesinsemillahash
 #[allow(non_snake_case)]
-pub(crate) fn hash_to_point(
-    domain_prefix: &str,
-    msg: impl Iterator<Item = bool> + ExactSizeIterator,
-) -> pallas::Point {
-    assert!(msg.len() <= K * C);
-    let pad = msg.len() % K;
-    let padded: Vec<_> = msg.chain(iter::repeat(false).take(pad)).collect();
+pub(crate) fn hash_to_point(domain_prefix: &str, msg: impl Iterator<Item = bool>) -> pallas::Point {
+    let padded: Vec<_> = Pad::new(msg).collect();
 
     let hasher_S = pallas::Point::hash_to_curve(GROUP_HASH_S);
     let S = |chunk: &[bool]| hasher_S(&lebs2ip_32(chunk).to_le_bytes());
@@ -61,10 +100,7 @@ pub(crate) fn hash_to_point(
 /// `SinsemillaHash` from [§ 5.4.1.9].
 ///
 /// [§ 5.4.1.9]: https://zips.z.cash/protocol/orchard.pdf#concretesinsemillahash
-pub(crate) fn hash(
-    domain_prefix: &str,
-    msg: impl Iterator<Item = bool> + ExactSizeIterator,
-) -> pallas::Base {
+pub(crate) fn hash(domain_prefix: &str, msg: impl Iterator<Item = bool>) -> pallas::Base {
     extract(&hash_to_point(domain_prefix, msg))
 }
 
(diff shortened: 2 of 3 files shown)


src/primitives/sinsemilla.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/// `SinsemillaHashToPoint` from [§ 5.4.1.9].
+///
+/// [§ 5.4.1.9]: https://zips.z.cash/protocol/orchard.pdf#concretesinsemillahash
+#[allow(non_snake_case)]
+pub(crate) fn hash_to_point(
+    domain_prefix: &str,
+    msg: impl Iterator<Item = bool> + ExactSizeIterator,
+) -> pallas::Point {
```
therealyingtong, line 48: should this be `pallas::Affine` instead?

src/primitives/sinsemilla.rs
```diff
@@ -0,0 +1,97 @@
...
+fn lebs2ip_32(bits: &[bool]) -> u32 {
+    bits.iter()
+        .enumerate()
+        .fold(0u32, |acc, (i, b)| acc + if *b { 1 << i } else { 0 })
+}
+
+/// Hash extractor for Pallas, from [§ 5.4.8.7].
+///
+/// [§ 5.4.8.7]: https://zips.z.cash/protocol/orchard.pdf#concreteextractorpallas
+fn extract(point: &pallas::Point) -> pallas::Base {
```
dconnolly, line 26: Same here, should probably be `pallas::Affine`; `Extract_P` is defined as having a `P` element input which is an affine point:
image
image

Style opinion, as this is an open PR vs needing to open a new thing in `halo2`: naming all point types as explicitly *Point, such as `AffinePoint`, `ProjectivePoint`, etc, which can improve readability and human distinguishability of types. The [`jubjub` crate](https://docs.rs/jubjub/0.6.0/jubjub/) basically does this. 

I know in `halo2` a lot of these are generated by macros but the [names are chosen on macro invocation](https://github.com/zcash/halo2/blob/63083e418cdd836e0056dd55f6498ccf185af83f/src/pasta/curves.rs#L908) and renamed and exported in the [`pallas` module](https://github.com/zcash/halo2/blob/main/src/pasta/pallas.rs#L15).  This is of course not limited to or blocking on this PR but is a kind of naming / typing style thing we zcash engineers can discuss, if desired.

src/primitives/sinsemilla.rs
dconnolly: Agreed, the spec currently defines `P` as the group of points `(x,y)` on the short Weierstrass form of the curve.

image
image

src/primitives/sinsemilla.rs
```diff
@@ -0,0 +1,97 @@
...
+    msg: impl Iterator<Item = bool> + ExactSizeIterator,
+) -> pallas::Base {
+    extract(&hash_to_point(domain_prefix, msg))
+}
+
+/// `SinsemillaCommit` from [§ 5.4.7.4].
+///
+/// [§ 5.4.7.4]: https://zips.z.cash/protocol/orchard.pdf#concretesinsemillacommit
+#[allow(non_snake_case)]
+pub(crate) fn commit(
```
dconnolly, line 75: `SinsemillaCommit` as an instance of `SinsemillaHashToPoint` also returns an affine point as an element of `P`

image
image
image

therealyingtong: yup, @str4d i've made a few changes to `primitives/sinsemilla.rs` to get it working with my WIP chip here: https://github.com/zcash/orchard/pull/25/files

src/primitives/sinsemilla.rs
str4d: `pallas::Point` and `pallas::Affine` are equivalent group elements by construction and have 1:1 conversion (enforced by the `group::Curve` and `group::prime::{PrimeCurve, PrimeCurveAffine}` traits). However, `pallas::Point: group::Curve` is the format that is efficient to perform operations on, and is intended to be what is passed around in-memory. `pallas::Affine` is only intended for encoding/decoding and accessing affine coordinates. Returning `pallas::Affine` from this API would force an unnecessary `projective -> affine -> projective` round-trip on every call to `sinsemilla::commit` or `sinsemilla::short_commit`.

src/primitives/sinsemilla.rs
str4d: Having this take `pallas::Affine` pushes a `point.to_affine()` call onto every caller, instead of locating them all inside this function. There are only a couple of callers though, so I don't overly mind.

Re: naming of `pallas::Point`, I opted in the `halo2` crate to leave the point representation out of the re-exported name for overall readability (and also because as above, it's irrelevant to the consumer). What mattered more to me was that I could write `pallas::Point` _everywhere_, so that it was clear that this is a Pallas point (rather than Vesta or some other curve); `pallas::ProjectivePoint` is almost twice as long.

`jubjub` is a slightly different case, because we need to represent both the curve type _and_ whether the point is constrained to the prime-order subgroup, so we already need to include more information in the name (in hindsight `jubjub::FullPoint` would have been a better name for this purpose than `jubjub::ExtendedPoint`). Pallas and Vesta are prime-order curves and don't have this problem.

src/primitives/sinsemilla.rs
dconnolly: I basically retract my comment, as I did basically the same thing in https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/sapling/commitment/pedersen_hashes.rs, even though the [Jubjub group element definitions in the spec](https://zips.z.cash/protocol/protocol.pdf#concretepedersenhash) are also 'just' `(u, v)` etc. 🙈

I would love if the type was named `pallas::ProjectivePoint` or something. 😅

src/primitives/sinsemilla.rs
dconnolly: `pallas::ProjectivePoint` may be longer but it's not super long, and since there are multiple point types, vs a 'Point' and an 'Affine', it would be nice in terms of having the code document itself. Luckily Rust compiles away token names so we need not be precious on name lengths, as long we we don't turn into Java (😅 )

str4d: I'm only weakly against a rename (I like Dodger's "weakly" vs "strongly" for indicating preference), so I'll defer to others on this.

src/primitives/sinsemilla.rs
therealyingtong: i've reverted these changes; now calling `to_affine()` from the caller instead. :+1:

Cargo.toml
```diff
@@ -19,7 +19,8 @@ publish = false
 rustdoc-args = [ "--html-in-header", "katex-header.html" ]
 
 [dependencies]
-halo2 = { git = "https://github.com/zcash/halo2.git", branch = "main" }
+group = "0.9"
+halo2 = { git = "https://github.com/zcash/halo2.git", rev = "84c63e5e84f97a3cdbd5621ed3d20acf9c0a5167" }
```
therealyingtong, line 23: was this change referencing the specific commit intentional?

src/primitives/sinsemilla.rs
daira: What @str4d said. Basically you're always allowed to use isomorphic representations.

src/primitives/sinsemilla.rs
daira: I like the current short names.

src/primitives/sinsemilla.rs
```diff
@@ -0,0 +1,97 @@
...
+    pasta::pallas,
+};
+
+const GROUP_HASH_Q: &str = "z.cash:SinsemillaQ";
+const GROUP_HASH_S: &str = "z.cash:SinsemillaS";
+
+const K: usize = 10;
+const C: usize = 253;
+
+fn lebs2ip_32(bits: &[bool]) -> u32 {
```
str4d, line 17:
```suggestion
fn lebs2ip_10(bits: &[bool]) -> u32 {
```

src/primitives/sinsemilla.rs
```diff
@@ -0,0 +1,97 @@
...
+pub(crate) fn hash_to_point(
+    domain_prefix: &str,
+    msg: impl Iterator<Item = bool> + ExactSizeIterator,
+) -> pallas::Point {
+    assert!(msg.len() <= K * C);
+    let pad = msg.len() % K;
+    let padded: Vec<_> = msg.chain(iter::repeat(false).take(pad)).collect();
+
+    let hasher_S = pallas::Point::hash_to_curve(GROUP_HASH_S);
+    let S = |chunk: &[bool]| hasher_S(&lebs2ip_32(chunk).to_le_bytes());
```
str4d, line 54:
```suggestion
    let S = |chunk: &[bool]| hasher_S(&lebs2ip_10(chunk).to_le_bytes());
```

ebfull (approved):

PR #24, opened 2021-02-26
[WIP] Sinsemilla gadget and chip

therealyingtong:
Based on #21.


therealyingtong:
The Sinsemilla API has changed significantly relative to the ECC chip, so we're frying this PR in favour of https://github.com/zcash/orchard/pull/57

PR #40, opened 2021-03-05
Orchard key components

### LINKED ISSUE #29: Implement Orchard key structure

str4d:
Closes #29.

### DIFF
--- a/src/address.rs
+++ b/src/address.rs
@@ -1,8 +1,16 @@
+use halo2::pasta::pallas;
+
 use crate::keys::Diversifier;
 
 /// A shielded payment address.
 #[derive(Debug)]
 pub struct Address {
     d: Diversifier,
-    pk_d: (),
+    pk_d: pallas::Point,
+}
+
+impl Address {
+    pub(crate) fn from_parts(d: Diversifier, pk_d: pallas::Point) -> Self {
+        Address { d, pk_d }
+    }
 }
--- a/src/address.rs
+++ b/src/address.rs
@@ -3,6 +3,15 @@ use halo2::pasta::pallas;
 use crate::keys::Diversifier;
 
 /// A shielded payment address.
+///
+/// # Examples
+///
+/// ```
+/// use orchard::keys::{SpendingKey, FullViewingKey};
+///
+/// let sk = SpendingKey::from_bytes([7; 32]).unwrap();
+/// let address = FullViewingKey::from(&sk).default_address();
+/// ```
 #[derive(Debug)]
 pub struct Address {
     d: Diversifier,
--- a/src/address.rs
+++ b/src/address.rs
@@ -1,6 +1,4 @@
-use halo2::pasta::pallas;
-
-use crate::keys::Diversifier;
+use crate::keys::{DiversifiedTransmissionKey, Diversifier};
 
 /// A shielded payment address.
 ///
@@ -15,11 +13,11 @@ use crate::keys::Diversifier;
 #[derive(Debug)]
 pub struct Address {
     d: Diversifier,
-    pk_d: pallas::Point,
+    pk_d: DiversifiedTransmissionKey,
 }
 
 impl Address {
-    pub(crate) fn from_parts(d: Diversifier, pk_d: pallas::Point) -> Self {
+    pub(crate) fn from_parts(d: Diversifier, pk_d: DiversifiedTransmissionKey) -> Self {
         Address { d, pk_d }
     }
 }
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -1,0 +1,4 @@
+//! Constants used in the Orchard protocol.
+
+/// $\ell^\mathsf{Orchard}_\mathsf{scalar}$
+pub(crate) const L_ORCHARD_SCALAR: usize = 255;
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -1,4 +1,4 @@
 //! Constants used in the Orchard protocol.
 
-/// $\ell^\mathsf{Orchard}_\mathsf{scalar}$
-pub(crate) const L_ORCHARD_SCALAR: usize = 255;
+/// $\ell^\mathsf{Orchard}_\mathsf{base}$
+pub(crate) const L_ORCHARD_BASE: usize = 255;
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -3,6 +3,8 @@
 use std::convert::TryInto;
 use std::mem;
 
+use aes::Aes256;
+use fpe::ff1::{BinaryNumeralString, FF1};
 use group::GroupEncoding;
 use halo2::{arithmetic::FieldExt, pasta::pallas};
 use subtle::CtOption;
@@ -176,8 +178,12 @@ impl DiversifierKey {
     }
 
     /// Returns the diversifier at the given index.
-    pub fn get(&self, _: impl Into<DiversifierIndex>) -> Diversifier {
-        todo!()
+    pub fn get(&self, j: impl Into<DiversifierIndex>) -> Diversifier {
+        let ff = FF1::<Aes256>::new(&self.0, 2).expect("valid radix");
+        let enc = ff
+            .encrypt(&[], &BinaryNumeralString::from_bytes_le(&j.into().0[..]))
+            .unwrap();
+        Diversifier(enc.to_bytes_le().try_into().unwrap())
     }
 }
 
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -20,9 +20,9 @@ use crate::{
 
 /// A spending key, from which all key material is derived.
 ///
-/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][§4.2.3].
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
 ///
-/// [§4.2.3]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
 #[derive(Debug)]
 pub struct SpendingKey([u8; 32]);
 
@@ -40,9 +40,9 @@ impl SpendingKey {
 
 /// A spend authorizing key, used to create spend authorization signatures.
 ///
-/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][§4.2.3].
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
 ///
-/// [§4.2.3]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
 #[derive(Debug)]
 pub(crate) struct SpendAuthorizingKey(redpallas::SigningKey<SpendAuth>);
 
@@ -81,11 +81,11 @@ impl From<&SpendAuthorizingKey> for AuthorizingKey {
 
 /// A key used to derive [`Nullifier`]s from [`Note`]s.
 ///
-/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][§4.2.3].
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
 ///
 /// [`Nullifier`]: crate::note::Nullifier;
 /// [`Note`]: crate::note::Note;
-/// [§4.2.3]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
 #[derive(Debug)]
 pub(crate) struct NullifierDerivingKey(pallas::Base);
 
@@ -117,9 +117,9 @@ impl From<&SpendingKey> for FullViewingKey {
 }
 
 impl FullViewingKey {
-    /// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][§4.2.3].
+    /// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
     ///
-    /// [§4.2.3]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
+    /// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
     fn derive_dk_ovk(&self) -> (DiversifierKey, OutgoingViewingKey) {
         let k = self.rivk.to_bytes();
         let b = [(&self.ak.0).into(), self.nk.0.to_bytes()];
@@ -238,9 +238,9 @@ impl From<&FullViewingKey> for OutgoingViewingKey {
 pub(crate) struct DiversifiedTransmissionKey(pallas::Point);
 
 impl DiversifiedTransmissionKey {
-    /// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][§4.2.3].
+    /// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
     ///
-    /// [§4.2.3]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
+    /// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
     fn derive(ivk: &IncomingViewingKey, d: &Diversifier) -> Self {
         let g_d = diversify_hash(&d.0);
         DiversifiedTransmissionKey(ka_orchard(&ivk.0, &g_d))
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -7,7 +7,7 @@ use aes::Aes256;
 use fpe::ff1::{BinaryNumeralString, FF1};
 use group::GroupEncoding;
 use halo2::{arithmetic::FieldExt, pasta::pallas};
-use subtle::{Choice, CtOption};
+use subtle::CtOption;
 
 use crate::{
     address::Address,
@@ -32,17 +32,9 @@ impl SpendingKey {
     /// Returns `None` if the bytes do not correspond to a valid Orchard spending key.
     pub fn from_bytes(sk: [u8; 32]) -> CtOption<Self> {
         let sk = SpendingKey(sk);
-        // If ask = 0, or the default address would be ⊥, discard this key.
-        let ask_not_zero = !SpendAuthorizingKey::derive_inner(&sk).ct_is_zero();
-        let have_default_address = Choice::from({
-            let fvk = FullViewingKey::from(&sk);
-            if fvk.default_address_inner().is_some() {
-                1
-            } else {
-                0
-            }
-        });
-        CtOption::new(sk, ask_not_zero & have_default_address)
+        // If ask = 0, discard this key.
+        let ask = SpendAuthorizingKey::derive_inner(&sk);
+        CtOption::new(sk, !ask.ct_is_zero())
     }
 }
 
@@ -164,31 +156,16 @@ impl FullViewingKey {
 
     /// Returns the default payment address for this key.
     pub fn default_address(&self) -> Address {
-        self.default_address_inner()
-            .expect("Default address works by construction")
-    }
-
-    fn default_address_inner(&self) -> Option<Address> {
         self.address(DiversifierKey::from(self).default_diversifier())
     }
 
     /// Returns the payment address for this key at the given index.
-    ///
-    /// Returns `None` if the diversifier does not correspond to an address. This happens
-    /// with negligible probability; in most cases unwrapping the result will be fine, but
-    /// if you have specific stability requirements then you can either convert this into
-    /// an error, or try another diversifier index (e.g. incrementing).
-    pub fn address_at(&self, j: impl Into<DiversifierIndex>) -> Option<Address> {
+    pub fn address_at(&self, j: impl Into<DiversifierIndex>) -> Address {
         self.address(DiversifierKey::from(self).get(j))
     }
 
     /// Returns the payment address for this key corresponding to the given diversifier.
-    ///
-    /// Returns `None` if the diversifier does not correspond to an address. This happens
-    /// with negligible probability; in most cases unwrapping the result will be fine, but
-    /// if you have specific stability requirements then you can either convert this into
-    /// an error, or try another diversifier.
-    pub fn address(&self, d: Diversifier) -> Option<Address> {
+    pub fn address(&self, d: Diversifier) -> Address {
         IncomingViewingKey::from(self).address(d)
     }
 }
@@ -275,13 +252,9 @@ impl From<&FullViewingKey> for IncomingViewingKey {
 
 impl IncomingViewingKey {
     /// Returns the payment address for this key corresponding to the given diversifier.
-    ///
-    /// Returns `None` if the diversifier does not correspond to an address. This happens
-    /// with negligible probability; in most cases unwrapping the result will be fine, but
-    /// if you have specific stability requirements then you can either convert this into
-    /// an error, or try another diversifier.
-    pub fn address(&self, d: Diversifier) -> Option<Address> {
-        DiversifiedTransmissionKey::derive(self, &d).map(|pk_d| Address::from_parts(d, pk_d))
+    pub fn address(&self, d: Diversifier) -> Address {
+        let pk_d = DiversifiedTransmissionKey::derive(self, &d);
+        Address::from_parts(d, pk_d)
     }
 }
 
@@ -315,7 +288,8 @@ impl DiversifiedTransmissionKey {
     /// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
     ///
     /// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
-    fn derive(ivk: &IncomingViewingKey, d: &Diversifier) -> Option<Self> {
-        diversify_hash(&d.0).map(|g_d| DiversifiedTransmissionKey(ka_orchard(&ivk.0, &g_d)))
+    fn derive(ivk: &IncomingViewingKey, d: &Diversifier) -> Self {
+        let g_d = diversify_hash(&d.0);
+        DiversifiedTransmissionKey(ka_orchard(&ivk.0, &g_d))
     }
 }
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -116,6 +116,10 @@ impl From<&SpendingKey> for CommitIvkRandomness {
 ///
 /// This key is useful anywhere you need to maintain accurate balance, but do not want the
 /// ability to spend funds (such as a view-only wallet).
+///
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
+///
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
 #[derive(Debug)]
 pub struct FullViewingKey {
     ak: SpendValidatingKey,
@@ -159,6 +163,10 @@ impl FullViewingKey {
 }
 
 /// A key that provides the capability to derive a sequence of diversifiers.
+///
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
+///
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
 #[derive(Debug)]
 pub struct DiversifierKey([u8; 32]);
 
@@ -205,6 +213,10 @@ impl DiversifierKey {
 
 /// A diversifier that can be used to derive a specific [`Address`] from a
 /// [`FullViewingKey`] or [`IncomingViewingKey`].
+///
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
+///
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
 #[derive(Debug)]
 pub struct Diversifier([u8; 11]);
 
@@ -216,6 +228,10 @@ pub struct Diversifier([u8; 11]);
 ///
 /// This key is not suitable for use on its own in a wallet, as it cannot maintain
 /// accurate balance. You should use a [`FullViewingKey`] instead.
+///
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
+///
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
 #[derive(Debug)]
 pub struct IncomingViewingKey(pallas::Scalar);
 
@@ -239,6 +255,10 @@ impl IncomingViewingKey {
 ///
 /// This key is not suitable for use on its own in a wallet, as it cannot maintain
 /// accurate balance. You should use a [`FullViewingKey`] instead.
+///
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
+///
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
 #[derive(Debug)]
 pub struct OutgoingViewingKey([u8; 32]);
 
@@ -249,6 +269,10 @@ impl From<&FullViewingKey> for OutgoingViewingKey {
 }
 
 /// The diversified transmission key for a given payment address.
+///
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
+///
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
 #[derive(Debug)]
 pub(crate) struct DiversifiedTransmissionKey(pallas::Point);
 
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -13,8 +13,7 @@ use crate::{
     address::Address,
     primitives::redpallas::{self, SpendAuth},
     spec::{
-        commit_ivk, diversify_hash, extract_p, ka_orchard, prf_expand, prf_expand_vec, to_base,
-        to_scalar,
+        commit_ivk, diversify_hash, ka_orchard, prf_expand, prf_expand_vec, to_base, to_scalar,
     },
 };
 
@@ -219,7 +218,7 @@ pub struct IncomingViewingKey(pallas::Scalar);
 impl From<&FullViewingKey> for IncomingViewingKey {
     fn from(fvk: &FullViewingKey) -> Self {
         let ak = pallas::Point::from_bytes(&(&fvk.ak.0).into()).unwrap();
-        IncomingViewingKey(commit_ivk(&extract_p(&ak), &fvk.nk.0, &fvk.rivk.0))
+        IncomingViewingKey(commit_ivk(&ak, &fvk.nk.0, &fvk.rivk.0))
     }
 }
 
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -7,7 +7,7 @@ use aes::Aes256;
 use fpe::ff1::{BinaryNumeralString, FF1};
 use group::GroupEncoding;
 use halo2::{arithmetic::FieldExt, pasta::pallas};
-use subtle::CtOption;
+use subtle::{Choice, CtOption};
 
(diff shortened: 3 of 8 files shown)


dconnolly (commented):
Before getting into the details, any tests?

In zebra-chain we implemented proptest strategies for the root of the key derivation chain, and wrote tests to check derivation like this:
https://github.com/ZcashFoundation/zebra/blob/main/zebra-chain/src/sapling/keys/tests.rs

src/keys.rs
```diff
@@ -1,70 +1,193 @@
...
+                DiversifierIndex(j_bytes)
+            }
+        }
+    };
+}
+di_from!(u8);
+di_from!(u16);
+di_from!(u32);
+di_from!(u64);
+di_from!(usize);
```
dconnolly, line 173: Are these all expected to be needed?

src/keys.rs
```diff
@@ -79,18 +202,20 @@ pub struct Diversifier([u8; 11]);
...
+        let ak = pallas::Point::from_bytes(&(&fvk.ak.0).into()).unwrap();
+        IncomingViewingKey(commit_ivk(&extract_p(&ak), &fvk.nk.0, &fvk.rivk))
     }
 }
 
 impl IncomingViewingKey {
     /// Returns the payment address for this key corresponding to the given diversifier.
-    pub fn address(&self, _: Diversifier) -> Address {
-        todo!()
+    pub fn address(&self, d: Diversifier) -> Address {
```
dconnolly, line 216: Perhaps an `impl From<(IncomingViewingKey, Diversifier)> for Address`?

src/address.rs
```diff
@@ -1,8 +1,16 @@
...
     d: Diversifier,
-    pk_d: (),
+    pk_d: pallas::Point,
+}
+
+impl Address {
+    pub(crate) fn from_parts(d: Diversifier, pk_d: pallas::Point) -> Self {
+        Address { d, pk_d }
+    }
 }
```
dconnolly, line 16: As mentioned in `keys.rs`, instead of a new method `from_parts` this could be accomplished with `impl From<(Diversifier, DiversifiedTransmissionKey)> for Address`

src/address.rs
```diff
@@ -1,8 +1,16 @@
+use halo2::pasta::pallas;
+
 use crate::keys::Diversifier;
 
 /// A shielded payment address.
 #[derive(Debug)]
 pub struct Address {
     d: Diversifier,
-    pk_d: (),
+    pk_d: pallas::Point,
```
dconnolly, line 9: `pallas::Point` is used in several places for different purposes in `orchard`, should we have a wrapper type like `orchard::keys::DiversifiedTransmissionKey` or whatever?  That's what the spec calls pk_d on the March 4 version.

str4d:
> Before getting into the details, any tests?

The Zebra tests you linked to fall into two categories:
- Checking against test vectors.
- Checking round-trip encodings of keys and addresses.

This PR does not implement encodings, so the latter do not apply. For the former, we are writing a parallel implementation in Python for the purpose of generating test vectors that this implementation can be tested against. This is started in https://github.com/zcash-hackworks/zcash-test-vectors/pull/14

src/keys.rs
```diff
@@ -1,70 +1,193 @@
...
 
 /// A spending key, from which all key material is derived.
 ///
-/// TODO: In Sapling we never actually used this, instead deriving everything via ZIP 32,
-/// so that we could maintain Bitcoin-like HD keys with properties like non-hardened
-/// derivation. If we decide that we don't actually require non-hardened derivation, then
-/// we could greatly simplify the HD structure and use this struct directly.
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][§4.2.3].
+///
+/// [§4.2.3]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
```
dconnolly, line 25: Is there a danger of these numbers changing? If so, maybe using `[spec]` or maybe `[orchardkeycomponents]` will be less likely to become inaccurate

src/keys.rs
str4d: I dislike overusing `From<(tuple)>` in APIs; the resulting `(a, b).into()` makes introspection of the code difficult without an IDE.

src/address.rs
str4d: This is an internal API, for which I prefer to be explicit.

src/keys.rs
```diff
@@ -1,70 +1,193 @@
...
+}
 
 impl From<&SpendingKey> for SpendAuthorizingKey {
-    fn from(_: &SpendingKey) -> Self {
-        todo!()
+    fn from(sk: &SpendingKey) -> Self {
+        let ask = Self::derive_inner(sk);
+        // SpendingKey cannot be constructed such that this assertion would fail.
+        assert!(!bool::from(ask.ct_is_zero()));
+        // TODO: Add TryFrom<S::Scalar> for SpendAuthorizingKey.
```
dconnolly, line 61: ❤️

src/keys.rs
str4d: If the numbers become inaccurate, they'd be inaccurate in the link text as well as the underlying link handle, and if we're updating the former then the latter would be handled at the same time. That being said, I'm fine with having the link handles follow the URL anchors.

src/keys.rs
str4d: `u8` and `u16` are probably the least likely to be used. `usize` makes sense from a Rust perspective, and `u32, u64` are quite common to see as e.g. customer IDs in SQL tables.

src/keys.rs
```diff
@@ -1,70 +1,193 @@
...
+    /// Returns `None` if the bytes do not correspond to a valid Orchard spending key.
+    pub fn from_bytes(sk: [u8; 32]) -> CtOption<Self> {
+        let sk = SpendingKey(sk);
+        // If ask = 0, discard this key.
+        let ask = SpendAuthorizingKey::derive_inner(&sk);
+        CtOption::new(sk, !ask.ct_is_zero())
+    }
+}
 
+/// A spending authorizing key, used to create spend authorization signatures.
```
dconnolly, line 41:
```suggestion
/// A spend authorizing key, used to create spend authorization signatures.
```

dconnolly:
> > Before getting into the details, any tests?
> 
> The Zebra tests you linked to fall into two categories:
> 
> * Checking against test vectors.
> * Checking round-trip encodings of keys and addresses.
> 
> This PR does not implement encodings, so the latter do not apply (yet). For the former, we are writing a parallel implementation in Python (ideally by someone who is not involved in writing _this_ implementation) for the purpose of generating test vectors that this implementation can be tested against. This is started in [zcash-hackworks/zcash-test-vectors#14](https://github.com/zcash-hackworks/zcash-test-vectors/pull/14)

Fantastic! 

Are there tests we can create to make sure the code in this PR is exercised somewhat, before those test vectors are available?

src/keys.rs
dconnolly: I find it nice as the docs then generate all the `From`s for an `Address` or another type, all together, but

src/address.rs
dconnolly: For Zebra usage this may not stay internal

str4d:
I don't see any meaningful unit tests (we could test e.g. `FullViewingKey::default_address`, but that's kind-of pointless for _me_ to write at the same time; if I made a mistake, I'd likely duplicate it in the test). But we could definitely add documentation tests that checks the code runs.

src/keys.rs
```diff
@@ -1,70 +1,193 @@
...
-        todo!()
+    fn from(sk: &SpendingKey) -> Self {
+        NullifierDerivingKey(to_base(prf_expand(&sk.0, &[0x07])))
     }
 }
 
 /// A key that provides the capability to view incoming and outgoing transactions.
 ///
 /// This key is useful anywhere you need to maintain accurate balance, but do not want the
 /// ability to spend funds (such as a view-only wallet).
```
dconnolly, line 101:
```suggestion
/// An Orchard full viewing key, as defined in [Zcash Protocol Spec §4.2.3: Orchard Full Viewing Keys][orchardfullviewingkeyencoding]
///
/// Provides the capability to view incoming and outgoing transactions. This key is 
/// useful anywhere you need to maintain accurate balance, but do not want the ability 
/// to spend funds (such as a view-only wallet).
///
/// [orchardfullviewingkeyencoding]: https://zips.z.cash/protocol/nu5.pdf#orchardfullviewingkeyencoding
```

Is there a better section in the spec? Also the line wrap might be wonky

src/address.rs
str4d: If it does become necessary to expose it for Zebra, we can do so in a subsequent PR.

src/keys.rs
str4d: That section of the specification defines the string encoding of `fvk`, not `fvk` itself.

I'm not sure yet whether it makes sense to implement string encoding in this crate or the downstream users (`zcash_primitives` and `zebra-chain`), since doing so requires knowledge of the network type, which the downstream crates define but the Orchard protocol is agnostic over. It would be nice if the two downstream crates could agree on a network type, but that seems like it might require a larger refactor beyond Orchard.

src/spec.rs
```diff
@@ -0,0 +1,102 @@
...
+pub(crate) fn to_base(hash: Hash) -> pallas::Base {
+    pallas::Base::from_bytes_wide(hash.as_array())
+}
+
+/// $\mathsf{ToScalar}^\mathsf{Orchard}(x) := LEOS2IP_{\ell_\mathsf{PRFexpand}}(x) (mod r_P)$
+///
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][§4.2.3].
+///
+/// [§4.2.3]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
+pub(crate) fn to_scalar(hash: Hash) -> pallas::Scalar {
```
dconnolly, line 31: Strictly speaking, `ToScalar` takes an array of bytes as input, not just a Blake2b `Hash`. If it's being defined, shall we make it accurate, especially since that `Hash` is turning into `[u8; 64]` anyway? Same with `ToBase`

src/keys.rs
```diff
@@ -1,70 +1,193 @@
...
-///
-/// TODO: Should we just define the FVK to include extended stuff like the diversifier key?
 #[derive(Debug)]
 pub struct FullViewingKey {
     ak: AuthorizingKey,
     nk: NullifierDerivingKey,
-    rivk: (),
+    rivk: pallas::Scalar,
```
dconnolly, line 106: Perhaps a wrapper type, `IvkCommitRandomness`, that has
```rust
struct IvkCommitRandomness(pallas::Scalar);

impl From<SpendingKey> for IvkCommitRandomness {
    pub from(sk: &SpendingKey) - > Self {
        Self(pallas::Scalar::from_bytes_wide(prf_expand(&sk.0, &[0x08])))
    }
}
```
?
or using the `to_scalar` function

str4d: Isn't it already using `to_scalar`?

src/keys.rs
```diff
@@ -20,9 +20,9 @@ use crate::{
 
 /// A spending key, from which all key material is derived.
 ///
-/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][§4.2.3].
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
```
dconnolly, line 23: 🎉

src/keys.rs
str4d: Confirmed it is. But yeah I can add that as another internal type.

dconnolly:
@str4d a Q about DRYness vs, idk, explicitness in docstrings: there are several types that are similarly named between Orchard and Sapling, such as Incoming Viewing Keys, etc. In prose, not code, should we consistently distinguish them as 'Orchard Incoming Viewing Keys' etc in this library, just to be clear? Just so i don't like, suggest the same thing all over every doc comment. I don't have a particularly strong opinion on it.

src/keys.rs
```diff
@@ -95,6 +95,20 @@ impl From<&SpendingKey> for NullifierDerivingKey {
     }
 }
 
+/// The randomness for $\mathsf{Commit}^\mathsf{ivk}$.
+///
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
+///
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
+#[derive(Debug)]
+struct CommitIvkRandomness(pallas::Scalar);
```
dconnolly, line 104: 🎉🙏

str4d:
> In prose, not code, should we consistently distinguish them as 'Orchard Incoming Viewing Keys' etc in this library, just to be clear?

I'm currently of the opinion that if you're looking at the `orchard` crate docs, it should already be pretty obvious that the types are part of the Orchard protocol. We do refer to "Orchard" in prose elsewhere, but here it feels like it would just be overly repetitive. Perhaps instead we could add a disclaimer to the top-level crate documentation that we implicitly mean Orchard types here, not Sapling tpes?

src/keys.rs
dconnolly: > That section of the specification defines the string encoding of `fvk`, not `fvk` itself.
> 
> I'm not sure yet whether it makes sense to implement string encoding in this crate or the downstream users (`zcash_primitives` and `zebra-chain`), since doing so requires knowledge of the network type, which the downstream crates define but the Orchard protocol is agnostic over. It would be nice if the two downstream crates could agree on a network type, but that seems like it might require a larger refactor beyond Orchard.

Oh very agree. I think there is no better layout/description of the Orchard Full Viewing Key than this section at present, the best other section i can see in version v2021.1.16-10-g928b3d is https://zips.z.cash/protocol/nu5.pdf#addressesandkeys, unless I'm missing something or there are incoming additions. I think this is a holdover from Sapling but.

src/keys.rs
```diff
@@ -1,70 +1,207 @@
...
+    /// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
+    ///
+    /// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
+    fn derive_dk_ovk(&self) -> (DiversifierKey, OutgoingViewingKey) {
+        let k = self.rivk.0.to_bytes();
+        let b = [(&self.ak.0).into(), self.nk.0.to_bytes()];
+        let r = prf_expand_vec(&k, &[&[0x82], &b[0][..], &b[1][..]]);
+        (
+            DiversifierKey(r.as_bytes()[..32].try_into().unwrap()),
+            OutgoingViewingKey(r.as_bytes()[32..].try_into().unwrap()),
```
dconnolly, line 143: Any chance we can match on these unwraps?

str4d: Sadly no; half-open range patterns are unstable, and exclusive range patterns are experimental.

src/lib.rs
```diff
@@ -10,9 +10,11 @@
 mod address;
 pub mod bundle;
 mod circuit;
+mod constants;
 pub mod keys;
 mod note;
 pub mod primitives;
+mod spec;
```
dconnolly, line 17: An idea: the `spec` module is full of a lot of stuff, what if it was broken up approximately by section in the spec, such as `prfs`, the functions only used for key derivation in `keys` such as `to_base`/`to_scalar` etc, `commit_ivk` with `simsemilla` _maybe_, etc. I kept thinking about doing something similar in `zebra-chain` for the prfs but never did. I'm mostly worried about a `spec` module becoming a `utils` or `toolbox ` or some other catchall/dumping ground.

src/keys.rs
dconnolly: oof that may mean we can't just impl FromStr and Display on this, we'd have to wrap it or transform from this key: https://doc.zebra.zfnd.org/src/zebra_chain/sapling/keys.rs.html#684

src/keys.rs
```diff
@@ -1,70 +1,207 @@
...
+    pub fn default_diversifier(&self) -> Diversifier {
+        self.get(0u8)
+    }
+
+    /// Returns the diversifier at the given index.
+    pub fn get(&self, j: impl Into<DiversifierIndex>) -> Diversifier {
+        let ff = FF1::<Aes256>::new(&self.0, 2).expect("valid radix");
+        let enc = ff
+            .encrypt(&[], &BinaryNumeralString::from_bytes_le(&j.into().0[..]))
+            .unwrap();
```
dconnolly, line 200: Any chance this unwrap will panic?

src/keys.rs
```diff
@@ -1,70 +1,207 @@
...
+        Diversifier(enc.to_bytes_le().try_into().unwrap())
```
dconnolly, line 201: Same here

dconnolly:
> Perhaps instead we could add a disclaimer to the top-level crate documentation that we implicitly mean Orchard types here, not Sapling types?

Sounds good to me! Thanks!

dconnolly:
> But we could definitely add documentation tests that checks the code runs.

This would be great. Do you recall if CodeCov includes doc tests? I forget

src/address.rs
```diff
@@ -1,8 +1,23 @@
...
+/// let address = FullViewingKey::from(&sk).default_address();
+/// ```
 #[derive(Debug)]
 pub struct Address {
     d: Diversifier,
-    pk_d: (),
+    pk_d: DiversifiedTransmissionKey,
+}
+
+impl Address {
```
dconnolly, line 19: An idea: `From/Into`'s _alongside_ the more descriptive type impl methods defined, where the `From/Into`'s call those, so there is no duplication of logic. I push on this because of the orphan rule, otherwise we could just impl all these `std` traits ourselves: if they aren't defined in here, we'd have to wrap some `orchard` types to achieve them. :/

src/keys.rs
dconnolly: Oh what I mean is, will these `unwrap`s ever panic? If so, can we `match` on the Result of the `try_into`

src/keys.rs
```diff
@@ -79,18 +216,20 @@ pub struct Diversifier([u8; 11]);
...
 #[derive(Debug)]
-pub struct IncomingViewingKey;
+pub struct IncomingViewingKey(pallas::Scalar);
 
 impl From<&FullViewingKey> for IncomingViewingKey {
-    fn from(_: &FullViewingKey) -> Self {
-        todo!()
+    fn from(fvk: &FullViewingKey) -> Self {
+        let ak = pallas::Point::from_bytes(&(&fvk.ak.0).into()).unwrap();
+        IncomingViewingKey(commit_ivk(&extract_p(&ak), &fvk.nk.0, &fvk.rivk))
```
dconnolly, line 224:
```suggestion
        IncomingViewingKey(commit_ivk(&extract_p(&ak), &fvk.nk.0, &fvk.rivk.0))
```

src/spec.rs
```diff
@@ -0,0 +1,104 @@
...
+///
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
+pub(crate) fn to_scalar(hash: Hash) -> pallas::Scalar {
+    pallas::Scalar::from_bytes_wide(hash.as_array())
+}
+
+/// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
+///
+/// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
+pub(crate) fn commit_ivk(
```
dconnolly, line 38: Is there ever a use of `Commit^ivk` outside of `Commit^ivk_rivk`?

src/keys.rs
str4d: No; the only way this API errors is if the radix doesn't match the numeral string, and we initialize both the radix and the numeral string to binary.

src/keys.rs
str4d: This cannot panic because FF1 is format-preserving; thus given an 11-byte diversifier index, we get an 11-byte diversifier as intended.

src/address.rs
```diff
@@ -1,8 +1,23 @@
...
+/// let sk = SpendingKey::from_bytes([7; 32]).unwrap();
+/// let address = FullViewingKey::from(&sk).default_address();
+/// ```
 #[derive(Debug)]
 pub struct Address {
     d: Diversifier,
-    pk_d: (),
+    pk_d: DiversifiedTransmissionKey,
+}
+
```
dconnolly, line 18:
```rust
impl From<(Diversifier, DiversifiedTransmissionKey)> for Address {
    pub fn from((d, pk_d): (Diversifier, DiversifiedTransmissionKey)) -> Self {
        Self::from_parts(d, pk_d)
    }
}

```

src/lib.rs
str4d: Heh, this was originally a `util.rs` file, but with far less in it! I decided to extract more small functions to here and rename it, in order to make the key component derivation visibly closer to the spec.

I'm ambivalent on adding module structure below this. I don't anticipate this file growing much larger, and I'm not really a fan of single-function modules, but I'm not strongly against the idea.

src/keys.rs
dconnolly: I wish I could dismiss my own comments 🙊

src/keys.rs
dconnolly: Thanks!

src/keys.rs
str4d: Oh. No, these can never panic; the lengths of the slices are guaranteed to match the array lengths (`&[u8; 64][..32]` obviously has the same length as `[u8; 32]`, etc.)

src/lib.rs
dconnolly: Would there be a single-function module? I don't want that either

src/keys.rs
```diff
@@ -1,70 +1,207 @@
...
+            rivk: sk.into(),
+        }
     }
 }
 
 impl FullViewingKey {
+    /// Defined in [Zcash Protocol Spec § 4.2.3: Orchard Key Components][orchardkeycomponents].
+    ///
+    /// [orchardkeycomponents]: https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents
+    fn derive_dk_ovk(&self) -> (DiversifierKey, OutgoingViewingKey) {
```
dconnolly, line 137: Is this `ovk` derivation incoming in an update to the draft? I don't see a clear definition alá Sapling

str4d: It's in the current WIP, and also in the published ZIP 32.

src/address.rs
nuttycom: :+1: this is how I like to write typeclass instances in Haskell, and the analogy to traits is apt: write a function that is typed to whatever you're working on, and then use that function to provide for the generalized dispatch via the trait/typeclass.

src/keys.rs
nuttycom: Would it make sense to also support u128 and/or UUID (16 octets) as a source directly? If you want to support database identifiers, UUIDs are becoming pretty common.

src/keys.rs
daira: Use https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents for the spec link.

src/keys.rs
```diff
@@ -1,70 +1,207 @@
...
+            fn from(j: $n) -> Self {
+                let mut j_bytes = [0; 11];
+                j_bytes[..mem::size_of::<$n>()].copy_from_slice(&j.to_le_bytes());
+                DiversifierIndex(j_bytes)
+            }
+        }
+    };
+}
+di_from!(u8);
+di_from!(u16);
```
daira, line 184:
```suggestion
```

src/keys.rs
```diff
@@ -1,70 +1,207 @@
...
+di_from!(u32);
+di_from!(u64);
+di_from!(usize);
+
+impl DiversifierKey {
+    /// Returns the diversifier at index 0.
+    pub fn default_diversifier(&self) -> Diversifier {
+        self.get(0u8)
```
daira, line 192:
```suggestion
        self.get(0u32)
```

src/keys.rs
str4d: I started putting together a `zcash_address` crate which I think we could use as a common base for encodings. We'd `impl FromAddress for orchard::PaymentAddress`, for example (only defining the Orchard payment address conversion, for example), which would erase network-specific information where we don't need it, but the `zebra-chain` address type could retain it.

src/spec.rs
```diff
@@ -0,0 +1,104 @@
...
+    ak: &pallas::Base,
```
daira, line 39:
```suggestion
    ak: &pallas::Base,
```
In the spec this will be a point. (It's currently wrong in the draft.)

src/spec.rs
```diff
@@ -0,0 +1,104 @@
...
+    nk: &pallas::Base,
+    rivk: &pallas::Scalar,
+) -> pallas::Scalar {
+    let ivk = sinsemilla::short_commit(
+        "z.cash:Orchard-CommitIvk",
+        iter::empty()
+            .chain(ak.to_le_bits().iter().by_val().take(L_ORCHARD_SCALAR))
```
daira, line 46:
```suggestion
            .chain(ak.to_bytes().iter().flat_map(|b| (0..8).map(|i| (b >> i) & 1 != 0)))
```
Consider using a `bytes_to_bits` helper or the `bitvec` crate.

src/spec.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+/// Hash extractor for Pallas.
+///
+/// Defined in [Zcash Protocol Spec § 5.4.8.7: Hash Extractor for Pallas][concreteextractorpallas].
+///
+/// [concreteextractorpallas]: https://zips.z.cash/protocol/nu5.pdf#concreteextractorpallas
+pub(crate) fn extract_p(point: &pallas::Point) -> pallas::Base {
+    // TODO: Should we return the actual bits in a Vec, or allow the caller to use
+    // PrimeField::to_le_bits on the returned pallas::Base?
```
daira, line 98: Either works.

src/keys.rs
str4d: These would require a `TryFrom` implementation (because they might be out-of-range), so I'm less inclined to support them directly; in that case the user probably needs to define an explicit mapping from the UUID to a diversifier (since the UUID is likely random, and thus almost guaranteed to be out-of-range).

src/keys.rs
str4d: (The above crate is focused on addresses rather than keys, but we could maybe generalise it for keys in a similar fashion. I'm a bit on-the-fence about whether I want key encodings in an address-focused crate, but I also don't want to see everything become a mish-mash of tiny crates...)

src/spec.rs
str4d: No. But with function names in the protocol spec, the superscript is part of the function "name" while the subscript is used for "keyed on", thus the key is a parameter of the function.

str4d:
@dconnolly I want to expand on why I'm reluctant to add `From<(tuple)>` impls to all conversions. My general thinking with crate APIs is that it's much easier to expand them than reduce them. Thus I'm inclined to prefer more narrowly-defined API surfaces that are easier for downstream users to understand and operate safely, rather than defining every possible transformation up-front and expecting downstream users to know how to use them correctly.

In my view, this is especially problematic for `From<(tuple)>` interfaces, for two reasons:
- Documentation of `From` interfaces is completely collapsed by default; it's necessary to perform multiple clicks through the documentation in order to see anything, and even then users won't always go looking because they don't generally expect to see custom documentation there (since most trait impls just use the trait documentation). This means they will generally just rely on the types involved in the impl, but:
- Tuples are not meaningfully self-describing. `(Diversifier, DiversifiedTransmissionKey)` doesn't mean anything to someone without intimate knowledge of the Zcash Protocol Specification, any more than `(Diversifier, IncomingViewingKey)` does.

This is not simply an API usability issue either. For example, there are no protocol restrictions on mixing-and-matching `Diversifier`s with `IncomingViewingKey`s because it does not matter, but a user mixing-and-matching `Diversifier`s with `DiversifiedTransmissionKey`s can end up creating a payment address to which funds can be sent, but not received, which is a **loss  of user funds** bug.

- (In short: if `d` and `pk_d` don't match in an address, then the sender will encrypt to `[esk] pk_d` and send `epk = [esk] g_d`, but since `pk_d != [ivk] g_d`, the recipient can't decrypt with `[ivk] epk`, and thus will never see the funds without both deviating from the standard protocol _and_ cooperating with the sender. If the sender does not use `ovk` to make the output recoverable, this can result in complete loss of funds.)

src/keys.rs
```diff
@@ -60,21 +60,25 @@ impl From<&SpendingKey> for SpendAuthorizingKey {
         // TODO: Add TryFrom<S::Scalar> for SpendAuthorizingKey.
         let ret = SpendAuthorizingKey(ask.to_bytes().try_into().unwrap());
         // If the last bit of repr_P(ak) is 1, negate ask.
-        if (<[u8; 32]>::from(AuthorizingKey::from(&ret).0)[31] >> 7) == 1 {
+        if (<[u8; 32]>::from(SpendValidatingKey::from(&ret).0)[31] >> 7) == 1 {
```
dconnolly, line 63: 👍

str4d:
> This would be great. Do you recall if CodeCov includes doc tests? I forget

AFAIK CodeCov will display whatever is thrown at it, so it's a matter of sourcing. I don't believe `cargo-tarpaulin` does doctest coverage. I haven't looked into the newer Rust source coverage either. (In any case, there is now a doctest that exercises this PR).

src/address.rs
str4d: We discussed this in Discord and agreed this particular case wouldn't be included.

src/constants.rs
```diff
@@ -0,0 +1,4 @@
+//! Constants used in the Orchard protocol.
+
+/// $\ell^\mathsf{Orchard}_\mathsf{scalar}$
+pub(crate) const L_ORCHARD_SCALAR: usize = 255;
```
daira, line 4:
```suggestion
pub(crate) const L_ORCHARD_BASE: usize = 255;
```

src/spec.rs
daira: The way it is now is closer to the spec. (It changed from returning a bit sequence to a ℙ<sub>x</sub> when I added the ℙ<sub>x</sub> type.)

src/keys.rs
str4d: Unfortunately I can't define this anymore, because `IncomingViewingKey::address` now returns `Option<Address>` to match the protocol spec, and both `From` and `Option` are foreign types.

ebfull (approved):

src/spec.rs
```diff
@@ -0,0 +1,109 @@
...
+    pallas::Scalar::from_repr(ivk.to_repr()).unwrap()
+}
+
+/// Defined in [Zcash Protocol Spec § 5.4.1.6: DiversifyHash^Sapling and DiversifyHash^Orchard Hash Functions][concretediversifyhash].
+///
+/// [concretediversifyhash]: https://zips.z.cash/protocol/nu5.pdf#concretediversifyhash
+pub(crate) fn diversify_hash(d: &[u8; 11]) -> Option<pallas::Point> {
+    let pk_d = pallas::Point::hash_to_curve("z.cash:Orchard-gd")(d);
+    if pk_d.is_identity().into() {
+        None
```
str4d, line 64: In my pairing with @ebfull, we discussed this, and it occurred to me that it's possible we actually never hit this: once we fix Pallas and the hash-to-curve domain separator, the diversifier space is fixed for all spending keys. Thus any diversifier that resulted in the identity would do so for _all_ spending keys (though the spending key's diversifier key would alter which index that occured at, which could affect the default address).

We could in theory brute-force the 88-bit space to confirm that no diversifier will result in the identity. It's not practical to do so, but given that it's an 88-bit value being hashed into a 255-bit space, I think it's probably reasonable to just assert we never hit it. Thoughts @daira?

src/spec.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/// Defined in [Zcash Protocol Spec § 5.4.4.5: Orchard Key Agreement][concreteorchardkeyagreement].
+///
+/// [concreteorchardkeyagreement]: https://zips.z.cash/protocol/nu5.pdf#concreteorchardkeyagreement
+pub(crate) fn ka_orchard(sk: &pallas::Scalar, b: &pallas::Point) -> pallas::Point {
+    b * sk
+}
+
+/// Hash extractor for Pallas.
```
daira, line 98:
```suggestion
/// Coordinate extractor for Pallas.
```

src/spec.rs
```diff
@@ -0,0 +1,109 @@
...
+///
+/// Defined in [Zcash Protocol Spec § 5.4.8.7: Hash Extractor for Pallas][concreteextractorpallas].
```
daira, line 100:
```suggestion
/// Defined in [Zcash Protocol Spec § 5.4.8.7: Coordinate Extractor for Pallas][concreteextractorpallas].
```

src/primitives/sinsemilla.rs
```diff
@@ -65,27 +64,14 @@ impl<I: Iterator<Item = bool>> Iterator for Pad<I> {
     }
 }
```
daira, line 65: Please document what `Pad` is doing.

src/spec.rs
daira: Another thing we could do is that if the result of the group hash is the zero point, we just replace it with another arbitrary point, such as another group hash output or (-1, 2). I would mildly prefer that.

daira (approved):
utACK with comments.

ebfull:
Spec is being modified: when DiversifyHash ordinarily samples the identity, we substitute it with the output of `GroupHash("z.cash:Orchard-gd", "")`.

src/spec.rs
daira: image

str4d: The type of DiversifyHash should now be B -> P* (without the bottom).

src/spec.rs
```diff
@@ -107,3 +108,16 @@ pub(crate) fn extract_p(point: &pallas::Point) -> pallas::Base {
...
+
+#[cfg(test)]
+mod tests {
+    use group::Group;
+    use halo2::{arithmetic::CurveExt, pasta::pallas};
+
+    #[test]
+    fn diversify_hash_substitution() {
+        assert!(!bool::from(
+            pallas::Point::hash_to_curve("z.cash:Orchard-gd")(&[]).is_identity()
```
ebfull, line 120: haha thanks for this test, I was joking with @daira about this.

ebfull (approved):

src/primitives/sinsemilla.rs
```diff
@@ -98,16 +98,16 @@ pub(crate) fn hash_to_point(domain_prefix: &str, msg: impl Iterator<Item = bool>
...
 ///
-/// [§ 5.4.1.9]: https://zips.z.cash/protocol/nu5.pdf#concretesinsemillahash
+/// [concretesinsemillahash]: https://zips.z.cash/protocol/nu5.pdf#concretesinsemillahash
 pub(crate) fn hash(domain_prefix: &str, msg: impl Iterator<Item = bool>) -> pallas::Base {
-    extract(&hash_to_point(domain_prefix, msg))
+    extract_p(&hash_to_point(domain_prefix, msg))
 }
 
-/// `SinsemillaCommit` from [§ 5.4.7.4].
+/// `SinsemillaCommit` from [§ 5.4.7.4][concretesinsemillacommit].
```
daira, line 108:
```suggestion
/// `SinsemillaCommit` from [§ 5.4.8.4][concretesinsemillacommit].
```

src/primitives/sinsemilla.rs
```diff
@@ -122,15 +122,15 @@ pub(crate) fn commit(
     hash_to_point(&m_prefix, msg) + hasher_r(&[]) * r
 }
 
-/// `SinsemillaShortCommit` from [§ 5.4.7.4].
+/// `SinsemillaShortCommit` from [§ 5.4.7.4][concretesinsemillacommit].
```
daira, line 125:
```suggestion
/// `SinsemillaShortCommit` from [§ 5.4.8.4][concretesinsemillacommit].
```

src/spec.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/// Defined in [Zcash Protocol Spec § 5.4.1.6: DiversifyHash^Sapling and DiversifyHash^Orchard Hash Functions][concretediversifyhash].
+///
+/// [concretediversifyhash]: https://zips.z.cash/protocol/nu5.pdf#concretediversifyhash
+pub(crate) fn diversify_hash(d: &[u8; 11]) -> pallas::Point {
+    let pk_d = pallas::Point::hash_to_curve("z.cash:Orchard-gd")(d);
+    if pk_d.is_identity().into() {
+        // If the identity occurs, we replace it with a different fixed point.
+        pallas::Point::hash_to_curve("z.cash:Orchard-gd")(&[])
```
daira, line 65:
```suggestion
    let hasher = pallas::Point::hash_to_curve("z.cash:Orchard-gd");
    let pk_d = hasher(d);
    if pk_d.is_identity().into() {
        // If the identity occurs, we replace it with a different fixed point.
        hasher(&[])
```

daira (approved):
ACK with comments.

src/spec.rs
daira: Caught it in 4.1.1 but not in 5.4.1.6. That'll be in the next spec revision.

PR #41, opened 2021-03-11
Poseidon primitive

### LINKED ISSUE #37: Implement Poseidon outside the circuit
See zcash/zcash#4717 for optimized addition chains for the x^{1/5} maps. See zcash/zcash#4708 for Rescue parameters.

str4d:
The Orchard specification defines `PoseidonHash` in the simplest way as 𝑓([𝑥, 𝑦, 2<sup>65</sup>])[0], but this PR implements the duplex sponge explicitly to make the sponge construction clear.

This PR includes the following specification for Poseidon-128:
- S-box: x<sup>5</sup>
- t = 3
- R<sub>F</sub> = 8
- R<sub>P</sub> = 58 (one more than the equivalent standard specification).
- PoseidonHash(𝑥, 𝑦) =𝑓([𝑥, 𝑦, 2<sup>65</sup>])[0] (using 0-based indexing).

Closes #37.


src/primitives/poseidon.rs
```diff
@@ -0,0 +1,317 @@
...
+    }
+}
+
+impl<F: FieldExt> Spec<F> for P256Pow5T3<F> {
+    type State = [F; 3];
+    type Rate = [Option<F>; 2];
+
+    fn arity() -> usize {
+        3
+    }
```
str4d, line 118: I need to verify whether these are correct for Poseidon-256; the paper is very unclear as to how many state and capacity words are in each of its specifications.

str4d: The Orchard spec uses a rate of 2 and a capacity of 1, so we just need to figure out how that translates to their specifications in Table 1.

str4d: Turns out I should have been looking at Table 2 in section 4.

src/primitives/poseidon.rs
```diff
@@ -0,0 +1,361 @@
...
+
+    /// The type used to hold duplex sponge state.
+    ///
+    /// This must be an array of length equal to the rate of the duplex sponge (allowing
+    /// for a capacity consistent with this specification's security level), that defaults
+    /// to `[None; RATE]`.
+    type Rate: Default + AsRef<[Option<F>]> + AsMut<[Option<F>]>;
+
+    /// The arity of this specification.
+    fn arity() -> usize;
```
daira, line 29:
```suggestion
    /// The width of this specification.
    fn width() -> usize;
```
The [Poseidon paper](https://eprint.iacr.org/2019/458.pdf) uses "width" for this, not "arity". It uses "arity" only in the context of Merkle tree arity (which is not equal to the width, when Poseidon is used for a Merkle tree).

str4d: Yeah, I renamed this in the circuit PR, but forgot to do so here.

src/primitives/poseidon.rs
```diff
@@ -0,0 +1,361 @@
...
+    }
+}
+
+/// Runs the Poseidon permutation on the given state.
+fn permute<F: FieldExt, S: Spec<F>>(
+    state: &mut S::State,
+    mds: &[S::State],
+    round_constants: &[S::State],
+) {
+    // TODO: Check what should happen for odd number of full rounds.
```
str4d, line 102: The answer is that R<sub>F</sub> must always be even by construction; in the paper it's defined as R<sub>F</sub> = 2R<sub>f</sub>.

str4d:
Rebased on #40.

src/primitives/poseidon/nullifier.rs
```diff
@@ -0,0 +1,1585 @@
...
+    fn constants(&self) -> (Vec<Self::State>, Vec<Self::State>, Vec<Self::State>) {
+        let round_constants = ROUND_CONSTANTS[..].to_vec();
+        let mds = MDS[..].to_vec();
+        let mds_inv = MDS_INV[..].to_vec();
+
+        (round_constants, mds, mds_inv)
+    }
+}
+
+// $ sage generate_parameters_grain.sage 1 0 255 3 8 58 0x40000000000000000000000000000000224698fc094cf91b992d30ed00000001
```
ebfull, line 47: I did not check these constants.

ebfull (approved):

therealyingtong (approved):

PR #48, opened 2021-03-18
Use the pasta_curves crate directly

str4d:
This brings us current with the `halo2` crate (and also moves us to pinning its revision).


daira (approved):
utACK

PR #49, opened 2021-03-19
Bundle builder

str4d:
This is a modular bundle builder, the pieces of which can be used within a top-level transaction builder.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -119,13 +119,7 @@ impl ActionInfo {
     /// Defined in [Zcash Protocol Spec § 4.7.3: Sending Notes (Orchard)][orchardsend].
     ///
     /// [orchardsend]: https://zips.z.cash/protocol/nu5.pdf#orchardsend
-    fn build(
-        self,
-        mut rng: impl RngCore,
-    ) -> (
-        Action<(Option<SpendAuthorizingKey>, SpendValidatingKey)>,
-        Circuit,
-    ) {
+    fn build(self, mut rng: impl RngCore) -> (Action<SigningMetadata>, Circuit) {
         let v_net = self.value_sum().expect("already checked this");
         let cv_net = ValueCommitment::derive(v_net, self.rcv);
 
@@ -152,10 +146,10 @@ impl ActionInfo {
                 cm_new.into(),
                 encrypted_note,
                 cv_net,
-                (
-                    self.spend.dummy_sk.as_ref().map(SpendAuthorizingKey::from),
+                SigningMetadata {
+                    dummy_ask: self.spend.dummy_sk.as_ref().map(SpendAuthorizingKey::from),
                     ak,
-                ),
+                },
             ),
             Circuit {},
         )
@@ -326,7 +320,21 @@ pub struct Unauthorized {
 }
 
 impl Authorization for Unauthorized {
-    type SpendAuth = (Option<SpendAuthorizingKey>, SpendValidatingKey);
+    type SpendAuth = SigningMetadata;
+}
+
+/// Container for metadata needed to sign an [`Action`].
+#[derive(Debug)]
+pub struct SigningMetadata {
+    /// If this action is spending a dummy note, this field holds that note's spend
+    /// authorizing key.
+    ///
+    /// These keys are used automatically in [`Bundle<Unauthorized>::prepare`] or
+    /// [`Bundle<Unauthorized>::apply_signatures`] to sign dummy spends.
+    dummy_ask: Option<SpendAuthorizingKey>,
+    /// The spend validating key for this action. Used to match spend authorizing keys to
+    /// actions they can create signatures for.
+    ak: SpendValidatingKey,
 }
 
 /// Marker for a partially-authorized bundle, in the process of being signed.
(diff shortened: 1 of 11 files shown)


src/builder.rs
```diff
@@ -0,0 +1,472 @@
...
+    ///
+    /// [orcharddummynotes]: https://zips.z.cash/protocol/nu5.pdf#orcharddummynotes
+    fn dummy(rng: &mut impl RngCore) -> Self {
+        let fvk: FullViewingKey = (&SpendingKey::random(rng)).into();
+        let recipient = fvk.default_address();
+
+        RecipientInfo {
+            ovk: None,
+            recipient,
+            value: NoteValue::default(),
```
nuttycom, line 89: I'd prefer `NoteValue::zero()` here.

src/builder.rs
```diff
@@ -0,0 +1,472 @@
...
+    /// Builds the action.
+    ///
+    /// Defined in [Zcash Protocol Spec § 4.7.3: Sending Notes (Orchard)][orchardsend].
+    ///
+    /// [orchardsend]: https://zips.z.cash/protocol/nu5.pdf#orchardsend
+    fn build(
+        self,
+        mut rng: impl RngCore,
+    ) -> (
+        Action<(Option<SpendAuthorizingKey>, SpendValidatingKey)>,
```
nuttycom, line 126: It's not obvious that this `SpendAuthorizingKey` will only be present if it's a dummy value. This at least needs documentation, but I feel like it might be better served by a bespoke enum type.

src/builder.rs
```diff
@@ -0,0 +1,472 @@
...
+    fn build(
+        mut self,
+        mut rng: impl RngCore,
+        pk: &ProvingKey,
+    ) -> Result<Bundle<Unauthorized>, Error> {
+        // Pair up the spends and recipients, extending with dummy values as necessary.
+        //
+        // TODO: Do we want to shuffle the order like we do for Sapling? And if we do, do
+        // we need the extra logic for mapping the user-provided input order to the
+        // shuffled order?
```
nuttycom, line 251: Could the randomization be left to be a wallet-level concern? What would the randomization buy us here?

src/builder.rs
```diff
@@ -0,0 +1,472 @@
...
+        };
+
+        // Move some things out of self that we will need.
+        let flags = self.flags;
+        let anchor = self.anchor;
+
+        // Determine the value balance for this bundle, ensuring it is valid.
+        let value_balance: ValueSum = pre_actions
+            .iter()
+            .fold(Ok(ValueSum::default()), |acc, action| {
```
nuttycom, line 283: `ValueSum::zero()` please. :)

src/note.rs
```diff
@@ -60,6 +60,25 @@ pub struct Note {
...
+    pub(crate) fn new(
+        recipient: Address,
+        value: NoteValue,
+        rho: Nullifier,
+        mut rng: impl RngCore,
+    ) -> Self {
+        Note {
+            recipient,
+            value,
+            rho: rho,
```
nuttycom:
```suggestion
            rho,
```

src/builder.rs
```diff
@@ -0,0 +1,472 @@
...
+    pub fn prepare<R: rand_7::RngCore + rand_7::CryptoRng>(
+        self,
+        mut rng: R,
+        sighash: [u8; 32],
+    ) -> Bundle<PartiallyAuthorized> {
+        self.authorize(
+            &mut rng,
+            |rng, _, (dummy_ask, ak)| {
+                (
+                    if let Some(ask) = dummy_ask {
```
nuttycom, line 357: Is there a good reason to do part of the work (the signatures for the dummy spends) now, and then part of it later? Do we not have `ask` later?

nuttycom (approved):
This all makes sense. utACK with @therealyingtong 

NB: a couple of requests for explicit `.zero()` rather than `.default()`.

src/builder.rs
str4d: We need to do it in an automated fashion (as the exact number of dummies is opaque to the caller), so it's either going to be triggered by `prepare` or `finalize`. I didn't see any point in holding onto the dummy `ask` values beyond `prepare`, and signing at that point slots the signatures nicely into the spots where `sign` is about to place the real ones.

src/builder.rs
str4d: It can't quite be a wallet-level concern because the bundle builder encapsulates the logic for dummy notes, amongst which any real notes would be randomized.

What randomization removes is any information about what the position of a spend or output within an Action might imply. For example, if outputs are always in order, then if there are three actions in a bundle and you receive a new note as the output of the second action, then you can reasonably infer that the third action's output is most likely change to the sender, _and_ there was another recipient (in the first action's output). If outputs are randomized, then that is not distinguishable from the sender's wallet just performing note management and sending itself two notes.

src/builder.rs
nuttycom: Alternately: instead of a tuple, return a struct? I just want somewhere that the documentation for that first element can live.

src/builder.rs
```diff
@@ -0,0 +1,472 @@
...
+
+        let mut builder = Builder::new(Flags::from_parts(true, true), Anchor);
+        builder
+            .add_recipient(None, recipient, NoteValue::from_raw(5000), None)
+            .unwrap();
+        let bundle = dbg!(builder
+            .build(&mut rng, &pk)
+            .unwrap()
+            .prepare(rand_7::rngs::OsRng, [0; 32]))
+        .finalize()
```
nuttycom, line 468: Should this be calling `sign` before finalization?

ebfull (approved):

src/builder.rs
str4d: No, because this test has no real spent notes. Once we have note encryption support (and thus can get the created notes out of a builder), we can extend the tests with a case that exercises note spending.

PR #44, opened 2021-03-16
Note structure

str4d:
Closes #31.

PR #50, opened 2021-03-20
Add small key derivation benchmarks

### LINKED ISSUE #39: Fix CI
There is a test that seems to fail consistently while changes are merged, and the clippy checks are ignored. If the clippy checks are informational only, the status code of the checks can be changed to report the clippy output as annotations while registering as a green check.

str4d:
This will silence the benchmark CI failure, as we will finally have some benchmarks 😄

Closes #39.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,6 +8,8 @@
 //! address, which is also shielded).
 
 #![cfg_attr(docsrs, feature(doc_cfg))]
+// Temporary until we have more of the crate implemented.
+#![allow(dead_code)]
 // Catch documentation errors caused by code changes.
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -8,6 +8,8 @@
 //! address, which is also shielded).
 
 #![cfg_attr(docsrs, feature(doc_cfg))]
+// Temporary until we have more of the crate implemented.
+#![allow(dead_code)]
```
str4d, line 12: This change silences almost all the spurious warnings. The remainder are addressed by #44.

ebfull (approved):

PR #52, opened 2021-03-25
Refactor Poseidon primitive to use const generics

str4d:
MSRV is now 1.51.0 🎉


str4d:
Force-pushed to fix some clippy lints. The only remaining lints are fixed by #44.

daira (approved):
This is great! utACK

PR #54, opened 2021-03-29
[book] Document ECC gadget in circuit

therealyingtong:


book/src/design/circuit/gadgets/ecc/complete-add.md
```diff
@@ -0,0 +1,52 @@
+# Complete addition
```
str4d, line 1:
```suggestion
# Complete addition

To implement complete addition inside the circuit, we need to check the following cases:
```

book/src/design/circuit/gadgets/ecc/complete-add.md
```diff
@@ -0,0 +1,52 @@
...
+$\begin{array}{rcll}
+\mathcal{O} &+& \mathcal{O} &= \mathcal{O} ✓\\
+\mathcal{O} &+& (x_q, y_q)  &= (x_q, y_q) ✓\\
+ (x_p, y_p) &+& \mathcal{O} &= (x_p, y_p) ✓\\
+   (x, y)   &+& (x, y)      &= [2] (x, y) ✓\\
+   (x, y)   &+& (x, -y)     &= \mathcal{O} ✓\\
+ (x_p, y_p) &+& (x_q, y_q)  &= (x_p, y_p) \;⸭\; (x_q, y_q), \text{if } x_p \neq x_q ✓
+\end{array}$
+
+Suppose that we represent $\mathcal{O}$ as $(0, 0)$. ($0$ is not an $x$-coordinate of a valid point because we would need $y^2 = x^3 + 5$, and $5$ is not square in $\mathbb{F}_q$.)
```
str4d, line 12:
```suggestion
We represent $\mathcal{O}$ as $(0, 0)$.

> $0$ is not an $x$-coordinate of a valid point because we would need $y^2 = x^3 + 5$, and $5$ is not square in $\mathbb{F}_q$.
```

book/src/design/circuit/gadgets/ecc/complete-add.md
```diff
@@ -0,0 +1,52 @@
...
+\end{aligned}
+$$
+
+For the doubling case, $\lambda$ has to instead be computed as $\frac{3x^2}{2y}$.
+
+Witness $\lambda, \alpha, \beta, \gamma, \delta, A, B, C, D$.
+
+$
+\begin{array}{rcl|l}
+&&& Meaning \\\hline
```
str4d, line 30:
```suggestion
\begin{array}{rcl|rcl}
\text{Constraint} &&& \text{Meaning} \\\hline
```

book/src/design/circuit/gadgets/ecc/complete-add.md
```diff
@@ -0,0 +1,52 @@
...
+            C \cdot (1-C) &=& 0 & C \in \mathbb{B} \\
+            D \cdot (1-D) &=& 0 & D \in \mathbb{B} \\
+ (x_q - x_p) \cdot \alpha &=& 1-A & x_q = x_p \implies A \\
+          x_p \cdot \beta &=& 1-B & x_p = 0 \implies B \\
+              B \cdot x_p &=& 0 & B \implies x_p = 0 \\
+         x_q \cdot \gamma &=& 1-C & x_q = 0 \implies C \\
+              C \cdot x_q &=& 0 & C \implies x_q = 0 \\
+ (y_q + y_p) \cdot \delta &=& 1-D & y_q = -y_p \implies D \\
+(x_q - x_p) \cdot ((x_q - x_p) \cdot \lambda - (y_q - y_p)) &=& 0 & x_q \neq x_p \implies \lambda = \frac{y_q - y_p}{x_q - x_p} \\
+A \cdot \left(2y_p \cdot \lambda - 3{x_p}^2\right) &=& 0 & A \wedge y_p \neq 0 \implies \lambda = \frac{3{x_p}^2}{2y_p} \\
```
str4d, line 42:
```suggestion
 (x_q - x_p) \cdot \alpha &=& 1-A & x_q = x_p &\implies& A \\
          x_p \cdot \beta &=& 1-B & x_p = 0 &\implies& B \\
              B \cdot x_p &=& 0 & B &\implies& x_p = 0 \\
         x_q \cdot \gamma &=& 1-C & x_q = 0 &\implies& C \\
              C \cdot x_q &=& 0 & C &\implies& x_q = 0 \\
 (y_q + y_p) \cdot \delta &=& 1-D & y_q = -y_p &\implies& D \\
(x_q - x_p) \cdot ((x_q - x_p) \cdot \lambda - (y_q - y_p)) &=& 0 & x_q \neq x_p &\implies& \lambda = \frac{y_q - y_p}{x_q - x_p} \\
A \cdot \left(2y_p \cdot \lambda - 3{x_p}^2\right) &=& 0 & A \wedge y_p \neq 0 &\implies& \lambda = \frac{3{x_p}^2}{2y_p} \\
```

book/src/design/circuit/gadgets/ecc/complete-add.md
```diff
@@ -0,0 +1,52 @@
...
+(1-B) \cdot (1-C) \cdot (\lambda^2 - x_p - x_q - x_r) + B \cdot (x_r - x_q) &=& 0 & (¬B \wedge ¬C \implies x_r = \lambda^2 - x_p - x_q) \wedge (B \implies x_r = x_q) \\
+(1-B) \cdot (1-C) \cdot (\lambda \cdot (x_p - x_r) - y_p - y_r) + B \cdot (y_r - y_q) &=& 0 & (¬B \wedge ¬C \implies y_r = \lambda \cdot (x_p - x_r) - y_p) \wedge (B \implies y_r = y_q) \\
```
str4d, line 44:
```suggestion
\\
(1-B) \cdot (1-C) \cdot (\lambda^2 - x_p - x_q - x_r) && & (¬B \wedge ¬C &\implies& x_r = \lambda^2 - x_p - x_q) \\
+ B \cdot (x_r - x_q) &=& 0 & \wedge (B &\implies& x_r = x_q) \\
\\
(1-B) \cdot (1-C) \cdot (\lambda \cdot (x_p - x_r) - y_p - y_r) && & (¬B \wedge ¬C &\implies& y_r = \lambda \cdot (x_p - x_r) - y_p) \\
+ B \cdot (y_r - y_q) &=& 0 & \wedge (B &\implies& y_r = y_q) \\
\\
```

book/src/design/circuit/gadgets/ecc/complete-add.md
```diff
@@ -0,0 +1,52 @@
...
+      C \cdot (x_r - x_p) &=& 0 & C \implies x_r = x_p \\
+      C \cdot (y_r - y_p) &=& 0 & C \implies y_r = y_p \\
+              D \cdot x_r &=& 0 & D \implies x_r = 0 \\
+              D \cdot y_r &=& 0 & D \implies y_r = 0 \\
```
str4d, line 48:
```suggestion
      C \cdot (x_r - x_p) &=& 0 & C &\implies& x_r = x_p \\
      C \cdot (y_r - y_p) &=& 0 & C &\implies& y_r = y_p \\
              D \cdot x_r &=& 0 & D &\implies& x_r = 0 \\
              D \cdot y_r &=& 0 & D &\implies& y_r = 0 \\
```

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,102 @@
...
+```
+= $2^{n+1} + 2^n - 1$
+
+> The assertion labelled X obviously cannot fail because $u \neq 0$. It is possible to see that acc is monotonically increasing except in the last conditional. It reaches its largest value when $k$ is maximal, i.e. $2^{n+1} + 2^n - 1$.
+
+So to entirely avoid exceptional cases, we would need $2^{n+1} + 2^n - 1 < (q-1)/2$. But we can use $n$ larger by $c$ if the last $c$ iterations use [complete addition](./complete-add.md).
+
+The first $i$ for which the algorithm using **only** incomplete addition fails is going to be $252$, since $2^{252+1} + 2^{252} - 1 > (q - 1)/2$. We need $n = 254$ to make the wraparound technique above work.
+
+> sage: q = 0x40000000000000000000000000000000224698fc0994a8dd8c46eb2100000001
```
str4d, line 64:
```suggestion
```python
sage: q = 0x40000000000000000000000000000000224698fc0994a8dd8c46eb2100000001
```
and a corresponding ``` at the end (which I can't suggest in a GitHub comment).

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,102 @@
...
+
+We compute $\lambda_{1, i}, \lambda_{2, i}$:
+- $\lambda_{1, i} = \frac{y_{A,i} - y_{P, i}}{x_{A,i} - x_{P, i}},$
+- $\lambda_{2, i} = \frac{y_{A, i} + y_{A, i+1}}{x_{A,i} - x_{A, i+1}}$
+
+and similarly for $\lambda_{1, i+1}, \lambda_{2, i+1}$.
+
+We witness $x_{A,i}, x_{P,i}, x_{A, i+1},$ and $\lambda_{1, i}, \lambda_{2, i}, \lambda_{1, i+1}, \lambda_{2, i+1},$ and specify the following constraints on them (copied from ["Faster variable-base scalar multiplication in zk-SNARK circuits"](https://github.com/zcash/zcash/issues/3924), with some variable name changes):
+
+1. $
```
str4d, line 94:
```suggestion
$$
```

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,102 @@
...
+\lambda_{2,i}^2 - (x_{A,i+1} + (\lambda_{1,i}^2 - x_{A,i} - x_{P,i}) + x_{A,i}) = 0,
+$ and
```
str4d, line 96:
```suggestion
$$
```

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,102 @@
...
+
+2. $
+2 \cdot \lambda_{2,i} \cdot (x_{A,i} - x_{A,i+1}) - \big(
+        (\lambda_{1,i} + \lambda_{2,i}) \cdot (x_{A,i} - (\lambda_{1,i}^2 - x_{A,i} - x_{P,i})) +
+        (\lambda_{1,i+1} + \lambda_{2,i+1}) \cdot (x_{A,i+1} - (\lambda_{1,i+1}^2 - x_{A,i+1} - x_{P,i+1}))\big) = 0.
+$
```
str4d, line 102:
```suggestion
$$
\begin{aligned}
2 \cdot &\lambda_{2,i} \cdot (x_{A,i} - x_{A,i+1}) - \big( \\
    &\begin{aligned}
        (\lambda_{1,i} + \lambda_{2,i}) &\cdot (x_{A,i} - (\lambda_{1,i}^2 - x_{A,i} - x_{P,i})) + \\
        (\lambda_{1,i+1} + \lambda_{2,i+1}) &\cdot (x_{A,i+1} - (\lambda_{1,i+1}^2 - x_{A,i+1} - x_{P,i+1})) \\
    \end{aligned} \\
\big) &= 0.
\end{aligned}
$$
```

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -0,0 +1,66 @@
...
+- $\mathcal{V}$ and $\mathcal{R}$ bases for $\mathsf{ValueCommit}^{\mathsf{Orchard}}$; and
+- $\mathcal{R}$ base for $\mathsf{Commit}^{\mathsf{ivk}}$.
+
+## Witness scalar
+In most cases, we multiply the fixed bases by $255-$bit scalars from $\mathbb{F}_q$. We decompose a full-width scalar $\alpha$ into $85$ $3$-bit windows:
+
+$$\alpha = k_0 + k_1 \cdot (2^3)^1 + \cdots + k_{84} \cdot (2^3)^{84}, k_i \in [0..2^3).$$
+
+## Load fixed base
+Then, we precompute multiples of the fixed base $B$ for each window. This takes the form of a window table: $M[0..84][0..7]$ such that:
```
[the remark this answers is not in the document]
daira: `M[0..85)[0..8)` seems best to me.

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -0,0 +1,66 @@
...
+
+- for the first 84 rows $M[0..83][0..7]$: $M[w][k] = [(k+1) \cdot (2^3)^w]B$
+- in the last row $M[84][0..7]$: $M[w][k] = [k \cdot (2^3)^w - \sum\limits_{j=0}^{83} (2^3)^j]B$
+
+The additional $(k + 1)$ term lets us avoid adding the point at infinity in the case $k = 0$. We offset these accumulated terms by subtracting them in the final window $- \sum\limits_{j=0}^{83} (2^3)^j$.
```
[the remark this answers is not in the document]
daira:
```suggestion
The additional $(k + 1)$ term lets us avoid adding the point at infinity in the case $k = 0$. We offset these accumulated terms by subtracting them in the final window, i.e. we subtract $\sum\limits_{j=0}^{83} (2^3)^j$.
```

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
daira: I think Pygments supports `sage` as a language.

str4d: I tried that and didn't get any highlighting, so `mdbook` doesn't have that support yet.

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -1 +1,76 @@
...
+Inputs: $P = (x_P, y_P), Q = (x_Q, y_Q)$
+Output: $A = P + Q = (x_A, y_A)$
+
+Formulae:
+- $\lambda \cdot (x_p - x_{q}) = y_p - y_{q}$
+- $x_{a} = \lambda^2 - x_{q} - x_p$
+- $y_{a} = \lambda(x_{q} - x_{a}) - y_{q}$
+
+Substituting for $\lambda$, we get the constraints:
+- $(x_{a} + x_{q} + x_p) \cdot (x_p - x_q)^2 - (y_p - y_{q})^2 = 0$
```
str4d, line 13: This constraint is unsatisfiable for `P + (-P)`, and so we can't use it with arbitrary inputs (for the performance optimisation where ⊥ is represented either by unconstrained variables or arbitrary constraints). Let `(x_q, y_q) = (x_p, -y_p)` and `P != -P` (i.e. `y_p != 0`):
```
(x_a + x_p + x_p) * (x_p - x_p)^2 - (y_p + y_p)^2 = 0
(x_a + 2 x_p) * 0 - 4 y_p^2 = 0 <-- x_a is unconstrained, as expected
                  - 4 y_p^2 = 0
                      y_p   = 0 <-- contradiction
```
The other constraint is fine in this case:
```
(y_a - y_p)(x_p - x_p) - (y_p + y_p)(x_p - x_a) = 0
(y_a - y_p) * 0 - 2 y_p (x_p - x_a) = 0 <-- y_a is unconstrained, as expected
                - 2 y_p (x_p - x_a) = 0
                         x_p = x_a <-- x_a is constrained, but that's fine
```

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -1 +1,76 @@
...
+
+> $0$ is not an $x$-coordinate of a valid point because we would need $y^2 = x^3 + 5$, and $5$ is not square in $\mathbb{F}_q$.
+
+$$
+\begin{aligned}
+P + Q &= R\\
+(x_p, y_p) + (x_q, y_q) &= (x_r, y_r) \\
+                \lambda &= \frac{y_q - y_p}{x_q - x_p} \\
+                    x_r &= \lambda^2 - x_p - x_q \\
+                    y_r &= \lambda(x_p - x_r) - y_p
```
str4d, line 39: We should be consistent between incomplete addition and complete addition about the order of variables. It looks like `P` and `Q` are swapped here relative to the incomplete case above.

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -1 +1,76 @@
...
+\end{aligned}
+$$
+
+For the doubling case, $\lambda$ has to instead be computed as $\frac{3x^2}{2y}$.
+
+Witness $\lambda, \alpha, \beta, \gamma, \delta, A, B, C, D$.
```
str4d, line 45: If I'm reading the table correctly, we need to require that `\alpha, \beta, \gamma, \delta` are non-zero, otherwise the implied relations are only partial (e.g. `x_p = 0 \implies B` holds but `x_p != 0 \implies ¬B` does not, yet we rely on `¬B` later on). So we need additional constraints that require the prover to witness their inverses.

str4d (changes requested):
I haven't finished reviewing the complete addition table.

book/src/design/circuit/gadgets/ecc.md
daira: I will clarify this in the spec.

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -0,0 +1,76 @@
+# Elliptic Curve Cryptography
+
+## Incomplete addition
+Inputs: $P = (x_P, y_P), Q = (x_Q, y_Q)$
+Output: $A = P + Q = (x_A, y_A)$
+
+Formulae:
+- $\lambda \cdot (x_p - x_{q}) = y_p - y_{q}$
+- $x_{a} = \lambda^2 - x_{q} - x_p$
+- $y_{a} = \lambda(x_{q} - x_{a}) - y_{q}$
```
daira, line 10:
```suggestion
- $\lambda \cdot (x_p - x_q) = y_p - y_q$
- $x_a = \lambda^2 - x_q - x_p$
- $y_a = \lambda(x_q - x_a) - y_q$
```

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -0,0 +1,76 @@
...
+
+Substituting for $\lambda$, we get the constraints:
+- $(x_{a} + x_{q} + x_p) \cdot (x_p - x_q)^2 - (y_p - y_{q})^2 = 0$
+- $(y_{a} + y_{q})(x_p - x_{q}) - (y_p - y_{q})(x_{q} - x_{a}) = 0$
```
daira, line 14:
```suggestion
- $(x_a + x_q + x_p) \cdot (x_p - x_q)^2 - (y_p - y_q)^2 = 0$
- $(y_a + y_q)(x_p - x_q) - (y_p - y_q)(x_q - x_a) = 0$
```

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -0,0 +1,76 @@
+# Elliptic Curve Cryptography
+
+## Incomplete addition
+Inputs: $P = (x_P, y_P), Q = (x_Q, y_Q)$
+Output: $A = P + Q = (x_A, y_A)$
```
daira, line 5: Use R = (x_r, y_r).

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -0,0 +1,76 @@
...
+We represent $\mathcal{O}$ as $(0, 0)$.
+
+> $0$ is not an $x$-coordinate of a valid point because we would need $y^2 = x^3 + 5$, and $5$ is not square in $\mathbb{F}_q$.
+
+$$
+\begin{aligned}
+P + Q &= R\\
+(x_p, y_p) + (x_q, y_q) &= (x_r, y_r) \\
+                \lambda &= \frac{y_q - y_p}{x_q - x_p} \\
+                    x_r &= \lambda^2 - x_p - x_q \\
```
daira, line 38:
```suggestion
                \lambda &= \frac{y_p - y_q}{x_p - x_q} \\
                    x_r &= \lambda^2 - x_q - x_p \\
```

book/src/design/circuit/gadgets/ecc.md
str4d: @daira clarified in a pairing that it's the cross-interaction of the two B constraints that fully constraint the implications. For example, the contra-positive of the first constraint's implication `x_p == 0 ==> B` is `!B ==> x_p != 0`, which is the other half of the second constraint's implication.

book/src/design/circuit/gadgets/ecc.md
therealyingtong: I've changed this to `y_r &= \lambda(x_q - x_r) - y_q` which matches the incomplete case above.

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -0,0 +1,78 @@
+# Elliptic Curve Cryptography
+
+## Incomplete addition
+- Inputs: $P = (x_p, y_p), Q = (x_q, y_q)$
+- Output: $R = P + Q = (x_r, y_r)$
```
daira, line 5:
```suggestion
- Output: $R = P \;⸭\; Q = (x_r, y_r)$
```

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -0,0 +1,78 @@
...
++ B \cdot (y_r - y_q) &=& 0 & \wedge (B &\implies& y_r = y_q) \\
+\\
+      C \cdot (x_r - x_p) &=& 0 & C &\implies& x_r = x_p \\
+      C \cdot (y_r - y_p) &=& 0 & C &\implies& y_r = y_p \\
+              D \cdot x_r &=& 0 & D &\implies& x_r = 0 \\
+              D \cdot y_r &=& 0 & D &\implies& y_r = 0 \\
+\end{array}
+$
+
+Max degree: $4$
```
daira, line 78:
```suggestion
Max degree: $4$

Note: It is the cross-interaction of the two $B$ constraints that fully constrain
the implications. For example, the contrapositive of the first constraint's implication
$x_p = 0 \implies B$ is $¬B \implies x_p \neq 0$, which is the other half of the
second constraint's implication. The same applies to $C$.
```

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -0,0 +1,70 @@
...
+Witness the sign $s$ and magnitude $m$ such that
+$$
+s \in \{-1, 1\} \\
+m \in [0, 2^{64}) \\
+\mathsf{v^{old}} - \mathsf{v^{new}} = s \cdot m
+$$
+
+Then compute $P = [m] \mathcal{V}$, and conditionally negate $P$ using $(x, y) \mapsto (x, s \cdot y)$.
+
+We can reuse the window table from full-width fixed-base scalar multiplication, but with only $\mathsf{ceil}(64 / 3) = 22$ windows.
```
daira, line 70:
```suggestion
We compute the window table in a similar way to full-width fixed-base scalar multiplication,
but with only $\mathsf{ceil}(64 / 3) = 22$ windows.
```

daira (approved):
ACK, but I wrote much of this text originally, so it also needs some fresh eyes.

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -0,0 +1,84 @@
...
+              B \cdot x_p &=& 0 & B &\implies& x_p = 0 \\
+         x_q \cdot \gamma &=& 1-C & x_q = 0 &\implies& C \\
+              C \cdot x_q &=& 0 & C &\implies& x_q = 0 \\
+ (y_q + y_p) \cdot \delta &=& 1-D & y_q = -y_p &\implies& D \\
+(x_q - x_p) \cdot ((x_q - x_p) \cdot \lambda - (y_q - y_p)) &=& 0 & x_q \neq x_p &\implies& \lambda = \frac{y_q - y_p}{x_q - x_p} \\
+A \cdot \left(2y_p \cdot \lambda - 3{x_p}^2\right) &=& 0 & A \wedge y_p \neq 0 &\implies& \lambda = \frac{3{x_p}^2}{2y_p} \\
+\\
+(1-B) \cdot (1-C) \cdot (\lambda^2 - x_p - x_q - x_r) && & (¬B \wedge ¬C &\implies& x_r = \lambda^2 - x_p - x_q) \\
++ B \cdot (x_r - x_q) &=& 0 & \wedge (B &\implies& x_r = x_q) \\
+\\
```
therealyingtong, line 68: Do we need `(1-D)` here to handle the `(x_r, y_r) = (0,0)` case?

```suggestion
(1-B) \cdot (1-C) \cdot (1-D) \cdot (\lambda^2 - x_p - x_q - x_r) && & (¬B \wedge ¬C &\implies& x_r = \lambda^2 - x_p - x_q) \\
+ B \cdot (x_r - x_q) &=& 0 & \wedge (B &\implies& x_r = x_q) \\
\\
```

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -0,0 +1,84 @@
...
+(1-B) \cdot (1-C) \cdot (\lambda \cdot (x_p - x_r) - y_p - y_r) && & (¬B \wedge ¬C &\implies& y_r = \lambda \cdot (x_p - x_r) - y_p) \\
++ B \cdot (y_r - y_q) &=& 0 & \wedge (B &\implies& y_r = y_q) \\
+\\
```
therealyingtong, line 71: Similarly here.
```suggestion
(1-B) \cdot (1-C) \cdot (1-D) \cdot (\lambda \cdot (x_p - x_r) - y_p - y_r) && & (¬B \wedge ¬C &\implies& y_r = \lambda \cdot (x_p - x_r) - y_p) \\
+ B \cdot (y_r - y_q) &=& 0 & \wedge (B &\implies& y_r = y_q) \\
\\
```

book/src/design/circuit/gadgets/ecc.md
daira: See the hackmd. (1-D) is not quite correct because there exist points for which x_q ≠ x_p and y_q = -y_p.

book/src/design/circuit/gadgets/ecc.md
```diff
@@ -0,0 +1,84 @@
...
+              D \cdot y_r &=& 0 & D &\implies& y_r = 0 \\
+\end{array}
+$
+
+Max degree: $4$
+
+Note: It is the cross-interaction of the two $B$ constraints that fully constrain
+the implications. For example, the contrapositive of the first constraint's implication
+$x_p = 0 \implies B$ is $¬B \implies x_p \neq 0$, which is the other half of the
+second constraint's implication. The same applies to $C$.
```
daira, line 84: This comment no longer applies for the optimized constraints in the hackmd.

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,175 @@
...
+>     if k_0 = 0 {
+>         assert acc ≠ 1
+>         acc := acc - 1
+>     }
+
+The maximum value of $acc$ is:
+```
+    <--n 1s--->
+  1011111111111
+= 1100000000000 - 1
```
daira, line 54: Add dots here. (n is not actually 11.)

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -14,17 +14,17 @@ $$\alpha = k_0 + k_1 \cdot (2^3)^1 + \cdots + k_{84} \cdot (2^3)^{84}, k_i \in [
 ## Load fixed base
 Then, we precompute multiples of the fixed base $B$ for each window. This takes the form of a window table: $M[0..85)[0..8)$ such that:
 
-- for the first 84 rows $M[0..84)[0..8)$: $$M[w][k] = [(k+1) \cdot (2^3)^w]B$$
-- in the last row $M[84][0..8)$: $$M[w][k] = [k \cdot (2^3)^w - \sum\limits_{j=0}^{83} (2^3)^j]B$$
+- for the first 84 rows $M[0..84)[0..8)$: $$M[w][k] = [(k+2) \cdot (2^3)^w]B$$
+- in the last row $M[84][0..8)$: $$M[w][k] = [k \cdot (2^3)^w - \sum\limits_{j=0}^{83} 2^{3j+1}]B$$
 
-The additional $(k + 1)$ term lets us avoid adding the point at infinity in the case $k = 0$. We offset these accumulated terms by subtracting them in the final window, i.e. we subtract $\sum\limits_{j=0}^{83} (2^3)^j$.
+The additional $(k + 2)$ term lets us avoid adding the point at infinity in the case $k = 0$. We offset these accumulated terms by subtracting them in the final window, i.e. we subtract $\sum\limits_{j=0}^{83} 2^{3j+1}$.
```
therealyingtong, line 20: Explain why `k+2` is needed and `k+1` is not sufficient

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -0,0 +1,82 @@
...
+- in the last row $M[84][0..8)$: $$M[w][k] = [k \cdot (2^3)^w - \sum\limits_{j=0}^{83} 2^{3j+1}]B$$
+
+The additional $(k + 2)$ term lets us avoid adding the point at infinity in the case $k = 0$. We offset these accumulated terms by subtracting them in the final window, i.e. we subtract $\sum\limits_{j=0}^{83} 2^{3j+1}$.
+
+> Note: Although an offset of $(k + 1)$ would naively suffice, it introduces an edge case when $k_0 = 7, k_1= 0$.
+> In this case, the window table entries evaluate to the same point:
+> * $M[0][k_0] = [(7+1)*(2^3)^0]B = [8]B,$
+> * $M[1][k_1] = [(0+1)*(2^3)^1]B = [8]B.$
+>
+> In fixed-base scalar multiplication, we summing the multiples of $B$ at each window using incomplete addition.
```
daira, line 27:
```suggestion
> In fixed-base scalar multiplication, we sum the multiples of $B$ at each window using incomplete addition.
```

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -0,0 +1,82 @@
...
+Repeating this for all $85$ windows, we end up with:
+- an $85 \times 8$ table $\mathcal{L}_x$ storing $8$ coefficients interpolating the $x-$coordinate for each window. Each $x$-coordinate interpolation polynomial will be of the form
+$$\mathcal{L}_x[w](k) = c_0 + c_1 \cdot k + c_2 \cdot k^2 + \cdots + c_7 \cdot k^7,$$
+where $k \in [0..8), w \in [0..85)$ and $c_k$'s are the coefficients for each power of $k$; and
+- a length-$85$ array $Z$ of $z_w$'s.
+
+We load these precomputed values into fixed columns whenever we do fixed-base scalar multiplication in the circuit.
+
+## Fixed-base scalar multiplication
+Given a decomposed scalar $\alpha$ and a fixed base $B$, we compute $[\alpha]B$ as such:
```
daira, line 49:
```suggestion
Given a decomposed scalar $\alpha$ and a fixed base $B$, we compute $[\alpha]B$ as follows:
```

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -0,0 +1,82 @@
...
+
+1. For each $k_w, w \in [0..85), k_w \in [0..8)$ in the scalar decomposition, witness the $x$- and $y$-coordinates $(x_w,y_w) = M[w][k_w].$
+2. Check that $(x_w, y_w)$ is on the curve: $y_w^2 = x_w^3 + b$.
+3. Witness $u_w$ such that $y_w + z_w = u_w^2$.
+4. For all windows but the last, use [incomplete addition](./incomplete-add.md) to sum the $M[w][k_w]$'s, resulting in $[\alpha]B$.
+5. For the last window, use complete addition $M[83][k_{83}] + M[84][k_{84}]$ and return the final result.
+
+> Note: complete addition is required in the final step to correctly map $[0]B$ to a representation of the point at infinity, $(0,0)$.
```
daira, line 57:
```suggestion
> Note: complete addition is required in the final step to correctly map $[0]B$ to a representation of the point at infinity, $(0,0)$; and also to handle a corner case for which the last step is a doubling.
```

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,175 @@
...
+\end{array}$
+
+provided that $\alpha + t_q \in [0, 2^{n+1})$, i.e. $\alpha < 2^{n+1} - t_q$ which covers the whole range we need because in fact $2^{255} - t_q > q$.
+
+Thus, given a scalar $\alpha$, we witness the boolean decomposition of $k = \alpha + t_q.$ (We use big-endian bit order for convenient input into the variable-base scalar multiplication algorithm.)
+
+$$k = k_{254} \cdot 2^{254} + k_{253} \cdot 2^{253} + \cdots + k_0.$$
+
+## Variable-base scalar multiplication
+We use an optimized double-and-add algorithm is (copied from ["Faster variable-base scalar multiplication in zk-SNARK circuits"](https://github.com/zcash/zcash/issues/3924), with some variable name changes):
```
daira, line 22:
```suggestion
We use an optimized double-and-add algorithm, copied from ["Faster variable-base scalar multiplication in zk-SNARK circuits"](https://github.com/zcash/zcash/issues/3924) with some variable name changes:
```

book/src/design/circuit/gadgets/ecc/addition.md
```diff
@@ -0,0 +1,337 @@
...
+
+## Complete addition
+
+$\begin{array}{rcll}
+\mathcal{O} &+& \mathcal{O} &= \mathcal{O} ✓\\
+\mathcal{O} &+& (x_q, y_q)  &= (x_q, y_q) ✓\\
+ (x_p, y_p) &+& \mathcal{O} &= (x_p, y_p) ✓\\
+   (x, y)   &+& (x, y)      &= [2] (x, y) ✓\\
+   (x, y)   &+& (x, -y)     &= \mathcal{O} ✓\\
+ (x_p, y_p) &+& (x_q, y_q)  &= (x_p, y_p) \;⸭\; (x_q, y_q), \text{if } x_p \neq x_q ✓
```
daira, line 25: The ✓ marks were just added for me to keep track of which cases had been checked; they're superfluous now.
```suggestion
\mathcal{O} &+& \mathcal{O} &= \mathcal{O} \\
\mathcal{O} &+& (x_q, y_q)  &= (x_q, y_q) \\
 (x_p, y_p) &+& \mathcal{O} &= (x_p, y_p) \\
   (x, y)   &+& (x, y)      &= [2] (x, y) \\
   (x, y)   &+& (x, -y)     &= \mathcal{O} \\
 (x_p, y_p) &+& (x_q, y_q)  &= (x_p, y_p) \;⸭\; (x_q, y_q), \text{if } x_p \neq x_q
```

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,258 @@
...
+
+for $i$ from $254$ down to $4$:
+$$
+\begin{aligned}
+    &(\mathbf{k}_i)(\mathbf{k}_i-1) = 0\\
+    &\mathbf{z}_{i} = 2\mathbf{z}_{i+1} + \mathbf{k}_{i}\\
+    & x_{U,i} = x_T\\
+    & y_{U,i} = (2 \mathbf{k}_i - 1) \cdot y_T  \hspace{2em}\text{(conditionally negate)}\\
+    & \lambda_{1,i} \cdot (x_{A,i} - x_{U,i}) = y_{A,i} - y_{U,i}\\
+    & \lambda_{1,i}^2 = x_{R,i} + x_{A,i} + x_{U,i}\\
```
str4d, line 112: Everywhere else on this page we use P instead of U:
```suggestion
    & x_{P,i} = x_T\\
    & y_{P,i} = (2 \mathbf{k}_i - 1) \cdot y_T  \hspace{2em}\text{(conditionally negate)}\\
    & \lambda_{1,i} \cdot (x_{A,i} - x_{P,i}) = y_{A,i} - y_{P,i}\\
    & \lambda_{1,i}^2 = x_{R,i} + x_{A,i} + x_{P,i}\\
```

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,258 @@
...
+    & (\lambda_{1,i} + \lambda_{2,i}) \cdot (x_{A,i} - x_{R,i}) = 2 y_{\mathsf{A},i}\\
+    & \lambda_{2,i}^2 = x_{A,i-1} + x_{R,i} + x_{A,i}\\
+    & \lambda_{2,i} \cdot (x_{A,i} - x_{A,i-1}) = y_{A,i} + y_{A,i-1}\\
+
+\end{aligned}
+$$
+
+After substitution of $y_{P,i}$, $x_{R,i}$, $y_{A,i}$, and $y_{A,i+1}$, this becomes:
```
str4d, line 120:
```suggestion
After substitution of $x_{P,i}$, $y_{P,i}$, $x_{R,i}$, $y_{A,i}$, and $y_{A,i+1}$, this becomes:
```

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,258 @@
...
+    x_T   &  y_T   &    \mathbf{z}_{129}       &   x_{A,129}   & \lambda_{1,129}  & \lambda_{2,129}      &   \mathbf{z}_3        & x_{A,3}     & \lambda_{1,3}       & \lambda_{2,3}         \\\hline
+\end{array}
+$$
+
+For each $hi$ and $lo$ half, the following constraints are enforced:
+$$
+\begin{array}{|c|l|}
+\hline
+\text{Degree} & \text{Constraint} \\\hline
+2   & q_{ECC,i} \cdot (\mathbf{z}_{i+1} - 2\mathbf{z}_i - \mathbf{k}_i) = 0 \\\hline
```
str4d, line 171: This isn't a constraint we can enforce (since we define `k_i` in terms of the other two, this will always be satisfied).
```suggestion
```

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,258 @@
...
+3 & q_{ECC,i} \cdot \mathbf{k}_i \cdot (\mathbf{k}_i - 1) = 0 \\\hline
+4 & q_{ECC,i} \cdot \left(\lambda_{1,i} \cdot (x_{A,i} - x_{P,i}) - y_{A,i} + (2\mathbf{k}_i - 1) \cdot y_{P,i}\right) = 0 \\\hline
+4 & q_{ECC,i} \cdot \left((\lambda_{1,i} + \lambda_{2,i}) \cdot (x_{A,i} - (\lambda_{1,i}^2 - x_{A,i} - x_{P,i})) - 2 y_{A,i}\right) = 0 \\\hline
```
str4d, line 174: Ditto (this is a rearrangement of the definition of `y_{A,i}`).
```suggestion
```

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,258 @@
...
+3 & q_{ECC,i} \cdot \left(\lambda_{2,i}^2 - x_{A,i+1} - (\lambda_{1,i}^2 - x_{A,i} - x_{P,i}) - x_{A,i}\right) = 0 \\\hline
+3   & q_{ECC,i} \cdot \left(\lambda_{2,i} \cdot (x_{A,i} - x_{A,i+1}) - y_{A,i} - y_{A,i+1}\right) = 0 \\\hline
```
str4d, line 176: This constraint needs to be modified similar to the Sinsemilla gate:
- On all but the last round of incomplete addition, `y_{A,i+1}` should be substituted with its representation in terms of `lambda_{1,i+1}` etc.
- On the last round of incomplete addition, `y_{A,i+1}` should be a queried cell.

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,258 @@
...
+    & (\lambda_{1,i} + \lambda_{2,i}) \cdot (x_{A,i} - x_{R,i}) = 2 y_{\mathsf{A},i}\\
+    & \lambda_{2,i}^2 = x_{A,i-1} + x_{R,i} + x_{A,i}\\
+    & \lambda_{2,i} \cdot (x_{A,i} - x_{A,i-1}) = y_{A,i} + y_{A,i-1}\\
+
+\end{aligned}
+$$
+
+After substitution of $y_{P,i}$, $x_{R,i}$, $y_{A,i}$, and $y_{A,i+1}$, this becomes:
+
+Initialize $A_{254} = [2] T$
```
str4d, line 122: The `y_A` component of this is missing. We need a constraint that binds `y_A` to `lambda_{1,254}` etc.

str4d (changes requested):

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,258 @@
...
+
+The maximum value of `acc` is:
+```ignore
+    <--- n 1s --->
+  1011111...111111
+= 1100000...000000 - 1
+```
+= $2^{n+1} + 2^n - 1$
+
+> The assertion labelled X obviously cannot fail because $u \neq 0$. It is possible to see that acc is monotonically increasing except in the last conditional. It reaches its largest value when $k$ is maximal, i.e. $2^{n+1} + 2^n - 1$.
```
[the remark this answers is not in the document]
therealyingtong: Yes, `u` -> `p` here.

str4d (commented):
Improvements to overflow check formatting.

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,309 @@
...
+
+Then the needed gates are:
+
+$s = \alpha + \mathbf{k}_{254} \cdot 2^{130}$
+$\mathbf{z}_0 = \alpha + t_q \pmod{p}$
+$\mathbf{k}_{254} \cdot (\mathbf{z}_{130} - 2^{124}) = 0$
+$\mathbf{k}_{254} \cdot (s - \sum\limits_{i=0}^{129} 2^i \cdot \mathbf{s}_i) = 0$
+$(1 - \mathbf{k}_{254}) \cdot (1 - \mathbf{z}_{130} \cdot \eta) \cdot (s - \sum\limits_{i=0}^{129} 2^i \cdot \mathbf{s}_i) = 0$
+
+where $\sum\limits_{i=0}^{129} 2^i \cdot \mathbf{s}_i$ can be computed by another running sum. [TODO: be explicit about how to do that.]
```
str4d, line 309: Fill this in. Specifically, note that what we get from the running sum is
```
(s - \sum\limits_{i=0}^{129} 2^i \cdot \mathbf{s}_i) / 2^{130}
```
and that this is fine to use in the constraints because constant factors don't matter in multiplication terms.

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -0,0 +1,83 @@
...
+$$
+s \in \{-1, 1\} \\
+m \in [0, 2^{64}) \\
+\mathsf{v^{old}} - \mathsf{v^{new}} = s \cdot m
+$$
+
+Then compute $P = [m] \mathcal{V}$, and conditionally negate $P$ using $(x, y) \mapsto (x, s \cdot y)$.
+
+We compute the window table in a similar way to full-width fixed-base scalar multiplication,
+but with only $\mathsf{ceil}(64 / 3) = 22$ windows.
```
str4d, line 83:
```suggestion
but with only $\mathsf{ceil}(64 / 3) = 22$ windows. The first 21 windows are full

$$
\begin{array}{|c|l|}
\hline
\text{Degree} & \text{Constraint} \\\hline
9 & \text{21 3-bit constraints} \\\hline
3 & q_\text{scalar-fixed-short} \cdot \left(\mathbf{m}_0 \cdot (1 - \mathbf{m}_0)\right) = 0 \\\hline
3 & q_\text{scalar-fixed-short} \cdot \left(s^2 - 1\right) = 0 \\\hline
\end{array}
$$
```

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -0,0 +1,83 @@
...
+3. Witness $u_w$ such that $y_w + z_w = u_w^2$.
+4. For all windows but the last, use [incomplete addition](./incomplete-add.md) to sum the $M[w][k_w]$'s, resulting in $[\alpha - k_{84} \cdot (2^3)^{84} + \sum\limits_{j=0}^{83} 2^{3j+1}]B$.
+5. For the last window, use complete addition $M[83][k_{83}] + M[84][k_{84}]$ and return the final result.
+
+> Note: complete addition is required in the final step to correctly map $[0]B$ to a representation of the point at infinity, $(0,0)$; and also to handle a corner case for which the last step is a doubling.
+
+Constraints:
+ - $x_w = \mathcal{L}_x[w](k_w)$;
+ - $y_w^2 = x_w^3 + b,$ where $b = 5$ (from the Pallas curve equation);
+ - $u_w^2 = y_w + Z[w].$
```
str4d, line 63:
```suggestion$$
\begin{array}{|c|l|}
\hline
\text{Degree} & \text{Constraint} \\\hline
8 & q_\text{mul-fixed} \cdot \left( \mathcal{L}_x[w](k_w) - x_w \right) = 0 \\\hline
4 & q_\text{mul-fixed} \cdot \left( y_w^2 - x_w^3 - b \right) = 0 \\\hline
3 & q_\text{mul-fixed} \cdot \left( u_w^2 - y_w - Z[w] \right) = 0 \\\hline
\end{array}
$$

where $b = 5$ (from the Pallas curve equation).
```

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -0,0 +1,83 @@
...
+- $\mathcal{K}^{\mathsf{Orchard}}$, used in deriving the nullifier;
+- $\mathcal{G}^{\mathsf{Orchard}}$, used in spend authorization;
+- $\mathcal{R}$ base for $\mathsf{NoteCommit}^{\mathsf{Orchard}}$;
+- $\mathcal{V}$ and $\mathcal{R}$ bases for $\mathsf{ValueCommit}^{\mathsf{Orchard}}$; and
+- $\mathcal{R}$ base for $\mathsf{Commit}^{\mathsf{ivk}}$.
+
+## Witness scalar
+In most cases, we multiply the fixed bases by $255-$bit scalars from $\mathbb{F}_q$. We decompose a full-width scalar $\alpha$ into $85$ $3$-bit windows:
+
+$$\alpha = k_0 + k_1 \cdot (2^3)^1 + \cdots + k_{84} \cdot (2^3)^{84}, k_i \in [0..2^3).$$
```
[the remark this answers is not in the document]
daira: This doesn't seem correct to me. \alpha actually does have to be canonical, because if it's only correct modulo p, that is different from it being correct modulo q (i.e. in the scalar field).

book/src/design/circuit/gadgets/ecc/addition.md
```diff
@@ -0,0 +1,337 @@
+## Incomplete addition
+- Inputs: $P = (x_p, y_p), Q = (x_q, y_q)$
+- Output: $R = P \;⸭\; Q = (x_r, y_r)$
+
+Formulae:
+- $\lambda \cdot (x_p - x_q) = y_p - y_q$
+- $x_r = \lambda^2 - x_q - x_p$
+- $y_r = \lambda(x_q - x_r) - y_q$
```
daira, line 8:
```suggestion
```suggestion
We will use formulae for curve arithmetic using affine coordinates on short Weierstrass curves,
derived from section 4.1 of [Hüseyin Hışıl's thesis](https://core.ac.uk/download/pdf/10898289.pdf).

## Incomplete addition

- Inputs: $P = (x_p, y_p), Q = (x_q, y_q)$
- Output: $R = P \;⸭\; Q = (x_r, y_r)$

The formulae from Hışıl's thesis are:

- $x_3 = \left(\frac{y_1 - y_2}{x_1 - x_2}\right)^2 - x_1 - x_2$
- $y_3 = \frac{y_1 - y_2}{x_1 - x_2} \cdot (x_1 - x_3) - y_1$

Rename:
- $(x_1, y_1)$ to $(x_q, y_q)$
- $(x_2, y_2)$ to $(x_p, y_p)$
- $(x_3, y_3)$ to $(x_r, y_r)$.

Let $\lambda = \frac{y_q - y_p}{x_q - x_p} = \frac{y_p - y_q}{x_p - x_q}$, which we implement as

$\lambda \cdot (x_p - x_q) = y_p - y_q$

Also,
- $x_r = \lambda^2 - x_q - x_p$
- $y_r = \lambda \cdot (x_q - x_r) - y_q$

which is equivalent to

- $x_r + x_q + x_p = \lambda^2$

Assuming $x_p \neq x_q$,

\begin{array}{|rrll|}
\hline
&(x_r + x_q + x_p) \cdot (x_p - x_q)^2 &=& \lambda^2 \cdot (x_p - x_q)^2\\
\implies &(x_r + x_q + x_p) \cdot (x_p - x_q)^2 &=& (\lambda \cdot (x_p - x_q))^2\\
\\\hline
         &y_r &=& \lambda \cdot (x_q - x_r) - y_q\\
\implies &y_r + y_q &=& \lambda \cdot (x_q - x_r)\\
\implies &(y_r + y_q) \cdot (x_p - x_q) &=& \lambda \cdot (x_p - x_q) \cdot (x_q - x_r)\\
\hline
\end{array}
```

book/src/design/circuit/gadgets/ecc/addition.md
```diff
@@ -0,0 +1,337 @@
...
+\begin{aligned}
+P + Q &= R\\
+(x_p, y_p) + (x_q, y_q) &= (x_r, y_r) \\
+                \lambda &= \frac{y_q - y_p}{x_q - x_p} \\
+                    x_r &= \lambda^2 - x_p - x_q \\
+                    y_r &= \lambda(x_p - x_r) - y_p
+\end{aligned}
+$$
+
+For the doubling case, $\lambda$ has to instead be computed as $\frac{3x^2}{2y}$.
```
daira, line 40:
```suggestion
For the doubling case, Hışıl's thesis tells us that $\lambda$ has to
instead be computed as $\frac{3x^2}{2y}$.
```

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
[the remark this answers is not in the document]
str4d: My point is that we never need to represent `alpha` as a base field element, or compute with it modulo p; it is witnessed purely as bits, and the fixed-base scalar mul will calculate the correct result whether or not `alpha` is canonical modulo q. If you are assuming something different, please say so!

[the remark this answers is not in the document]
str4d: If necessary, my suggestion can be clarified thusly:
```
$\alpha$ is allowed to be non-canonical modulo $q$; ...
```

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -0,0 +1,176 @@
...
+
+At the point of witnessing the scalar, we range-constrain each $3$-bit word of its decomposition.
+$$
+\begin{array}{|c|l|}
+\hline
+\text{Degree} & \text{Constraint} \\\hline
+9 & q_\text{witness-scalar-fixed} \cdot \texttt{range\_check}(k_i, 8) = 0 \\\hline
+\end{array}
+$$
+where $\texttt{range\_check}(k_i, \texttt{range}) = k_i \cdot (1 - k_i) \cdots (\texttt{range} - 1 - k_i).$
```
str4d, line 23: This isn't what's actually implemented. The implementation uses `Iterator::fold(1, ...)` which introduces a (superfluous) constant (which is fixed into the constraint polynomial). However, I just noticed that Rust 1.51 stabilised [`Iterator::reduce`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.reduce) which is the same as `fold` but uses the first element of the iterator as the initial value. We should check through the chip impls to use this where it makes sense, such as here (which would bring the impl back in line with this).

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -0,0 +1,176 @@
...
+
+To do this, we decompose $\alpha$ into three pieces: $$\alpha = \alpha_0 \text{ (252 bits) }  || \alpha_1 \text{ (2 bits) } || \alpha_2 \text{ (1 bit) }.$$
+
+We check the correctness of this decomposition by:
+$$
+\begin{array}{|c|l|}
+\hline
+\text{Degree} & \text{Constraint} \\\hline
+5 & q_\text{canon-base-field} \cdot \texttt{range\_check}(\alpha_1), 4) = 0 \\\hline
+3 & q_\text{canon-base-field} \cdot \texttt{range\_check}(\alpha_2), 2) = 0 \\\hline
```
str4d, line 157:
```suggestion
5 & q_\text{canon-base-field} \cdot \texttt{range\_check}(\alpha_1, 4) = 0 \\\hline
3 & q_\text{canon-base-field} \cdot \texttt{range\_check}(\alpha_2, 2) = 0 \\\hline
```

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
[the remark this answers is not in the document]
daira: I think saying "$\alpha$ is allowed to be non-canonical" does more harm than good. We do need the scalar used in nullifier computation to be canonical, otherwise there would be a double-spend bug! How about this:

```suggestion
$$\alpha = k_0 + k_1 \cdot (2^3)^1 + \cdots + k_{84} \cdot (2^3)^{84}, k_i \in [0..2^3).$$

The scalar multiplication will be computed correctly for $k_{0..84}$ representing any integer in the range $[0, 2^{255})$. If $k_{0..84}$ is witnessed directly then no issue of canonicity arises. If the scalar is given as a base field element, then care must be taken to ensure a canonical representation, since $2^{255} > p$. This occurs, for example, in the scalar multiplication for the nullifier computation of the Action circuit.

$$
\begin{array}{|c|l|}
\hline
\text{Degree} & \text{Constraint} \\\hline
9 & q_\text{scalar-fixed} \cdot 1 \cdot \left(\sum\limits_{i=0}^7{w - i}\right) = 0 \\\hline
\end{array}
$$
```

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
daira: Yes we should change the implementation, not the doc.

book/src/design/circuit/gadgets/ecc/addition.md
```diff
@@ -0,0 +1,337 @@
+## Incomplete addition
+- Inputs: $P = (x_p, y_p), Q = (x_q, y_q)$
+- Output: $R = P \;⸭\; Q = (x_r, y_r)$
+
+Formulae:
+- $\lambda \cdot (x_p - x_q) = y_p - y_q$
+- $x_r = \lambda^2 - x_q - x_p$
+- $y_r = \lambda(x_q - x_r) - y_q$
+
+Substituting for $\lambda$, we get the constraints:
```
daira, line 10:
```suggestion
We will use formulae for curve arithmetic using affine coordinates on short Weierstrass curves,
derived from section 4.1 of [Hüseyin Hışıl's thesis](https://core.ac.uk/download/pdf/10898289.pdf).

## Incomplete addition

- Inputs: $P = (x_p, y_p), Q = (x_q, y_q)$
- Output: $R = P \;⸭\; Q = (x_r, y_r)$

The formulae from Hışıl's thesis are:

- $x_3 = \left(\frac{y_1 - y_2}{x_1 - x_2}\right)^2 - x_1 - x_2$
- $y_3 = \frac{y_1 - y_2}{x_1 - x_2} \cdot (x_1 - x_3) - y_1$

Rename:
- $(x_1, y_1)$ to $(x_q, y_q)$
- $(x_2, y_2)$ to $(x_p, y_p)$
- $(x_3, y_3)$ to $(x_r, y_r)$.

Let $\lambda = \frac{y_q - y_p}{x_q - x_p} = \frac{y_p - y_q}{x_p - x_q}$, which we implement as

$\lambda \cdot (x_p - x_q) = y_p - y_q$

Also,
- $x_r = \lambda^2 - x_q - x_p$
- $y_r = \lambda \cdot (x_q - x_r) - y_q$

which is equivalent to

- $x_r + x_q + x_p = \lambda^2$

Assuming $x_p \neq x_q$,

\begin{array}{|rrll|}
\hline
&(x_r + x_q + x_p) \cdot (x_p - x_q)^2 &=& \lambda^2 \cdot (x_p - x_q)^2\\
\implies &(x_r + x_q + x_p) \cdot (x_p - x_q)^2 &=& (\lambda \cdot (x_p - x_q))^2\\
\\\hline
         &y_r &=& \lambda \cdot (x_q - x_r) - y_q\\
\implies &y_r + y_q &=& \lambda \cdot (x_q - x_r)\\
\implies &(y_r + y_q) \cdot (x_p - x_q) &=& \lambda \cdot (x_p - x_q) \cdot (x_q - x_r)\\
\hline
\end{array}

Substituting for $\lambda \cdot (x_p - x_q)$, we get the constraints:
```

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
therealyingtong: The implementation has been updated to match this doc.

book/src/design/circuit/gadgets/ecc/addition.md
```diff
@@ -0,0 +1,372 @@
...
+                 \text{or } &(x_q - x_p) \cdot \lambda - (y_q - y_p) = 0 \\
+    \end{aligned} \\
+  && \text{must be satisfied for the constraint to be satisfied.} \\
+  && \\
+  && \text{If } x_q - x_p \neq 0, \text{ then } (x_q - x_p) \cdot \lambda - (y_q - y_p) = 0 \\
+  && \\
+  && \text{If } (x_q - x_p) \cdot \lambda - (y_q - y_p) = 0, \text{ then because } x_q - x_p \neq 0, \\
+  && \text{ by rearranging both sides we get } \lambda = (y_q - y_p) / (x_q - x_p) \\
+  && \\
+  && \text{and therefore } :: x_q \neq x_p \implies \lambda = (y_q - y_p) / (x_q - x_p).\\
```
daira, line 133:
```suggestion
  && \text{and therefore:} \\
  && x_q \neq x_p \implies \lambda = (y_q - y_p) / (x_q - x_p).\\
```

book/src/design/circuit/gadgets/ecc/addition.md
```diff
@@ -0,0 +1,372 @@
...
+  && \text{must be satisfied for the constraint to be satisfied.} \\
+  && \\
+  && \text{If } x_q = x_p, \text{ then } 1 - (x_q - x_p) \cdot \alpha = 0 \text{ has no solution for } \alpha, \\
+  && \text{ so it must be that } 2y_p \cdot \lambda - 3x_p^2 = 0. \\
+  && \\
+  && \text{If } y_p = 0 \text{ then } x_p = 0, \text{ and the constraint is satisfied.}\\
+  && \\
+  && \text{If } y_p \neq 0 \text{ then by rearranging both sides we get }\\
+  && \lambda = 3x_p^2 / 2y_p \\
+  && \text{Therefore } :: (x_q = x_p) \wedge y_p \neq 0 \implies \lambda = 3x_p^2 / 2y_p. \\
```
daira, line 149: formatting

book/src/design/circuit/gadgets/ecc/addition.md
```diff
@@ -0,0 +1,372 @@
...
+                 \text{or } x_r - x_q &= 0
+    \end{aligned} \\
+ && \\
+ && \text{must be satisfied for constraint (a) to be satisfied.} \\
+ && \text{If } x_p = 0 \text{ then } 1 - x_p \cdot \beta = 0 \text{ has no solutions for } \beta, \\
+ && \text{and so it must be that } x_r - x_q = 0. \\
+ && \\
+ && \text{Similarly, constraint (b) imposes that } y_r - y_q = 0. \\
+ && \\
+ && \text{Therefore } :: x_p = 0 => (x_r, y_r) = (x_q, y_q). \\
```
therealyingtong, line 196:
```suggestion
 && \text{Therefore:} \\
 && x_p = 0 \implies (x_r, y_r) = (x_q, y_q). \\
```

book/src/design/circuit/gadgets/ecc/addition.md
```diff
@@ -0,0 +1,372 @@
...
+  && \\
+  && \text{Let } x_p \neq 0 \wedge x_q \neq 0 \wedge y_q \neq -y_p. \\
+  && \begin{aligned}
+        &\text{• Similarly, constraint (c) imposes that } x_r = \lambda^2 - x_p - x_q \text{ is satisfied.} \\
+        &\text{• Similarly, constraint (d) imposes that } y_r = \lambda \cdot (x_p - x_r) - y_p \text{ is satisfied.} \\
+     \end{aligned} \\
+  && \\
+  && \text{Therefore} \\
+  && \begin{aligned}
+        :: &(x_p \neq 0) \wedge (x_q \neq 0) \wedge ((x_q \neq x_p) \vee (y_q \neq -y_p)) \\
```
daira, line 178: formatting

book/src/design/circuit/gadgets/ecc/addition.md
```diff
@@ -0,0 +1,372 @@
...
+    \end{aligned} \\
+ && \\
+ && \text{must be satisfied for constraint (a) to be satisfied.} \\
+ && \\
+ && \text{If } x_q = 0 \text{ then } 1 - x_q \cdot \beta = 0 \text{ has no solutions for } \beta, \\
+ && \text{and so it must be that } x_r - x_p = 0. \\
+ && \\
+ && \text{Similarly, constraint (b) imposes that } y_r - y_p = 0. \\
+ && \\
+ && \text{Therefore } :: x_q = 0 \implies (x_r, y_r) = (x_p, y_p). \\
```
daira, line 213: formatting

book/src/design/circuit/gadgets/ecc/addition.md
```diff
@@ -0,0 +1,372 @@
...
+    \text{At least one of } &1 - (x_q - x_p) \cdot \alpha - (y_q + y_p) \cdot \delta = 0 \\
+                 \text{or } &x_r = 0
+    \end{aligned} \\
+ && \\
+ && \text{must be satisfied for constraint (a) to be satisfied.}
+ && \\
+ && \text{If } x_r \neq 0, \text{ then it must be that } 1 - (x_q - x_p) \cdot \alpha - (y_q + y_p) \cdot \delta = 0. \\
+ && \\
+ && \text{However, if } x_q = x_p \wedge y_q = -y_p, \text{ then there are no solutions for } \alpha \text { and } \delta. \\
+ && \text{Therefore } : x_q = x_p \wedge y_q = -y_p \implies (x_r, y_r) = (0, 0).
```
daira, line 228:
```suggestion
 && \text{Therefore:} x_q = x_p \wedge y_q = -y_p \implies (x_r, y_r) = (0, 0).
```

book/src/design/circuit/gadgets/ecc/var-base-scalar-mul.md
```diff
@@ -0,0 +1,324 @@
...
+    &\texttt{// let } y_{A,i} = \frac{(\lambda_{1,i} + \lambda_{2,i}) \cdot (x_{A,i} - (\lambda_{1,i}^2 - x_{A,i} - x_T))}{2}\\
+    &(\mathbf{k}_i)(\mathbf{k}_i-1) = 0\\
+    &\lambda_{1,i} \cdot (x_{A,i} - x_T) = y_{A,i} - (2 \mathbf{k}_i - 1) \cdot y_T\\
+    &\lambda_{2,i}^2 = x_{A,i-1} + \lambda_{1,i}^2 - x_T\\
+\end{aligned}
+$$
+$$
+\begin{cases}
+    \lambda_{2,i} \cdot (x_{A,i} - x_{A,i-1}) = y_{A,i} + y_{A, i-1}, &\text{if } i > 4 \\[0.5ex]
+    \lambda_{2,4} \cdot (x_{A,4} - x_{A,3}) = y_{A,4} + y_{A,3}^\text{witnessed}, &\text{if } i = 4.
```
daira, line 138: Define `y_{A,3}^\text{witnessed}`.

therealyingtong:
(Eagerly merged; we can fix bugs later if we find them.)

PR #56, opened 2021-03-30
Use const generics for poseidon::ConstantLength

str4d:


daira (approved):
utACK. The length inference is really nice.

nuttycom (approved):

ebfull (approved):

PR #58, opened 2021-04-15
Proving and verifying keys, and placeholder proof logic

str4d:
The circuit is just a single gate to ensure that the prover has polynomials to run on.


daira (approved):
utACK

ebfull (approved):

PR #59, opened 2021-04-15
Implement ValueCommit^Orchard

str4d:


ebfull (approved):

src/value.rs
```diff
@@ -1,34 +1,241 @@
 //! Monetary values within the Orchard shielded pool.
 //!
 //! Values are represented in two places within Orchard:
-//! - The value of an individual note, which is unsigned.
-//! - The sum of note values within an Orchard [`Action`] or [`Bundle`], which is signed.
+//! - The value of an individual note, which is an unsigned 63-bit integer.
```
daira, line 4: 64-bit.

(If we implement UDAs, then the amount in a single token can be up to 2<sup>64</sup> - 1, even though the balance field is signed and in the range [-2<sup>63</sup>, 2<sup>63</sup>).)

str4d: I'll open a follow-up PR for this.

PR #63, opened 2021-04-19
Note commitment updates

str4d:
This brings in the type changes from the spec, that were made to address feedback in the NCC interim report. It also fixes a type error in `bundle::Action`.

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -33,7 +33,10 @@ impl SpendingKey {
     /// Returns `None` if the bytes do not correspond to a valid Orchard spending key.
     pub fn from_bytes(sk: [u8; 32]) -> CtOption<Self> {
         let sk = SpendingKey(sk);
-        // If ask = 0, discard this key.
+        // If ask = 0, discard this key. We call `derive_inner` rather than
+        // `SpendAuthorizingKey::from` here because we only need to know
+        // whether ask = 0; the adjustment to potentially negate ask is not
+        // needed. Also, `from` would panic on ask = 0.
         let ask = SpendAuthorizingKey::derive_inner(&sk);
         // If ivk = ⊥, discard this key.
         let ivk = IncomingViewingKey::derive_inner(&(&sk).into());
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -35,7 +35,9 @@ impl SpendingKey {
         let sk = SpendingKey(sk);
         // If ask = 0, discard this key.
         let ask = SpendAuthorizingKey::derive_inner(&sk);
-        CtOption::new(sk, !ask.ct_is_zero())
+        // If ivk = ⊥, discard this key.
+        let ivk = IncomingViewingKey::derive_inner(&(&sk).into());
+        CtOption::new(sk, !(ask.ct_is_zero() | ivk.is_none()))
     }
 }
 
@@ -263,12 +265,19 @@ pub struct IncomingViewingKey(pallas::Scalar);
 
 impl From<&FullViewingKey> for IncomingViewingKey {
     fn from(fvk: &FullViewingKey) -> Self {
-        let ak = extract_p(&pallas::Point::from_bytes(&(&fvk.ak.0).into()).unwrap());
-        IncomingViewingKey(commit_ivk(&ak, &fvk.nk.0, &fvk.rivk.0))
+        let ivk = IncomingViewingKey::derive_inner(fvk);
+        // IncomingViewingKey cannot be constructed such that this unwrap would fail.
+        IncomingViewingKey(ivk.unwrap())
     }
 }
 
 impl IncomingViewingKey {
+    /// Derives ask from sk. Internal use only, does not enforce all constraints.
+    fn derive_inner(fvk: &FullViewingKey) -> CtOption<pallas::Scalar> {
+        let ak = extract_p(&pallas::Point::from_bytes(&(&fvk.ak.0).into()).unwrap());
+        commit_ivk(&ak, &fvk.nk.0, &fvk.rivk.0)
+    }
+
     /// Returns the payment address for this key corresponding to the given diversifier.
     pub fn address(&self, d: Diversifier) -> Address {
         let pk_d = DiversifiedTransmissionKey::derive(self, &d);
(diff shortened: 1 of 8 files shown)


src/keys.rs
```diff
@@ -35,7 +35,9 @@ impl SpendingKey {
         let sk = SpendingKey(sk);
         // If ask = 0, discard this key.
         let ask = SpendAuthorizingKey::derive_inner(&sk);
```
daira, line 37:
```suggestion
        let ask = SpendAuthorizingKey::from(&sk);
```

daira, line 37: This is not doing the adjustment to make the ỹ bit 0. We can't do
```
        let ask = SpendAuthorizingKey::from(&sk);
```
because the assertion on line 63 might fail. The simplest way to fix this is to use `from` but remove the assertion on line 63. (The possible negation will not change ask if it is 0.)

src/keys.rs
```diff
@@ -35,7 +35,9 @@ impl SpendingKey {
         let sk = SpendingKey(sk);
         // If ask = 0, discard this key.
```
daira, line 36:
```suggestion
        // If ask = 0, discard this key. We call `derive_inner` rather than
        // `SpendAuthorizingKey::from` here because we only need to know
        // whether ask = 0; the adjustment to potentially negate ask is not needed.
```

str4d: This comment is not quite correct here: `SpendAuthorizingKey::from` panics if `ask = 0` because it relies on the invariant that a `SpendingKey` cannot be constructed such that `ask = 0`. So we _cannot_ call it here (and this is exactly why `dreive_inner` exists).

The additional comment about `ask` negation is maybe still useful, but if we did need it, it would be factored out of `SpendAuthorizingKey::from`.

src/keys.rs
```diff
@@ -263,12 +265,19 @@ pub struct IncomingViewingKey(pallas::Scalar);
...
+        let ivk = IncomingViewingKey::derive_inner(fvk);
+        // IncomingViewingKey cannot be constructed such that this unwrap would fail.
+        IncomingViewingKey(ivk.unwrap())
     }
 }
 
 impl IncomingViewingKey {
+    /// Derives ask from sk. Internal use only, does not enforce all constraints.
+    fn derive_inner(fvk: &FullViewingKey) -> CtOption<pallas::Scalar> {
+        let ak = extract_p(&pallas::Point::from_bytes(&(&fvk.ak.0).into()).unwrap());
```
nuttycom, line 277: I take it that the full viewing key is constrained by construction such that it cannot produce an invalid ivk?

src/spec.rs
```diff
@@ -120,11 +120,20 @@ pub(crate) fn ka_orchard(sk: &pallas::Scalar, b: &pallas::Point) -> pallas::Poin
...
+        .map(|c| *c.x())
+        .unwrap_or_else(pallas::Base::zero)
+}
+
+/// Coordinate extractor for Pallas.
+///
+/// Defined in [Zcash Protocol Spec § 5.4.9.7: Coordinate Extractor for Pallas][concreteextractorpallas].
+///
+/// [concreteextractorpallas]: https://zips.z.cash/protocol/nu5.pdf#concreteextractorpallas
+pub(crate) fn extract_p_bottom(point: CtOption<pallas::Point>) -> CtOption<pallas::Base> {
```
nuttycom, line 135: Should `IncompletePoint` be moved out of sinsemilla::addition and `extract_p_bottom` defined over that newtype?

nuttycom (approved):
One non-blocking question about `extract_p_bottom` (with @therealyingtong)

src/keys.rs
str4d: Correct. In the current code this only happens via `SpendingKey::from_bytes`, but we would also check this in parsing constructors.

src/spec.rs
str4d: I decided against that because incomplete addition is an implementation detail of Sinsemilla; what matters externally is the return type. The function signature `P u {bottom} -> P::Base u {bottom}` maps more intuitively to `Option<P> -> Option<P::Base>` with `bottom` represented as `None`.

nuttycom: Ordinarily, it's a design smell to take an `Option` as input unless you're doing to do something other than just map over it, because the caller is constrained by the types to do so anyway. It's not strictly wrong, of course, but perhaps just superfluous? If this is a pure function, then there's no possible implementation other than the one given, and the caller could do that just as easily?

src/keys.rs
daira: I adjusted the suggestion.

src/spec.rs
daira: This directly corresponds to a function in the spec, and I think it's better to make that correspondence clear at call sites.

daira:
We should decide whether or not we intend to follow the convention for `to_*` methods taking `&self`, and either `#[allow(clippy::wrong_self_convention)]` globally or fix those lints. (I would be fine with not following it, but I highly value being lint-clean either way.)

src/spec.rs
str4d: @nuttycom you're correct that the caller could do this map themselves. The existence of `ectract_p_bottom` as a function is purely for the benefit of people reviewing against the spec.

str4d:
@daira that lint was fixed in https://github.com/zcash/orchard/pull/59 (which isn't in this PR's history).

src/primitives/sinsemilla/addition.rs
```diff
@@ -0,0 +1,55 @@
...
+                    // P ⊹ 0 = ⊥
+                    q.to_affine().coordinates().and_then(|c_q| {
+                        // (x, y) ⊹ (x', y') = ⊥ if x == x'
+                        // (x, y) ⊹ (x', y') = (x, y) + (x', y') if x != x'
+                        CtOption::new(p + q, !c_p.x().ct_eq(c_q.x()))
+                    })
+                })
+            })
+        }))
+    }
```
ebfull, line 46: Had a discussion with @str4d, we can do these operations in Jacobian still and just condition it on `P != Q` and `P != (-Q)`.

ebfull (changes requested):

ebfull (approved):
ACK modulo clippy fixes

daira (commented):
Post-hoc utACK

PR #67, opened 2021-04-21
Sinsemilla chip with HashDomain

therealyingtong:
Part of #96. This PR only covers `SinsemillaHash`, not `SinsemillaCommit`.

The `SinsemillaCommit` functionality is in https://github.com/zcash/orchard/pull/118.

### DIFF
--- a/src/circuit/gadget/sinsemilla.rs
+++ b/src/circuit/gadget/sinsemilla.rs
@@ -6,7 +6,7 @@ use crate::circuit::gadget::{
 use halo2::{arithmetic::CurveAffine, circuit::Layouter, plonk::Error};
 use std::fmt::Debug;
 
-// pub mod chip;
+pub mod chip;
 mod message;
 
 // pub use chip::{SinsemillaChip, SinsemillaConfig};
--- a/src/circuit/gadget/sinsemilla.rs
+++ b/src/circuit/gadget/sinsemilla.rs
@@ -408,4 +408,18 @@ mod tests {
         let prover = MockProver::run(k, &circuit, vec![]).unwrap();
         assert_eq!(prover.verify(), Ok(()))
     }
+
+    #[cfg(feature = "dev-graph")]
+    #[test]
+    fn print_sinsemilla_chip() {
+        use plotters::prelude::*;
+
+        let root =
+            BitMapBackend::new("sinsemilla-hash-layout.png", (1024, 7680)).into_drawing_area();
+        root.fill(&WHITE).unwrap();
+        let root = root.titled("SinsemillaHash", ("sans-serif", 60)).unwrap();
+
+        let circuit = MyCircuit {};
+        halo2::dev::circuit_layout(&circuit, &root).unwrap();
+    }
 }
--- a/src/circuit/gadget/sinsemilla.rs
+++ b/src/circuit/gadget/sinsemilla.rs
@@ -9,8 +9,6 @@ use std::fmt::Debug;
 pub mod chip;
 mod message;
 
-// pub use chip::{SinsemillaChip, SinsemillaConfig};
-
 /// The set of circuit instructions required to use the [`Sinsemilla`](https://zcash.github.io/halo2/design/gadgets/sinsemilla.html) gadget.
 /// This trait is bounded on two constant parameters: `K`, the number of bits
 /// in each word accepted by the Sinsemilla hash, and `MAX_WORDS`, the maximum
@@ -230,4 +228,184 @@ where
 #[allow(non_snake_case)]
 pub trait HashDomains<C: CurveAffine>: Clone + Debug {
     fn Q(&self) -> C;
+}
+
+#[cfg(test)]
+mod tests {
+    use halo2::{
+        circuit::{layouter::SingleChipLayouter, Layouter},
+        dev::MockProver,
+        pasta::pallas,
+        plonk::{Assignment, Circuit, ConstraintSystem, Error},
+    };
+
+    use super::{
+        chip::SinsemillaHashDomains,
+        chip::{SinsemillaChip, SinsemillaConfig},
+        HashDomain, Message, SinsemillaInstructions,
+    };
+
+    use crate::{
+        circuit::gadget::ecc::{
+            chip::{EccChip, EccConfig},
+            Point,
+        },
+        constants::MERKLE_CRH_PERSONALIZATION,
+        primitives::sinsemilla::{self, K},
+    };
+
+    use group::Curve;
+
+    use std::convert::TryInto;
+
+    struct MyCircuit {}
+
+    impl Circuit<pallas::Base> for MyCircuit {
+        type Config = (EccConfig, SinsemillaConfig, SinsemillaConfig);
+
+        #[allow(non_snake_case)]
+        fn configure(meta: &mut ConstraintSystem<pallas::Base>) -> Self::Config {
+            let advices = [
+                meta.advice_column(),
+                meta.advice_column(),
+                meta.advice_column(),
+                meta.advice_column(),
+                meta.advice_column(),
+                meta.advice_column(),
+                meta.advice_column(),
+                meta.advice_column(),
+                meta.advice_column(),
+                meta.advice_column(),
+            ];
+
+            let constants = meta.fixed_column();
+            let perm = meta.permutation(
+                &advices
+                    .iter()
+                    .map(|advice| (*advice).into())
+                    .chain(Some(constants.into()))
+                    .collect::<Vec<_>>(),
+            );
+
+            let ecc_config = EccChip::configure(meta, advices, perm.clone());
+
+            // Fixed columns for the Sinsemilla generator lookup table
+            let lookup = (
+                meta.fixed_column(),
+                meta.fixed_column(),
+                meta.fixed_column(),
+            );
+
+            let config1 = SinsemillaChip::configure(
+                meta,
+                advices[..5].try_into().unwrap(),
+                lookup,
+                constants,
+                perm.clone(),
+            );
+            let config2 = SinsemillaChip::configure(
+                meta,
+                advices[5..].try_into().unwrap(),
+                lookup,
+                constants,
+                perm,
+            );
+            (ecc_config, config1, config2)
+        }
+
+        fn synthesize(
+            &self,
+            cs: &mut impl Assignment<pallas::Base>,
+            config: Self::Config,
+        ) -> Result<(), Error> {
+            let mut layouter = SingleChipLayouter::new(cs)?;
+            let ecc_chip = EccChip::construct(config.0);
+
+            // The two `SinsemillaChip`s share the same lookup table.
+            SinsemillaChip::load(config.1.clone(), &mut layouter)?;
+
+            // This MerkleCRH example is purely for illustrative purposes.
+            // It is not an implementation of the Orchard protocol spec.
+            {
+                let chip1 = SinsemillaChip::construct(config.1);
+
+                let merkle_crh = HashDomain::new(
+                    chip1.clone(),
+                    ecc_chip.clone(),
+                    &SinsemillaHashDomains::MerkleCrh,
+                );
+
+                // Layer 31, l = MERKLE_DEPTH_ORCHARD - 1 - layer = 0
+                let l_bitstring = vec![Some(false); K];
+                let l = chip1
+                    .witness_message_piece_bitstring(layouter.namespace(|| "l"), &l_bitstring)?;
+
+                // Left leaf
+                let left_bitstring: Vec<Option<bool>> =
+                    (0..250).map(|_| Some(rand::random::<bool>())).collect();
+                let left = chip1.witness_message_piece_bitstring(
+                    layouter.namespace(|| "left"),
+                    &left_bitstring,
+                )?;
+
+                // Right leaf
+                let right_bitstring: Vec<Option<bool>> =
+                    (0..250).map(|_| Some(rand::random::<bool>())).collect();
+                let right = chip1.witness_message_piece_bitstring(
+                    layouter.namespace(|| "right"),
+                    &right_bitstring,
+                )?;
+
+                let l_bitstring: Option<Vec<bool>> = l_bitstring.into_iter().collect();
+                let left_bitstring: Option<Vec<bool>> = left_bitstring.into_iter().collect();
+                let right_bitstring: Option<Vec<bool>> = right_bitstring.into_iter().collect();
+
+                // Witness expected parent
+                let expected_parent = {
+                    let expected_parent = if let (Some(l), Some(left), Some(right)) =
+                        (l_bitstring, left_bitstring, right_bitstring)
+                    {
+                        let merkle_crh = sinsemilla::HashDomain::new(MERKLE_CRH_PERSONALIZATION);
+                        let point = merkle_crh
+                            .hash_to_point(
+                                l.into_iter()
+                                    .chain(left.into_iter())
+                                    .chain(right.into_iter()),
+                            )
+                            .unwrap();
+                        Some(point.to_affine())
+                    } else {
+                        None
+                    };
+
+                    Point::new(
+                        ecc_chip,
+                        layouter.namespace(|| "Witness expected parent"),
+                        expected_parent,
+                    )?
+                };
+
+                // Parent
+                let parent = {
+                    let message = Message::from_pieces(chip1, vec![l, left, right]);
+                    merkle_crh.hash_to_point(layouter.namespace(|| "parent"), message)?
+                };
+
+                parent.constrain_equal(
+                    layouter.namespace(|| "parent == expected parent"),
+                    &expected_parent,
+                )?;
+            }
+
+            Ok(())
+        }
+    }
+
+    #[test]
+    fn sinsemilla_chip() {
+        let k = 11;
+        let circuit = MyCircuit {};
+        let prover = MockProver::run(k, &circuit, vec![]).unwrap();
+        assert_eq!(prover.verify(), Ok(()))
+    }
 }
--- a/src/circuit/gadget/sinsemilla.rs
+++ b/src/circuit/gadget/sinsemilla.rs
@@ -1,0 +1,233 @@
+//! Gadget and chips for the Sinsemilla hash function.
+use crate::circuit::gadget::{
+    ecc::{self, EccInstructions},
+    utilities::Var,
+};
+use halo2::{arithmetic::CurveAffine, circuit::Layouter, plonk::Error};
+use std::fmt::Debug;
+
+// pub mod chip;
+// mod message;
+
+// pub use chip::{SinsemillaChip, SinsemillaConfig};
+
+/// The set of circuit instructions required to use the [`Sinsemilla`](https://zcash.github.io/halo2/design/gadgets/sinsemilla.html) gadget.
+/// This trait is bounded on two constant parameters: `K`, the number of bits
+/// in each word accepted by the Sinsemilla hash, and `MAX_WORDS`, the maximum
+/// number of words that a single hash instance can process.
+pub trait SinsemillaInstructions<C: CurveAffine, const K: usize, const MAX_WORDS: usize> {
+    /// A variable in the circuit.
+    type CellValue: Var<C::Base>;
+
+    /// A message composed of [`Self::MessagePiece`]s.
+    type Message: From<Vec<Self::MessagePiece>>;
+
+    /// A piece in a message containing a number of `K`-bit words.
+    /// A [`Self::MessagePiece`] fits in a single base field element,
+    /// which means it can only contain up to `N` words, where
+    /// `N*K <= C::Base::NUM_BITS`.
+    ///
+    /// For example, in the case `K = 10`, `NUM_BITS = 255`, we can fit
+    /// up to `N = 25` words in a single base field element.
+    type MessagePiece;
+
+    /// The x-coordinate of a point output of [`Self::hash_to_point`].
+    type X;
+    /// A point output of [`Self::hash_to_point`].
+    type Point: Clone + Debug;
+
+    /// HashDomains used in this instruction.
+    type HashDomains: HashDomains<C>;
+
+    /// Witness a message in the given bitstring.
+    /// Returns a vector of [`Self::MessagePiece`]s encoding the given message.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the message length is not a multiple of `K`.
+    ///
+    /// Panics if the message length exceeds `K * MAX_WORDS`.
+    fn witness_message(
+        &self,
+        layouter: impl Layouter<C::Base>,
+        message: Vec<Option<bool>>,
+    ) -> Result<Self::Message, Error>;
+
+    /// Witnesses a message piece given a field element and the intended number of `K`-bit
+    /// words it contains.
+    ///
+    /// Returns a [`Self::MessagePiece`] encoding the given message.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the message length is not a multiple of `K`.
+    ///
+    /// Panics if the message length exceeds the maximum number of words
+    /// that can fit in a field element.
+    fn witness_message_piece_bitstring(
+        &self,
+        layouter: impl Layouter<C::Base>,
+        message: &[Option<bool>],
+    ) -> Result<Self::MessagePiece, Error>;
+
+    /// Witness a message piece given a field element. Returns a [`Self::MessagePiece`]
+    /// encoding the given message.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `num_words` exceed the maximum number of `K`-bit words that
+    /// can fit into a single base field element.
+    fn witness_message_piece_field(
+        &self,
+        layouter: impl Layouter<C::Base>,
+        value: Option<C::Base>,
+        num_words: usize,
+    ) -> Result<Self::MessagePiece, Error>;
+
+    /// Hashes a message to an ECC curve point.
+    /// This returns both the resulting point, as well as the message
+    /// decomposition in the form of intermediate values in a cumulative
+    /// sum.
+    ///
+    /// A cumulative sum `z` is used to decompose a Sinsemilla message. It
+    /// produces intermediate values for each word in the message, such
+    /// that `z_next` = (`z_cur` - `word_next`) / `2^K`.
+    ///  
+    /// These intermediate values are useful for range checks on subsets
+    /// of the Sinsemilla message. Sinsemilla messages in the Orchard
+    /// protocol are composed of field elements, and we need to check
+    /// the canonicity of the field element encodings in certain cases.
+    ///
+    #[allow(non_snake_case)]
+    #[allow(clippy::type_complexity)]
+    fn hash_to_point(
+        &self,
+        layouter: impl Layouter<C::Base>,
+        Q: C,
+        message: Self::Message,
+    ) -> Result<(Self::Point, Vec<Vec<Self::CellValue>>), Error>;
+
+    /// Extracts the x-coordinate of the output of a Sinsemilla hash.
+    fn extract(point: &Self::Point) -> Self::X;
+}
+
+/// A message to be hashed.
+///
+/// Composed of [`MessagePiece`]s with bitlength some multiple of `K`.
+///
+/// [`MessagePiece`]: SinsemillaInstructions::MessagePiece
+#[derive(Clone, Debug)]
+pub struct Message<C: CurveAffine, SinsemillaChip, const K: usize, const MAX_WORDS: usize>
+where
+    SinsemillaChip: SinsemillaInstructions<C, K, MAX_WORDS> + Clone + Debug + Eq,
+{
+    chip: SinsemillaChip,
+    inner: SinsemillaChip::Message,
+}
+
+impl<C: CurveAffine, SinsemillaChip, const K: usize, const MAX_WORDS: usize>
+    Message<C, SinsemillaChip, K, MAX_WORDS>
+where
+    SinsemillaChip: SinsemillaInstructions<C, K, MAX_WORDS> + Clone + Debug + Eq,
+{
+    fn from_bitstring(
+        chip: SinsemillaChip,
+        layouter: impl Layouter<C::Base>,
+        bitstring: Vec<Option<bool>>,
+    ) -> Result<Self, Error> {
+        let inner = chip.witness_message(layouter, bitstring)?;
+        Ok(Self { chip, inner })
+    }
+
+    /// Constructs a message from a vector of [`MessagePiece`]s.
+    ///
+    /// [`MessagePiece`]: SinsemillaInstructions::MessagePiece
+    fn from_pieces(chip: SinsemillaChip, pieces: Vec<SinsemillaChip::MessagePiece>) -> Self {
+        Self {
+            chip,
+            inner: pieces.into(),
+        }
+    }
+}
+
+/// A domain in which $\mathsf{SinsemillaHashToPoint}$ and $\mathsf{SinsemillaHash}$ can
+/// be used.
+#[allow(non_snake_case)]
+pub struct HashDomain<
+    C: CurveAffine,
+    SinsemillaChip,
+    EccChip,
+    const K: usize,
+    const MAX_WORDS: usize,
+> where
+    SinsemillaChip: SinsemillaInstructions<C, K, MAX_WORDS> + Clone + Debug + Eq,
+    EccChip: EccInstructions<
+            C,
+            Point = <SinsemillaChip as SinsemillaInstructions<C, K, MAX_WORDS>>::Point,
+        > + Clone
+        + Debug
+        + Eq,
+{
+    sinsemilla_chip: SinsemillaChip,
+    ecc_chip: EccChip,
+    Q: C,
+}
+
+impl<C: CurveAffine, SinsemillaChip, EccChip, const K: usize, const MAX_WORDS: usize>
+    HashDomain<C, SinsemillaChip, EccChip, K, MAX_WORDS>
+where
+    SinsemillaChip: SinsemillaInstructions<C, K, MAX_WORDS> + Clone + Debug + Eq,
+    EccChip: EccInstructions<
+            C,
+            Point = <SinsemillaChip as SinsemillaInstructions<C, K, MAX_WORDS>>::Point,
+        > + Clone
+        + Debug
+        + Eq,
+{
+    #[allow(non_snake_case)]
+    /// Constructs a new `HashDomain` for the given domain.
+    pub fn new(
+        sinsemilla_chip: SinsemillaChip,
+        ecc_chip: EccChip,
+        domain: &SinsemillaChip::HashDomains,
+    ) -> Self {
+        HashDomain {
+            sinsemilla_chip,
+            ecc_chip,
+            Q: domain.Q(),
+        }
+    }
+
+    /// $\mathsf{SinsemillaHashToPoint}$ from [§ 5.4.1.9][concretesinsemillahash].
+    ///
+    /// [concretesinsemillahash]: https://zips.z.cash/protocol/protocol.pdf#concretesinsemillahash
+    pub fn hash_to_point(
+        &self,
+        layouter: impl Layouter<C::Base>,
+        message: Message<C, SinsemillaChip, K, MAX_WORDS>,
+    ) -> Result<ecc::Point<C, EccChip>, Error> {
+        assert_eq!(self.sinsemilla_chip, message.chip);
+        self.sinsemilla_chip
+            .hash_to_point(layouter, self.Q, message.inner)
+            .map(|(point, _)| ecc::Point::from_inner(self.ecc_chip.clone(), point))
+    }
+
+    /// $\mathsf{SinsemillaHash}$ from [§ 5.4.1.9][concretesinsemillahash].
+    ///
+    /// [concretesinsemillahash]: https://zips.z.cash/protocol/protocol.pdf#concretesinsemillahash
+    pub fn hash(
+        &self,
+        layouter: impl Layouter<C::Base>,
+        message: Message<C, SinsemillaChip, K, MAX_WORDS>,
+    ) -> Result<ecc::X<C, EccChip>, Error> {
+        assert_eq!(self.sinsemilla_chip, message.chip);
+        let p = self.hash_to_point(layouter, message);
+        p.map(|p| p.extract_p())
+    }
+}
+
+/// Trait allowing circuit's Sinsemilla HashDomains to be enumerated.
+#[allow(non_snake_case)]
+pub trait HashDomains<C: CurveAffine>: Clone + Debug {
+    fn Q(&self) -> C;
+}
--- a/src/circuit/gadget/sinsemilla.rs
+++ b/src/circuit/gadget/sinsemilla.rs
@@ -7,7 +7,7 @@ use halo2::{arithmetic::CurveAffine, circuit::Layouter, plonk::Error};
 use std::fmt::Debug;
 
 // pub mod chip;
-// mod message;
+mod message;
 
 // pub use chip::{SinsemillaChip, SinsemillaConfig};
 
--- a/src/circuit/gadget/sinsemilla/chip.rs
+++ b/src/circuit/gadget/sinsemilla/chip.rs
@@ -16,7 +16,10 @@ use ff::PrimeField;
 use halo2::{
     arithmetic::{CurveAffine, FieldExt},
     circuit::{Chip, Layouter},
-    plonk::{Advice, Column, ConstraintSystem, Error, Expression, Fixed, Permutation, Selector},
+    plonk::{
+        Advice, Column, ConstraintSystem, Error, Expression, Fixed, Permutation, Selector,
+        VirtualCells,
+    },
     poly::Rotation,
 };
 use pasta_curves::pallas;
@@ -126,25 +129,32 @@ impl SinsemillaChip {
         // Set up lookup argument
         GeneratorTableConfig::configure(meta, config.clone());
 
+        // Constant expressions
         let two = Expression::Constant(pallas::Base::from_u64(2));
+
+        // Closures for expressions that are derived multiple times
+        // x_r = lambda_1^2 - x_a - x_p
+        let x_r = |meta: &mut VirtualCells<pallas::Base>, rotation| {
+            let x_a = meta.query_advice(config.x_a, rotation);
+            let x_p = meta.query_advice(config.x_p, rotation);
+            let lambda_1 = meta.query_advice(config.lambda_1, rotation);
+            lambda_1.square() - x_a - x_p
+        };
+
+        // Y_A = (lambda_1 + lambda_2) * (x_a - x_r)
+        let Y_A = |meta: &mut VirtualCells<pallas::Base>, rotation| {
+            let x_a = meta.query_advice(config.x_a, rotation);
+            let lambda_1 = meta.query_advice(config.lambda_1, rotation);
+            let lambda_2 = meta.query_advice(config.lambda_2, rotation);
+            (lambda_1 + lambda_2) * (x_a - x_r(meta, rotation))
+        };
+
         // Check that the initial x_A, x_P, lambda_1, lambda_2 are consistent with y_Q.
         meta.create_gate("Initial y_Q", |meta| {
             let fixed_y_q = meta.query_fixed(config.fixed_y_q, Rotation::cur());
 
             // Y_A = (lambda_1 + lambda_2) * (x_a - x_r)
-            let Y_A = {
-                let lambda_1 = meta.query_advice(config.lambda_1, Rotation::cur());
-                let lambda_2 = meta.query_advice(config.lambda_2, Rotation::cur());
-                let x_a = meta.query_advice(config.x_a, Rotation::cur());
-
-                // x_r = lambda_1^2 - x_a - x_p
-                let x_r = {
-                    let x_p = meta.query_advice(config.x_p, Rotation::cur());
-                    lambda_1.clone().square() - x_a.clone() - x_p
-                };
-
-                (lambda_1 + lambda_2) * (x_a - x_r)
-            };
+            let Y_A = Y_A(meta, Rotation::cur());
 
             // fixed_y_q * (2 * fixed_y_q - Y_{A,0}) = 0
             vec![fixed_y_q.clone() * (two.clone() * fixed_y_q - Y_A)]
@@ -152,16 +162,12 @@ impl SinsemillaChip {
 
         meta.create_gate("Secant line", |meta| {
             let q_s1 = meta.query_selector(config.q_sinsemilla1);
-            let lambda_1 = meta.query_advice(config.lambda_1, Rotation::cur());
             let lambda_2 = meta.query_advice(config.lambda_2, Rotation::cur());
             let x_a_cur = meta.query_advice(config.x_a, Rotation::cur());
             let x_a_next = meta.query_advice(config.x_a, Rotation::next());
 
             // x_r = lambda_1^2 - x_a_cur - x_p
-            let x_r = {
-                let x_p = meta.query_advice(config.x_p, Rotation::cur());
-                lambda_1.square() - x_a_cur.clone() - x_p
-            };
+            let x_r = x_r(meta, Rotation::cur());
 
             // lambda2^2 - (x_a_next + x_r + x_a_cur) = 0
             let secant_line = lambda_2.square() - (x_a_next + x_r + x_a_cur);
@@ -178,32 +184,14 @@ impl SinsemillaChip {
             };
             let x_a_cur = meta.query_advice(config.x_a, Rotation::cur());
             let x_a_next = meta.query_advice(config.x_a, Rotation::next());
-            let lambda_1_cur = meta.query_advice(config.lambda_1, Rotation::cur());
             let lambda_2_cur = meta.query_advice(config.lambda_2, Rotation::cur());
             let lambda_1_next = meta.query_advice(config.lambda_1, Rotation::next());
-            let lambda_2_next = meta.query_advice(config.lambda_2, Rotation::next());
 
             // Y_A = (lambda_1 + lambda_2) * (x_a - x_r)
-            let Y_A_cur = {
-                // x_r = lambda_1^2 - x_a - x_p
-                let x_r = {
-                    let x_p = meta.query_advice(config.x_p, Rotation::cur());
-                    lambda_1_cur.clone().square() - x_a_cur.clone() - x_p
-                };
-
-                (lambda_1_cur + lambda_2_cur.clone()) * (x_a_cur.clone() - x_r)
-            };
+            let Y_A_cur = Y_A(meta, Rotation::cur());
 
             // Y_A = (lambda_1 + lambda_2) * (x_a - x_r)
-            let Y_A_next = {
-                // x_r = lambda_1^2 - x_a - x_p
-                let x_r = {
-                    let x_p = meta.query_advice(config.x_p, Rotation::next());
-                    lambda_1_next.clone().square() - x_a_next.clone() - x_p
-                };
-
-                (lambda_1_next.clone() + lambda_2_next) * (x_a_next.clone() - x_r)
-            };
+            let Y_A_next = Y_A(meta, Rotation::next());
 
             // lhs = 4 * lambda_2_cur * (x_a_cur - x_a_next)
             let lhs = lambda_2_cur * pallas::Base::from_u64(4) * (x_a_cur - x_a_next);
--- a/src/circuit/gadget/sinsemilla/chip.rs
+++ b/src/circuit/gadget/sinsemilla/chip.rs
@@ -160,21 +160,6 @@ impl SinsemillaChip {
             vec![fixed_y_q.clone() * (two.clone() * fixed_y_q - Y_A)]
         });
 
-        meta.create_gate("Secant line", |meta| {
-            let q_s1 = meta.query_selector(config.q_sinsemilla1);
-            let lambda_2 = meta.query_advice(config.lambda_2, Rotation::cur());
-            let x_a_cur = meta.query_advice(config.x_a, Rotation::cur());
-            let x_a_next = meta.query_advice(config.x_a, Rotation::next());
-
-            // x_r = lambda_1^2 - x_a_cur - x_p
-            let x_r = x_r(meta, Rotation::cur());
-
-            // lambda2^2 - (x_a_next + x_r + x_a_cur) = 0
-            let secant_line = lambda_2.square() - (x_a_next + x_r + x_a_cur);
-
-            vec![q_s1 * secant_line]
-        });
-
         meta.create_gate("Sinsemilla gate", |meta| {
             let q_s1 = meta.query_selector(config.q_sinsemilla1);
             let q_s2 = meta.query_fixed(config.q_sinsemilla2, Rotation::cur());
@@ -182,10 +167,14 @@ impl SinsemillaChip {
                 let one = Expression::Constant(pallas::Base::one());
                 q_s2.clone() * (q_s2 - one)
             };
+
+            let lambda_1_next = meta.query_advice(config.lambda_1, Rotation::next());
+            let lambda_2_cur = meta.query_advice(config.lambda_2, Rotation::cur());
             let x_a_cur = meta.query_advice(config.x_a, Rotation::cur());
             let x_a_next = meta.query_advice(config.x_a, Rotation::next());
-            let lambda_2_cur = meta.query_advice(config.lambda_2, Rotation::cur());
-            let lambda_1_next = meta.query_advice(config.lambda_1, Rotation::next());
+
+            // x_r = lambda_1^2 - x_a_cur - x_p
+            let x_r = x_r(meta, Rotation::cur());
 
             // Y_A = (lambda_1 + lambda_2) * (x_a - x_r)
             let Y_A_cur = Y_A(meta, Rotation::cur());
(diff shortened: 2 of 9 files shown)


src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,281 @@
...
+                    || piece.field_elem().ok_or(Error::SynthesisError),
+                )?;
+                region.constrain_equal(&config.perm, piece.cell(), cell)?;
+                zs.push(CellValue::new(cell, piece.field_elem()));
+
+                // Assign cumulative sum such that
+                //          z_i = 2^10 * z_{i + 1} + m_{i + 1}
+                // => z_{i + 1} = (z_i - m_{i + 1}) / 2^10
+                //
+                // For a message m = m_1 + 2^K m_2 + ... + 2^{Kl} m_l}, initialize z_0 = m.
```
therealyingtong, line 175:
```suggestion
                // For a message m = m_1 + 2^K m_2 + ... + 2^{K(l-1)} m_l}, initialize z_0 = m.
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,333 @@
...
+            // Assign cumulative sum such that
+            //          z_i = 2^10 * z_{i + 1} + m_{i + 1}
+            // => z_{i + 1} = (z_i - m_{i + 1}) / 2^10
+            //
+            // For a message m = m_1 + 2^K m_2 + ... + 2^{Kl} m_l}, initialize z_0 = m.
+            // We end up with z_{l} = 0.
+            let mut z = piece.field_elem();
+            let inv_2_k = C::Base::from_u64(1 << K).invert().unwrap(); // TODO: Constant
+            for (idx, word) in words.iter().enumerate() {
+                // z_next = (z_cur - m_next) / 2^K
```
therealyingtong, line 194: The indices are slightly confusing here because Sinsemilla words are one-indexed (as per the [protocol spec](https://zips.z.cash/protocol/nu5.pdf#concretesinsemillahash)). So `m_{i + 1}` actually refers to the word on the same row as `z_i`.

image

```suggestion
                // z_{i + 1} = (z_i - m_{i + 1}) / 2^10
```

str4d:
[Rebased](https://github.com/zcash/orchard/compare/8eada408d0d853795c5c33f6dda04ab3c3b5f018..178689e37948ead96d6b2f3f71d9bb2f92a02a01) on #107 after it was rebased. Changes due to #109 were squashed into the relevant commits.

str4d:
This PR now only implements SinsemillaHash. SinsemillaCommit (which requires #111) has been moved to a separate branch (https://github.com/zcash/orchard/tree/sinsemilla-chip-commit), which is based on this PR but will be broken until #111 is complete.

src/circuit/gadget/sinsemilla.rs
```diff
@@ -0,0 +1,386 @@
...
+        num_words: usize,
+    ) -> Result<Self::MessagePiece, Error>;
+
+    /// Hashes a message to an ECC curve point.
+    /// This returns both the resulting point, as well as the message
+    /// decomposition in the form of intermediate values in a cumulative
+    /// sum.
+    ///
+    /// A cumulative sum `z` is used to decompose a Sinsemilla message. It
+    /// produced intermediate values for each word in the message, such
```
daira, line 92:
```suggestion
    /// produces intermediate values for each word in the message, such
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,301 @@
...
+            });
+
+        // Closure to decompose a message piece into `K`-bit pieces with a running sum `z`.
+        let decompose_message =
+            |region: &mut Region<'_, pallas::Base>, words: Vec<Option<u32>>| -> Result<(), Error> {
+                let mut zs = Vec::with_capacity(piece.num_words() + 1);
+
+                // Copy message and initialize running sum `z` to decompose message in-circuit
+                let cell = region.assign_advice(
+                    || "copy message",
```
str4d:
```suggestion
                    || "z_0 (copy of message)",
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,301 @@
...
+                offset + row,
+                || x_p.ok_or(Error::SynthesisError),
+            )?;
+
+            // Compute and assign `lambda1, lambda2`
+            let lambda1 = x_a
+                .zip(y_a)
+                .zip(x_p)
+                .zip(y_p)
+                .map(|(((x_a, y_a), x_p), y_p)| (y_a - y_p) * (x_a - x_p).invert().unwrap());
```
str4d: These inversions would be batchable if https://github.com/zcash/halo2/pull/293 merges.

src/primitives/sinsemilla.rs
```diff
@@ -14,7 +14,7 @@ mod constants;
 mod sinsemilla_s;
 pub use constants::*;
 
-fn lebs2ip_k(bits: &[bool]) -> u32 {
+pub fn lebs2ip_k(bits: &[bool]) -> u32 {
```
str4d, line 17:
```suggestion
pub(crate) fn lebs2ip_k(bits: &[bool]) -> u32 {
```

src/circuit/gadget/sinsemilla.rs
```diff
@@ -210,3 +210,164 @@ where
...
+                let l = {
+                    let merkle_depth_orchard = 32;
+                    let layer = 0;
+                    let l = pallas::Base::from_u64(merkle_depth_orchard - 1 - layer);
+                    chip1.witness_message_piece_field(layouter.namespace(|| "l"), Some(l), 1)?
+                };
+
+                // Parent
+                let parent = Message::from_pieces(chip1, vec![l, left, right]);
+                merkle_crh.hash_to_point(layouter.namespace(|| "parent"), parent)?;
```
str4d, line 371: If possible, witness the expected result and constrain it to equal the output of this.

str4d (changes requested):

src/circuit/gadget/sinsemilla.rs
```diff
@@ -0,0 +1,386 @@
...
+    /// The x-coordinate of a point output of [`Self::hash_to_point`].
+    type X;
+    /// A point output of [`Self::hash_to_point`].
+    type Point: Clone + Debug;
+
+    /// HashDomains used in this instruction.
+    type HashDomains: HashDomains<C>;
+
+    /// Witness a message given a bitstring and the intended number of `K`-bit
+    /// words represented by the bitstring.
```
daira, line 43: This doesn't explicitly take the length as input.

src/circuit/gadget/sinsemilla.rs
```diff
@@ -0,0 +1,386 @@
...
+        HashDomain {
+            sinsemilla_chip,
+            ecc_chip,
+            Q: domain.Q(),
+        }
+    }
+
+    /// $\mathsf{SinsemillaHashToPoint}$ from [§ 5.4.1.9][concretesinsemillahash].
+    ///
+    /// [concretesinsemillahash]: https://zips.z.cash/protocol/nu5.pdf#concretesinsemillahash
```
daira, line 195: These references can be to protocol.pdf now.

src/circuit/gadget/sinsemilla.rs
```diff
@@ -0,0 +1,386 @@
...
+            SinsemillaChip::<K, MAX_WORDS>::load(config.1.clone(), &mut layouter)?;
+
+            // This MerkleCRH example is purely for illustrative purposes.
+            // It is not an implementation of the Orchard protocol spec.
+            {
+                let chip1 = SinsemillaChip::<K, MAX_WORDS>::construct(config.1);
+
+                let merkle_crh = HashDomain::new(
+                    chip1.clone(),
+                    ecc_chip.clone(),
```
daira, line 321: Fix the clippy lints.

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,318 @@
...
+                // y_{A,i+1} = [(λ_{1,i+1} + λ_{2,i+1})⋅(x_{A,i+1} − (λ_{1,i+1}^2 − x_{A,i+1} − x_{P,i+1}))] / 2
+                let y_a_next = (lambda1_next + lambda2_next)
+                    * (x_a_next.clone() - (lambda1_next_sq - x_a_next.clone() - x_p_next))
+                    * pallas::Base::TWO_INV;
+
+                // 2⋅λ_{2,i}⋅(x_{A,i} − x_{A,i+1}) − (y_{A,i} + y_{A,i+1})
+                q_sinsemilla2 * (lambda2_cur * (x_a_cur - x_a_next) - (y_a_cur + y_a_next))
+            };
+
+            vec![expr2]
```
daira, line 144: I have not checked these constraints yet.

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,318 @@
...
+        mut layouter: impl Layouter<pallas::Base>,
+        message: Vec<Option<bool>>,
+    ) -> Result<Self::Message, Error> {
+        // Message must be composed of `K`-bit words.
+        assert_eq!(message.len() % K, 0);
+
+        // Message must have at most `MAX_WORDS` words.
+        assert!(message.len() / K <= MAX_WORDS);
+
+        // Message piece must be at most `ceil(pallas::Base::NUM_BITS / 10)` bits
```
daira, line 186: This code is parameterized over K, so shouldn't say 10 in the comment.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,368 @@
...
+                config.word,
+                offset,
+                || piece.field_elem().ok_or(Error::SynthesisError),
+            )?;
+            region.constrain_equal(&config.perm, piece.cell(), cell)?;
+            zs.push(CellValue::new(cell, piece.field_elem()));
+
+            // Assign cumulative sum such that
+            //          z_i = 2^10 * z_{i + 1} + m_{i + 1}
+            // => z_{i + 1} = (z_i - m_{i + 1}) / 2^10
```
daira, line 200: Again, the comments shouldn't "hard-code" k = 10.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,368 @@
...
+        let generators: Vec<Option<(pallas::Base, pallas::Base)>> =
+            if let Some(generators) = generators {
+                generators.into_iter().map(Some).collect()
+            } else {
+                vec![None; piece.num_words()]
+            };
+
+        for (row, gen) in generators.iter().enumerate() {
+            let x_p = gen.map(|gen| gen.0);
+            let y_p = gen.map(|gen| gen.1);
```
daira, line 245: ``let (x_p, y_p) = gen.map(|gen| (gen.0, gen.1));``

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,368 @@
...
+                )?;
+
+                lambda1
+            };
+
+            // Compute `x_r`
+            let x_r = lambda1
+                .zip(x_a.value())
+                .zip(x_p)
+                .map(|((lambda1, x_a), x_p)| lambda1 * lambda1 - x_a - x_p);
```
daira, line 279:
```suggestion
                .map(|((lambda1, x_a), x_p)| lambda1.square() - x_a - x_p);
```

str4d:
Test circuit layout as of 8d3b5f0271590cf4f626ba35e8bebe838d32fc98. There are two irregularities inside `hash_to_point`:

- The first row of the `x_a` column is assigned to more than once; I think this occurs at the start of a message piece, as we see it three times in the final invocation.
- In the final invocation, we additionally see repeat assignments in the `word` column at what looks like the boundary between message pieces.

sinsemilla-hash-layout-notext

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,301 @@
...
+            y_a = a.1;
+        }
+
+        // Assign the final `y_a`.
+        let y_a_cell = region.assign_advice(
+            || "y_a",
+            config.word,
+            offset + 1,
+            || y_a.ok_or(Error::SynthesisError),
+        )?;
```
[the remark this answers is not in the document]
daira: `y_a` is assigned from the return value of `hash_piece` in the last step.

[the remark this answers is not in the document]
daira: The x-coordinate is not sufficient to define the point, because sign is significant. It should be constrained by the last constraint in https://hackmd.io/iOw7-HpFQY6dPF1aFY8pAw#Constraint-program :

λ_{2,n} ⋅ (x_{A,n} − x_{A,n+1}) = (λ_{1,n} + λ_{2,n}) ⋅ (x_{A,n} − (λ_{1,n}^2 − x_{A,n} − x_{P,n})) + y_{A,n+1}

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,318 @@
...
+        let x_p = advices[2];
+        let lambda = (advices[3], advices[4]);
+
+        // Generator table config
+        let generator_table =
+            GeneratorTableConfig::configure(meta, q_sinsemilla1, lookup, word, x_a, x_p, lambda);
+
+        // Create custom gates
+        meta.create_gate("Sinsemilla gate 1", |meta| {
+            // (λ_{2,i})^2 − (x_{A,i+1} + (λ_{1,i}^2 − x_{A,i} − x_{P,i}) + x_{A,i})
```
daira, line 99:
```suggestion
            // (λ_{2,i})^2 − (x_{A,i+1} + (λ_{1,i}^2 − x_{A,i} − x_{P,i}) + x_{A,i}) = 0
```

str4d:
> * The first row of the `x_a` column is assigned to more than once; I think this occurs at the start of a message piece, as we see it three times in the final invocation.

The first of these occurs due to these two assignments:

https://github.com/zcash/orchard/blob/8d3b5f0271590cf4f626ba35e8bebe838d32fc98/src/circuit/gadget/sinsemilla/chip/hash_to_point.rs#L35-L36

https://github.com/zcash/orchard/blob/8d3b5f0271590cf4f626ba35e8bebe838d32fc98/src/circuit/gadget/sinsemilla/chip/hash_to_point.rs#L223-L232

In combination with my earlier review comment (https://github.com/zcash/orchard/pull/67#discussion_r650607724), I think the first of these should be assigning to a fixed column that is part of `config.perm`. The `hash_piece` subregion will correctly copy that into the advice column.

The repeat assignment to `x_a` between message pieces occurs due to these two assignments overlapping (which is fine, because they are assigned to the same value):

https://github.com/zcash/orchard/blob/8d3b5f0271590cf4f626ba35e8bebe838d32fc98/src/circuit/gadget/sinsemilla/chip/hash_to_point.rs#L299-L314

https://github.com/zcash/orchard/blob/8d3b5f0271590cf4f626ba35e8bebe838d32fc98/src/circuit/gadget/sinsemilla/chip/hash_to_point.rs#L49-L51

https://github.com/zcash/orchard/blob/8d3b5f0271590cf4f626ba35e8bebe838d32fc98/src/circuit/gadget/sinsemilla/chip/hash_to_point.rs#L223-L232

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,318 @@
...
+            let expr1 = {
+                let q_sinsemilla1 = meta.query_selector(q_sinsemilla1);
+                let lambda1 = meta.query_advice(lambda.0, Rotation::cur());
+                let lambda1_sq = lambda1.clone() * lambda1;
+                let lambda2 = meta.query_advice(lambda.1, Rotation::cur());
+                let lambda2_sq = lambda2.clone() * lambda2;
+                let x_a_next = meta.query_advice(x_a, Rotation::next());
+                let x_p_cur = meta.query_advice(x_p, Rotation::cur());
+
+                q_sinsemilla1 * (lambda2_sq - (x_a_next + (lambda1_sq - x_p_cur)))
```
daira, line 109:
```suggestion
                q_sinsemilla1 * (lambda2.square() - (x_a_next + (lambda1.square() - x_p_cur)))
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -94,6 +94,56 @@ impl<const K: usize, const MAX_WORDS: usize> SinsemillaChip<K, MAX_WORDS> {
...
 
+        // Create custom gates
+        meta.create_gate("Sinsemilla gate 1", |meta| {
+            // (λ_{2,i})^2 − (x_{A,i+1} + (λ_{1,i}^2 − x_{A,i} − x_{P,i}) + x_{A,i})
+            let expr1 = {
+                let q_sinsemilla1 = meta.query_selector(q_sinsemilla1);
+                let lambda1 = meta.query_advice(lambda.0, Rotation::cur());
+                let lambda1_sq = lambda1.clone() * lambda1;
+                let lambda2 = meta.query_advice(lambda.1, Rotation::cur());
+                let lambda2_sq = lambda2.clone() * lambda2;
```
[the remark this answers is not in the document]
daira:
```suggestion
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -94,6 +94,56 @@ impl<const K: usize, const MAX_WORDS: usize> SinsemillaChip<K, MAX_WORDS> {
         let generator_table =
             GeneratorTableConfig::configure(meta, q_sinsemilla1, lookup, word, x_a, x_p, lambda);
 
+        // Create custom gates
+        meta.create_gate("Sinsemilla gate 1", |meta| {
+            // (λ_{2,i})^2 − (x_{A,i+1} + (λ_{1,i}^2 − x_{A,i} − x_{P,i}) + x_{A,i})
+            let expr1 = {
+                let q_sinsemilla1 = meta.query_selector(q_sinsemilla1);
+                let lambda1 = meta.query_advice(lambda.0, Rotation::cur());
+                let lambda1_sq = lambda1.clone() * lambda1;
```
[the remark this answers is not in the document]
daira:
```suggestion
```

str4d:
> * In the final invocation, we additionally see repeat assignments in the `word` column at what looks like the boundary between message pieces.

These repeat assignments occur due to the running sums from adjacent pieces overlapping:

https://github.com/zcash/orchard/blob/8d3b5f0271590cf4f626ba35e8bebe838d32fc98/src/circuit/gadget/sinsemilla/chip/hash_to_point.rs#L211-L216

https://github.com/zcash/orchard/blob/8d3b5f0271590cf4f626ba35e8bebe838d32fc98/src/circuit/gadget/sinsemilla/chip/hash_to_point.rs#L49-L51

https://github.com/zcash/orchard/blob/8d3b5f0271590cf4f626ba35e8bebe838d32fc98/src/circuit/gadget/sinsemilla/chip/hash_to_point.rs#L188-L194

AFAICT this is a bug: the running sum for each message piece is initialized (`z_0`) to the value of that piece. This causes the running sum to track the words within that piece correctly, and the sum reaches zero at its end. The next piece's `z_0` overwrites the previous piece's zero. The reason the test circuit is not failing is that the running sum lookup gate is disabled at the boundary (as I noted in my review here: https://github.com/zcash/orchard/pull/67#discussion_r650611507).

I believe what should instead be happening is that `hash_piece` should require that the caller initializes `z_0` at the correct offset, and passes in its value (not cell). We then need to enable `config.q_sinsemilla1` for the last row of every piece, replacing this special-case for the final piece:

https://github.com/zcash/orchard/blob/8d3b5f0271590cf4f626ba35e8bebe838d32fc98/src/circuit/gadget/sinsemilla/chip/hash_to_point.rs#L107-L108

src/circuit/gadget/sinsemilla/chip/generator_table.rs
```diff
@@ -0,0 +1,122 @@
...
+            // m_{i+1} = z_{i} - 2^K * z_{i + 1}
+            let bits_cur = meta.query_advice(bits, Rotation::cur());
+            let bits_next = meta.query_advice(bits, Rotation::next());
+            let bits = bits_cur - (bits_next * C::Base::from_u64(1 << K));
+
+            let x_a_cur = meta.query_advice(x_a, Rotation::cur());
+            let x_p_cur = meta.query_advice(x_p, Rotation::cur());
+            let lambda1_cur = meta.query_advice(lambda.0, Rotation::cur());
+            let lambda2_cur = meta.query_advice(lambda.1, Rotation::cur());
+            let y_a_cur = (lambda1_cur.clone() + lambda2_cur)
```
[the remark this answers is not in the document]
therealyingtong: Fixed this in the book PR: https://github.com/zcash/orchard/pull/82/commits/beaf4490f1f30cc6db8a15b6457c492df90b4c3c

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,369 @@
...
+            let (x, y, zs) = self.hash_piece(region, offset, piece, x_a, y_a)?;
+
+            // Since each message word takes one row to process, we increase
+            // the offset by `piece.num_words` on each iteration.
+            offset += piece.num_words() + 1;
+
+            // Update the accumulator to the latest hash value.
+            x_a = x;
+            y_a = y;
+            zs_sum.extend_from_slice(&zs);
```
str4d, line 56: Now that I understand the `zs` returned from `hash_piece` is intended to be disjoint for each piece, this should not be extending `zs_sum` (because that leaves us with a single `Vec` containing all of the pieces' running sums, which then need to be correctly split apart again). Instead, make `zs_sum` a `Vec<Vec<CellValue<_>>>` so the caller gets each running sum independently.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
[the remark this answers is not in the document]
therealyingtong: This constraint is called "Sinsemilla gate 2" in this PR.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,301 @@
...
+                    .chunks(K)
+                    .fold(Q.to_curve(), |acc, chunk| (acc + S(chunk)) + acc);
+                let actual_point =
+                    pallas::Affine::from_xy(x_a.value().unwrap(), y_a.unwrap()).unwrap();
+                assert_eq!(expected_point.to_affine(), actual_point);
+            }
+        }
+
+        // Enable `Sinsemilla` selector on the last double-and-add row for expr1
+        config.q_sinsemilla1.enable(region, offset - 1)?;
```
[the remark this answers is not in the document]
therealyingtong: This has been corrected in https://github.com/zcash/orchard/pull/67/commits/ea7cdd7dc983176be678df67e7816ad062b23157.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,301 @@
...
+        region: &mut Region<'_, pallas::Base>,
+        offset: usize,
+        piece: &<Self as SinsemillaInstructions<pallas::Affine, K, MAX_WORDS>>::MessagePiece,
+        x_a: CellValue<pallas::Base>,
+        y_a: Option<pallas::Base>,
+    ) -> Result<(CellValue<pallas::Base>, Option<pallas::Base>), Error> {
+        let config = self.config().clone();
+
+        // Enable `Sinsemilla` selectors for expr1 and expr2
+        for row in 0..(piece.num_words() - 1) {
```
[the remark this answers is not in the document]
therealyingtong: This has been corrected in ea7cdd7.

src/circuit/gadget/sinsemilla/chip.rs
therealyingtong: It's the degree 3 constraint here (the `x_{A,i}`'s cancel out):
image

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,301 @@
...
+        message: &<Self as SinsemillaInstructions<pallas::Affine, K, MAX_WORDS>>::Message,
+    ) -> Result<EccPoint, Error> {
+        let config = self.config().clone();
+
+        // Get the `x`- and `y`-coordinates of the starting `Q` base.
+        let x_q = *Q.coordinates().unwrap().x();
+        let y_q = *Q.coordinates().unwrap().y();
+
+        // Assign the `x`-coordinate of the starting `Q` base.
+        let x_q_cell = region.assign_advice(|| "x_q", config.x_a, 0, || Ok(x_q))?;
```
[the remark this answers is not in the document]
therealyingtong: A constraint has been added in https://github.com/zcash/orchard/pull/67/commits/5a3fefadd1c782cd4edc6ca8e5eb251613a563b5.

src/circuit/gadget/sinsemilla/message.rs
```diff
@@ -0,0 +1,146 @@
...
+    let bytearray: Vec<u8> = bits
+        .chunks_exact(8)
+        .map(|byte| byte.iter().rev().fold(0u8, |acc, bit| acc * 2 + *bit as u8))
+        .collect();
+
+    F::from_bytes(&bytearray.try_into().unwrap()).unwrap()
+}
+
+#[derive(Clone, Debug)]
+pub struct MessageSubPiece<F: FieldExt, const K: usize> {
```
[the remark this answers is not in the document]
therealyingtong: `MessageSubPiece` is used in later PRs that input field elements to SinsemillaHash. I've removed it from this PR and will re-introduce it in #98.

src/circuit/gadget/sinsemilla/message.rs
```diff
@@ -0,0 +1,146 @@
...
+        self.cell
+    }
+
+    pub fn field_elem(&self) -> Option<F> {
+        self.field_elem
+    }
+
+    // Assembles a field element encoding a few subpieces and the number of
+    // words encoded by them.
+    pub fn from_subpieces(subpieces: &[MessageSubPiece<F, K>], num_words: usize) -> Option<F> {
```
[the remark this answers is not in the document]
therealyingtong: I will make this change in #98 since `MessageSubPiece` has been removed from this PR.

str4d (changes requested):

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,378 @@
...
+        let mut zs_sum: Vec<Vec<CellValue<pallas::Base>>> = Vec::new();
+
+        // Hash each piece in the message.
+        let mut offset = 0;
+        for piece in message.0.iter() {
+            // The value of the hash after this piece is processed.
+            let (x, y, zs) = self.hash_piece(region, offset, piece, x_a, y_a)?;
+
+            // Since each message word takes one row to process, we increase
+            // the offset by `piece.num_words` on each iteration.
```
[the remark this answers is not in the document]
therealyingtong: I have clarified the +1 offset in a separate step just below this.

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,372 @@
...
+    ) -> <Self as Chip<pallas::Base>>::Config {
+        let config = SinsemillaConfig {
+            q_sinsemilla1: meta.selector(),
+            q_sinsemilla2: meta.fixed_column(),
+            fixed_y_q: meta.fixed_column(),
+            x_a: advices[0],
+            x_p: advices[1],
+            bits: advices[2],
+            lambda_1: advices[3],
+            lambda_2: advices[4],
```
daira, line 116: I changed the advice columns in #82 to be the same as here.

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,372 @@
...
+            let Y_A = {
+                let lambda_1 = meta.query_advice(config.lambda_1, Rotation::cur());
+                let lambda_2 = meta.query_advice(config.lambda_2, Rotation::cur());
+                let x_a = meta.query_advice(config.x_a, Rotation::cur());
+
+                // x_r = lambda_1^2 - x_a - x_p
+                let x_r = {
+                    let x_p = meta.query_advice(config.x_p, Rotation::cur());
+                    lambda_1.clone().square() - x_a.clone() - x_p
+                };
```
daira, line 141: Does it work to move the duplicated code outside the `create_gate` calls?

daira: Oh, it's not the same `meta`.

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,372 @@
...
+
+                (lambda_1 + lambda_2) * (x_a - x_r)
+            };
```
str4d, line 144:
```suggestion
            let Y_A = Y_A(meta, Rotation::cur());
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,372 @@
...
+            // Y_A = (lambda_1 + lambda_2) * (x_a - x_r)
+            let Y_A_next = {
+                // x_r = lambda_1^2 - x_a - x_p
+                let x_r = {
+                    let x_p = meta.query_advice(config.x_p, Rotation::next());
+                    lambda_1_next.clone().square() - x_a_next.clone() - x_p
+                };
+
+                (lambda_1_next.clone() + lambda_2_next) * (x_a_next.clone() - x_r)
+            };
```
str4d, line 203:
```suggestion
            let Y_A_cur = Y_A(meta, Rotation::cur());
            let Y_A_next = Y_A(meta, Rotation::next());
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,372 @@
...
+        meta.create_gate("Sinsemilla gate", |meta| {
+            let q_s1 = meta.query_selector(config.q_sinsemilla1);
+            let q_s2 = meta.query_fixed(config.q_sinsemilla2, Rotation::cur());
+            let q_s3 = {
+                let one = Expression::Constant(pallas::Base::one());
+                q_s2.clone() * (q_s2 - one)
+            };
+            let x_a_cur = meta.query_advice(config.x_a, Rotation::cur());
+            let x_a_next = meta.query_advice(config.x_a, Rotation::next());
+            let lambda_1_cur = meta.query_advice(config.lambda_1, Rotation::cur());
```
str4d, line 178: With the `Y_A` refactor, this becomes unused:
```suggestion
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,372 @@
...
+            let lambda_2_cur = meta.query_advice(config.lambda_2, Rotation::cur());
+            let lambda_1_next = meta.query_advice(config.lambda_1, Rotation::next());
+            let lambda_2_next = meta.query_advice(config.lambda_2, Rotation::next());
```
str4d, line 181: With the `Y_A` refactor, this becomes unused:
```suggestion
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,372 @@
...
+            let lambda_1 = meta.query_advice(config.lambda_1, Rotation::cur());
+            let lambda_2 = meta.query_advice(config.lambda_2, Rotation::cur());
+            let x_a_cur = meta.query_advice(config.x_a, Rotation::cur());
+            let x_a_next = meta.query_advice(config.x_a, Rotation::next());
+
+            // x_r = lambda_1^2 - x_a_cur - x_p
+            let x_r = {
+                let x_p = meta.query_advice(config.x_p, Rotation::cur());
+                lambda_1.square() - x_a_cur.clone() - x_p
+            };
```
str4d, line 161:
```suggestion
            let x_r = x_r(meta, Rotation::cur());
```

str4d (commented):
Partial review.

src/circuit/gadget/sinsemilla/chip.rs
str4d: It can work, see my review.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+        let x_q = *Q.coordinates().unwrap().x();
+        let y_q = *Q.coordinates().unwrap().y();
+
+        // Initialize the accumulator to `Q`.
+        let (mut x_a, mut y_a): (X<pallas::Base>, Y<pallas::Base>) = {
+            // Constrain the initial x_q to equal the x-coordinate of the domain's `Q`.
+            let fixed_x_q =
+                region.assign_fixed(|| "fixed x_q", config.constants, offset, || Ok(x_q))?;
+            let x_q_cell = region.assign_advice(|| "x_q", config.x_a, offset, || Ok(x_q))?;
+            region.constrain_equal(&config.perm, fixed_x_q, x_q_cell)?;
```
daira, line 47: It seems like there should be a more concise way to constrain a new cell equal to an existing one, without having to make two calls one of which repeats the value.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+
+        // Get the `x`- and `y`-coordinates of the starting `Q` base.
+        let x_q = *Q.coordinates().unwrap().x();
+        let y_q = *Q.coordinates().unwrap().y();
+
+        // Initialize the accumulator to `Q`.
+        let (mut x_a, mut y_a): (X<pallas::Base>, Y<pallas::Base>) = {
+            // Constrain the initial x_q to equal the x-coordinate of the domain's `Q`.
+            let fixed_x_q =
+                region.assign_fixed(|| "fixed x_q", config.constants, offset, || Ok(x_q))?;
```
daira, line 45: The constant doesn't have to be assigned in the same region or at the same offset; if it is, then it is likely that one fixed could will not be sufficient. There should be a utility to assign a constant at the next unused cell in that column.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+
+            // Since each message word takes one row to process, we increase
+            // the offset by `piece.num_words` on each iteration.
+            offset += piece.num_words();
+
+            // Update the accumulator to the latest value.
+            x_a = x;
+            y_a = y;
+            zs_sum.push(zs);
+
```
str4d, line 92: This is identical to the main loop, and AFAICT none of the temporaries generated in this section are used below, so we should just do this in the main loop (but keep the rest of this section for assigning the final `y_a`:
```suggestion
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+            // Assign and constrain the final `y_a`.
+            region.assign_fixed(
+                || "qs_2 = 2 on final row",
+                config.q_sinsemilla2,
+                offset - 1,
+                || Ok(pallas::Base::from_u64(2)),
+            )?;
```
str4d, line 99: This causes a double-assignment, as it is (correctly) overwriting the final `0` assigned inside `hash_piece`:
image

We should pass a `last` boolean flag into `hash_piece` (which we can detect in the main loop), and then set `qs_2` for the last piece to either `0` or `2` based on that flag.
```suggestion
            // Assign and constrain the final `y_a`.
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+        &self,
+        region: &mut Region<'_, pallas::Base>,
+        offset: usize,
+        piece: &<Self as SinsemillaInstructions<
+            pallas::Affine,
+            { sinsemilla::K },
+            { sinsemilla::C },
+        >>::MessagePiece,
+        x_a: X<pallas::Base>,
+        y_a: Y<pallas::Base>,
```
str4d, line 183:
```suggestion
        y_a: Y<pallas::Base>,
        last_piece: bool,
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+                    || "q_s2 = 1",
+                    config.q_sinsemilla2,
+                    offset + row,
+                    || Ok(pallas::Base::one()),
+                )?;
+            }
+
+            // Set `q_sinsemilla2` fixed column to 0 on the last row.
+            region.assign_fixed(
+                || "q_s2 = 1",
```
str4d, line 213:
```suggestion
                || if last_piece { "qs_2 for final piece" } else { "q_s2 between pieces" },
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+                config.q_sinsemilla2,
+                offset + piece.num_words() - 1,
+                || Ok(pallas::Base::zero()),
```
str4d, line 216:
```suggestion
                || Ok(if last_piece { pallas::Base::from_u64(2) } else { pallas::Base::zero() }),
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+            vec![None; piece.num_words()]
+        };
+
+        // Decompose message into `K`-bit pieces with a running sum `z`.
+        let zs = {
+            let mut zs = Vec::with_capacity(piece.num_words() + 1);
+
+            // Copy message and initialize running sum `z` to decompose message in-circuit
+            let cell = region.assign_advice(
+                || "z_0 (copy of message)",
```
str4d, line 266:
```suggestion
                || "z_0 (copy of message piece)",
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+            )?;
+
+            // Compute and assign `lambda_1`
+            let lambda_1 = {
+                let lambda_1 = x_a
+                    .value()
+                    .zip(y_a.0)
+                    .zip(x_p)
+                    .zip(y_p)
+                    .map(|(((x_a, y_a), x_p), y_p)| (y_a - y_p) * (x_a - x_p).invert().unwrap());
```
str4d, line 340: Now that https://github.com/zcash/halo2/pull/293 has merged, we can batch these inversions.

Do not make this change in this PR; I want to do this as a separate optimization (I'm just noting it here so we remember).

str4d (changes requested):

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,372 @@
...
+                };
+
+                (lambda_1 + lambda_2) * (x_a - x_r)
+            };
+
+            // fixed_y_q * (2 * fixed_y_q - Y_{A,0}) = 0
+            vec![fixed_y_q.clone() * (two.clone() * fixed_y_q - Y_A)]
+        });
+
+        meta.create_gate("Secant line", |meta| {
```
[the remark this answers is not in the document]
daira: If it's the labelling, remember that https://github.com/zcash/halo2/pull/286 allows labelling constraints of a gate individually.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
daira: What's the extra cell in the fourth row that is the only one in the sixth column?

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+
+            let y_a_cell = region.assign_advice(
+                || "y_a",
+                config.lambda_1,
+                offset,
+                || y_a.ok_or(Error::SynthesisError),
+            )?;
+
+            // Assign lambda_2 and x_p zero values since they are queried
+            // in the gate.
```
daira, line 109: I don't understand this.

str4d: The last gate queries these values, even though in this case they are unconstrained (since we are instead constraining the y<sub>A,n</sub> output). So we need to assign them to something, or else `MockProver` complains that an active gate isn't fully assigned.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
str4d: That's `fixed_y_A`. The first three columns in the screenshot are the lookup table, and the other two are `q_sinsemilla1` (a `Selector`, which now has a darker blue background) and `q_sinsemilla2`.

The impl isn't setting `fixed_y_A` to zero on every other row because we aren't requiring it. `MockProver` can't detect that this gate is active because it doesn't involve a selector, and `MockProver` can't make any assumptions about other gates (which could be arbitrarily more complex).

daira: I think that the case of "any non-zero value means selected" is common enough to support in `MockProver`.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
daira: Oh, I see, they're multiplied by zero.

daira:
```suggestion
            // Assign lambda_2 and x_p zero values since they are queried
            // in the gate. (The actual values do not matter since they are
            // multiplied by zero.)
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+                            .take(K * piece.num_words())
+                            .collect::<Vec<_>>()
+                    })
+                    .flatten()
+                    .collect();
+
+                let hasher_S = pallas::Point::hash_to_curve(S_PERSONALIZATION);
+                let S = |chunk: &[bool]| hasher_S(&lebs2ip_k(chunk).to_le_bytes());
+
+                let expected_point = bitstring
```
daira, line 159:
```suggestion
                // We can use complete addition here because it differs from
                // incomplete addition with negligible probability.
                let expected_point = bitstring
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+            for row in 0..(piece.num_words() - 1) {
+                region.assign_fixed(
+                    || "q_s2 = 1",
+                    config.q_sinsemilla2,
+                    offset + row,
+                    || Ok(pallas::Base::one()),
+                )?;
+            }
+
+            // Set `q_sinsemilla2` fixed column to 0 on the last row.
```
daira, line 211:
```suggestion
            // Set `q_sinsemilla2` fixed column to 0 on the last row if this is not
            // the last piece, or to 2 on the last row of the last piece.
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
daira: Use `final_piece`, to be consistent with comments and to distinguish from "last row".

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+        });
+
+        // Get (x_p, y_p) for each word. We precompute this here so that we can use `batch_normalize()`.
+        let generators_projective: Option<Vec<pallas::Point>> = words
+            .clone()
+            .map(|words| words.iter().map(|word| get_s_by_idx(*word)).collect());
+        let generators: Option<Vec<(pallas::Base, pallas::Base)>> =
+            generators_projective.map(|generators_projective| {
+                let mut generators = vec![pallas::Affine::default(); generators_projective.len()];
+                pallas::Point::batch_normalize(&generators_projective, &mut generators);
```
daira, line 243: Rather than batch-normalizing the generators each time we use them, we should batch normalize them once and for all when the table is computed.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+            let cell = region.assign_advice(
+                || "z_0 (copy of message)",
+                config.bits,
+                offset,
+                || piece.field_elem().ok_or(Error::SynthesisError),
+            )?;
+            region.constrain_equal(&config.perm, piece.cell(), cell)?;
+            zs.push(CellValue::new(cell, piece.field_elem()));
+
+            // Assign cumulative sum such that
```
daira, line 274:
```suggestion
            // Assign cumulative sum such that for 0 <= i < n,
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+            //          z_i = 2^K * z_{i + 1} + m_{i + 1}
+            // => z_{i + 1} = (z_i - m_{i + 1}) / 2^K
+            //
+            // For a message m = m_1 + 2^K m_2 + ... + 2^{K(n-1)} m_n}, initialize z_0 = m.
+            // We end up with z_n = 0.
```
daira, line 279:
```suggestion
            // We end up with z_n = 0. (z_n is not directly encoded as a cell value;
            // it is implicitly taken as 0 by adjusting the definition of m_{i+1}.)
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,446 @@
...
+
+                lambda_2
+            };
+
+            // Compute and assign `x_a` for the next row.
+            let x_a_new: X<pallas::Base> = {
+                let x_a_new = lambda_2
+                    .zip(x_a.value())
+                    .zip(x_r)
+                    .map(|((lambda_2, x_a), x_r)| lambda_2 * lambda_2 - x_a - x_r);
```
daira, line 382:
```suggestion
                    .map(|((lambda_2, x_a), x_r)| lambda_2.square() - x_a - x_r);
```

daira (approved):
utACK modulo @str4d's comments (some of which are blocking).

src/circuit/gadget/sinsemilla.rs
```diff
@@ -0,0 +1,411 @@
...
+    /// sum.
+    ///
+    /// A cumulative sum `z` is used to decompose a Sinsemilla message. It
+    /// produces intermediate values for each word in the message, such
+    /// that `z_next` = (`z_cur` - `word_next`) / `2^K`.
+    ///  
+    /// These intermediate values are useful for range checks on subsets
+    /// of the Sinsemilla message. Sinsemilla messages in the Orchard
+    /// protocol are composed of field elements, and we need to check
+    /// the canonicity of the field element encodings in certain cases.
```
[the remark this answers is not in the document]
therealyingtong: `[z_0, ..., z_{n-1}]` is returned. `z_n` is no longer returned because it is now constrained to be zero.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
therealyingtong: Oh yeah, i wrote a `utilities::copy()` for precisely this reason.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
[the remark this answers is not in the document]
therealyingtong: Resolved by the new constraints.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
therealyingtong: Opened up https://github.com/zcash/halo2/issues/301 to look into this.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
therealyingtong: Now using `utilities::copy()`.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
therealyingtong: Is it worth just precomputing and storing the generators in `primitives::sinsemilla::constants`?

str4d: Probably, but let's try that after this PR.

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,365 @@
...
+pub struct SinsemillaConfig {
+    // Selector used in the lookup argument as well as Sinsemilla custom gates.
+    q_sinsemilla1: Selector,
+    // Fixed column used in Sinsemilla custom gates.
+    q_sinsemilla2: Column<Fixed>,
+    // Fixed column used to constrain hash initialization to be consistent with
+    // the y-coordinate of the domain Q.
+    fixed_y_q: Column<Fixed>,
+    // Advice column used to store the x-coordinate of the accumulator at each
+    // iteration of the hash.
```
str4d, line 46:
```suggestion
    /// Advice column used to store the x-coordinate of the accumulator at each
    /// iteration of the hash.
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,365 @@
...
+    x_a: Column<Advice>,
+    // Advice column used to store the x-coordinate of the generator corresponding
+    // to the message word at each iteration of the hash. This is looked up in the
+    // generator table.
```
str4d, line 50:
```suggestion
    /// Advice column used to store the x-coordinate of the generator corresponding
    /// to the message word at each iteration of the hash. This is looked up in the
    /// generator table.
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,365 @@
...
+    x_p: Column<Advice>,
+    // Advice column used to load the message.
```
str4d, line 52:
```suggestion
    /// Advice column used to load the message.
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,365 @@
...
+    bits: Column<Advice>,
+    // Advice column used to store the lambda_1 intermediate value at each
+    // iteration.
```
str4d, line 55:
```suggestion
    /// Advice column used to store the $\lambda_1$ intermediate value at each
    /// iteration.
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,365 @@
...
+    lambda_1: Column<Advice>,
+    // Advice column used to store the lambda_2 intermediate value at each
+    // iteration.
```
str4d, line 58:
```suggestion
    /// Advice column used to store the $\lambda_2$ intermediate value at each
    /// iteration.
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,365 @@
...
+    lambda_2: Column<Advice>,
+    // The lookup table where (idx, x_p, y_p) are loaded for the 2^K generators
+    // of the Sinsemilla hash.
```
str4d, line 61:
```suggestion
    /// The lookup table where $(\mathsf{idx}, x_p, y_p)$ are loaded for the $2^K$
    /// generators of the Sinsemilla hash.
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,365 @@
...
+    generator_table: GeneratorTableConfig,
+    // Fixed column shared by the whole circuit. This is used to load the
+    // x-coordinate of the domain Q, which is then constrained to equal the
+    // initial x_a.
```
str4d, line 65:
```suggestion
    /// Fixed column shared by the whole circuit. This is used to load the
    /// x-coordinate of the domain $Q$, which is then constrained to equal the
    /// initial $x_a$.
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,365 @@
...
+    constants: Column<Fixed>,
+    // Permutation over all advice columns and the `constants` fixed column.
```
str4d, line 67:
```suggestion
    /// Permutation over all advice columns and the `constants` fixed column.
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,365 @@
...
+        meta.create_gate("Initial y_Q", |meta| {
+            let fixed_y_q = meta.query_fixed(config.fixed_y_q, Rotation::cur());
+
+            // Y_A = (lambda_1 + lambda_2) * (x_a - x_r)
+            let Y_A = Y_A(meta, Rotation::cur());
+
+            // fixed_y_q * (2 * fixed_y_q - Y_{A,0}) = 0
+            vec![fixed_y_q.clone() * (two.clone() * fixed_y_q - Y_A)]
+        });
+
```
str4d, line 162: If we move this inside the "Sinsemilla gate", then `MockProver` will see the `fixed_y_q` query as semantically connected to `q_sinsemilla1` (which it is - we only want to use this constraint specifically in the context of Sinsemilla), and ensure that this constraint is assigned correctly. It _would_ mean that we need to explicitly assign zero to every row except the first within `SinsemillaChip::hash_message`, but I think that's fine.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,447 @@
...
+
+        Ok((EccPoint::from_coordinates_unchecked(x_a.0, y_a), zs_sum))
+    }
+
+    #[allow(clippy::type_complexity)]
+    // Hash a message piece containing `piece.length` number of `K`-bit words.
+    // To avoid a duplicate assignment, the accumulator x-coordinate provided
+    // by the caller is not copied. This only works because `hash_piece()` is
+    // an internal API. Before this call to `hash_piece()`, x_a MUST have been
+    // already assigned within this region at the correct offset.
```
str4d, line 168:
```suggestion
    /// Hashes a message piece containing `piece.length` number of `K`-bit words.
    ///
    /// To avoid a duplicate assignment, the accumulator x-coordinate provided
    /// by the caller is not copied. This only works because `hash_piece()` is
    /// an internal API. Before this call to `hash_piece()`, x_a MUST have been
    /// already assigned within this region at the correct offset.
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,447 @@
...
+            // Update the mutable `x_a`, `y_a` variables.
+            x_a = x_a_new;
+            y_a = y_a_new;
+        }
+
+        Ok((x_a, y_a, zs))
+    }
+}
+
+// The x-coordinate of the accumulator in a Sinsemilla hash instance.
```
str4d, line 412:
```suggestion
/// The x-coordinate of the accumulator in a Sinsemilla hash instance.
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -0,0 +1,447 @@
...
+
+    fn deref(&self) -> &CellValue<F> {
+        &self.0
+    }
+}
+
+// The y-coordinate of the accumulator in a Sinsemilla hash instance.
+// This is never actually witnessed until the last round, since it
+// can be derived from other variables. Thus it only exists as a field
+// element, not a `CellValue`.
```
str4d, line 432:
```suggestion
/// The y-coordinate of the accumulator in a Sinsemilla hash instance.
///
/// This is never actually witnessed until the last round, since it
/// can be derived from other variables. Thus it only exists as a field
/// element, not a `CellValue`.
```

src/circuit/gadget/sinsemilla/message.rs
```diff
@@ -0,0 +1,71 @@
...
+
+/// A [`MessagePiece`] of some bitlength.
+///
+/// The piece must fit within a base field element, which means its length
+/// cannot exceed the base field's `NUM_BITS`.
+#[derive(Copy, Clone, Debug)]
+pub struct MessagePiece<F: FieldExt, const K: usize> {
+    cell: Cell,
+    field_elem: Option<F>,
+    // The number of K-bit words in this message piece.
```
str4d, line 39:
```suggestion
    /// The number of K-bit words in this message piece.
```

str4d (approved):
utACK. One small suggested refactor; the rest is doc comment fixes.

str4d:
ACK, chip layout LGTM. I noticed that the final row of the `bits` column was not being set; this doesn't trigger the same `MockProver` error as `lambda_2` and `x_p` because the `bits` column is only involved in lookup argument expressions, and `MockProver` currently only checks custom gates. But adding a dummy assignment here later is a backwards-compatible change, as it's an advice column.

sinsemilla-hash-layout

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
therealyingtong: Opened an issue for this https://github.com/zcash/orchard/issues/124

str4d (approved):
ACK bd08808566b5fbeae9b89f0ae00157afee01e617.

sinsemilla-hash-layout

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -183,6 +173,10 @@ impl SinsemillaChip {
             // Y_A = (lambda_1 + lambda_2) * (x_a - x_r)
             let Y_A_next = Y_A(meta, Rotation::next());
 
+            // Check that the initial x_A, x_P, lambda_1, lambda_2 are consistent with y_Q.
+            // fixed_y_q * (2 * fixed_y_q - Y_{A,0}) = 0
+            let init_y_q_check = fixed_y_q.clone() * (two.clone() * fixed_y_q - Y_A_cur.clone());
+
             // lambda2^2 - (x_a_next + x_r + x_a_cur) = 0
             let secant_line =
                 lambda_2_cur.clone().square() - (x_a_next.clone() + x_r + x_a_cur.clone());
```
daira: Technically this is implementing two secant lines, because x_r is itself computed by a secant line. ("Secant" means a line between two distinct points on a curve, extended to infinity.) I think this naming is fine though.

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -207,6 +201,7 @@ impl SinsemillaChip {
             };
 
             vec![
+                ("Initial y_q", init_y_q_check),
                 ("Secant line", q_s1.clone() * secant_line),
                 ("Sinsemilla gate", q_s1 * expr),
```
daira, line 206:
```suggestion
                ("y poly", q_s1 * y_poly),
```

daira (approved):
utACK bd08808566b5fbeae9b89f0ae00157afee01e617 with minor comments.

daira:
> sinsemilla-hash-layout

Why does this show some fixed columns being assigned outside a region? (or am I misinterpreting it?)

str4d:
> Why does this show some fixed columns being assigned outside a region? (or am I misinterpreting it?)

They aren't. Notice those fixed columns line up with equivalent rectangles in the advice columns. This is just an artifact of non-sequential columns being forced to take a sequence for rendering purposes; I split the same region across several boxes. It's clearer when labels are turned on, but in this case the lables overlapped too much and were distracting. It will be easier with an interactive viewer to indicate which boxes are part of the same region.

daira (approved):
I resolved a remaining minor comment in 8af84479b31a45a61f89aead4dd93e32d9f3a7bd.

str4d (approved):
re-ACK 8af84479b31a45a61f89aead4dd93e32d9f3a7bd (trivial change)

PR #70, opened 2021-04-22
Bundle APIs

str4d:


src/bundle.rs
```diff
@@ -50,19 +135,112 @@ pub trait Authorization {
...
 #[derive(Debug)]
 pub struct Bundle<T: Authorization> {
+    /// The list of actions that make up this bundle.
     actions: NonEmpty<Action<T::SpendAuth>>,
-    flag: Flags,
+    /// Orchard-specific transaction-level flags for this bundle.
+    flags: Flags,
+    /// The net value moved into or out of the Orchard shielded pool.
+    ///
+    /// This is the sum of Orchard spends minus the sum Orchard outputs.
```
daira, line 144:
```suggestion
    /// This is the sum of Orchard spends minus the sum of Orchard outputs.
```

src/bundle.rs
```diff
@@ -50,19 +135,112 @@ pub trait Authorization {
 /// A bundle of actions to be applied to the ledger.
 #[derive(Debug)]
 pub struct Bundle<T: Authorization> {
+    /// The list of actions that make up this bundle.
     actions: NonEmpty<Action<T::SpendAuth>>,
-    flag: Flags,
+    /// Orchard-specific transaction-level flags for this bundle.
+    flags: Flags,
+    /// The net value moved into or out of the Orchard shielded pool.
```
daira, line 142:
```suggestion
    /// The net value moved out of the Orchard shielded pool.
```

str4d: It's into the pool if negative.

src/bundle.rs
```diff
@@ -50,19 +135,112 @@ pub trait Authorization {
...
+    }
+
     /// Computes a commitment to the effects of this bundle, suitable for inclusion within
     /// a transaction ID.
     pub fn commitment(&self) -> BundleCommitment {
         todo!()
     }
+
+    /// Transitions this bundle from one authorization state to another.
+    pub fn map<U: Authorization>(
```
daira, line 206:
```suggestion
    pub fn authorize<U: Authorization>(
```

src/bundle.rs
```diff
@@ -50,19 +135,112 @@ pub trait Authorization {
...
+                .map(|a| a.map(|a_auth| spend_auth(&authorization, a_auth))),
+            flags: self.flags,
+            value_balance: self.value_balance,
+            anchor: self.anchor,
+            authorization: step(authorization),
+        }
+    }
+
+    /// Transitions this bundle from one authorization state to another.
+    pub fn try_map<U: Authorization, E>(
```
daira, line 224:
```suggestion
    pub fn try_authorize<U: Authorization, E>(
```

daira (approved):
utACK (with @nuttycom and @therealyingtong).

src/bundle.rs
nuttycom: The rationale for this is that while this is `map`-adjacent, it's not `map` - I think it would be good to have the convention that `map` exclusively means `(a -> b) -> f a -> f b`.

nuttycom: I'm applying this change so that I can base additional work on it; if there's disagreement we can revisit later.

str4d: I disagree because this method does not fully authorize the bundle except in one specific case (where `U` is `Authorized`).

src/bundle.rs
str4d: Ditto my disagreement.

src/bundle.rs
str4d: Opened #71.

PR #72, opened 2021-04-27
Builder preparations

str4d:


src/bundle.rs
```diff
@@ -203,54 +231,48 @@ impl<T: Authorization> Bundle<T> {
...
             flags: self.flags,
             value_balance: self.value_balance,
             anchor: self.anchor,
-            authorization: step(authorization)?,
+            authorization: step(context, authorization)?,
         })
     }
 }
 
-/// Marker for an unauthorized bundle with no proofs or signatures.
```
nuttycom, line 246: Is the idea with this removal that it will be up to the call site to create their own `Unauthorized` type?

nuttycom (approved):
utACK

(one nonblocking question about usage, since I'm currently making use of the `Unauthorized` type.)

src/bundle.rs
str4d: Yes; the builder PR does this for example (its type includes the proof, and the `bsk` necessary to later compute the binding signature).

str4d: If it later becomes apparent that we have a commonly-replicated `Unauthorized` type, we can re-introduce this, but I think it was a premature refactor. Only `Authorized` is a known common requirement.

ebfull:
post-merge ACK

PR #73, opened 2021-04-28
Optimized ECC chip

### LINKED ISSUE #38: Implement generic ECC gadget
Specifically we want something that is generic over a (j = 0) curve. We need:

- In-circuit point representations.
- Complete point addition
- Scalar multiplication (which internally will use incomplete addition, not the above APIs)

therealyingtong:
Closes #38.

Depends on Utils chip (#89).


src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    // Compute R = P + Q
+    let r = x_p
+        .zip(y_p)
+        .zip(x_q)
+        .zip(y_q)
+        .map(|(((x_p, y_p), x_q), y_q)| {
+            let p = C::from_xy(x_p, y_p).unwrap();
+            let q = C::from_xy(x_q, y_q).unwrap();
+            p + q
```
ebfull, line 250: If either p or q are points at infinity then this will fail to `unwrap()`. Check if they're points at infinity first (aka `(0, 0)` in the circuit) and then if they are do `C::zero()` or `C::identity()` (forget which one it is).

This is slightly confusing because the curve implementation rejects `(0, 0)` as a point at infinity (such a point does not exist on the affine plane) but inside the circuit we represent the point at infinity as `(0, 0)`.

ebfull (changes requested):

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,910 @@
...
+#[derive(Copy, Clone, Debug)]
+pub struct OrchardFixedBasesShort<C: CurveAffine>(pub constants::ValueCommitV<C>);
+
+impl<C: CurveAffine> PartialEq for OrchardFixedBases<C> {
+    fn eq(&self, other: &Self) -> bool {
+        format!("{:?}", self) == format!("{:?}", other)
+    }
+}
+
+impl<C: CurveAffine> Eq for OrchardFixedBases<C> {}
```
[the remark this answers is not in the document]
therealyingtong: We don't need the `Eq, PartialEq` traits for `OrchardFixedBases` anymore. They were previously being used in a `BTreeMap` which has been removed.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,910 @@
...
+struct UShort<F: FieldExt>(pub Vec<WindowUs<F>>);
+
+impl<C: CurveAffine> EccLoaded<C> {
+    fn get(&self, point: OrchardFixedBases<C>) -> OrchardFixedBase<C> {
+        match point {
+            OrchardFixedBases::CommitIvkR(_) => self.commit_ivk_r.clone(),
+            OrchardFixedBases::NoteCommitR(_) => self.note_commit_r.clone(),
+            OrchardFixedBases::NullifierK(_) => self.nullifier_k.clone(),
+            OrchardFixedBases::ValueCommitR(_) => self.value_commit_r.clone(),
+            _ => unreachable!(),
```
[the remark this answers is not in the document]
therealyingtong: I've removed the `ValueCommitV` variant along with the `unreachable!()` line.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,910 @@
...
+    /// Witness full-width scalar for fixed-base scalar mul
+    pub q_scalar_fixed: Selector,
+    /// Witness signed short scalar for full-width fixed-base scalar mul
+    pub q_scalar_fixed_short: Selector,
+    /// Copy bits of decomposed scalars
+    pub perm_bits: Permutation,
+    /// Copy between (x_p, y_p) and (x_a, y_a)
+    pub perm_sum: Permutation,
+    /// Copy pre-loaded fixed values to fixed columns
+    pub perm_fixed: Permutation,
```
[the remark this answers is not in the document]
therealyingtong: I'm now using one permutation with the minimal set of columns required.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,910 @@
...
+    commit_ivk_r: OrchardFixedBase<C>,
+    note_commit_r: OrchardFixedBase<C>,
+    nullifier_k: OrchardFixedBase<C>,
+    value_commit_r: OrchardFixedBase<C>,
+    value_commit_v: OrchardFixedBaseShort<C>,
+}
+
+#[derive(Clone, Debug)]
+// 8 coefficients per window
+struct WindowLagrangeCoeffs<F: FieldExt>(pub Vec<CellValue<F>>);
```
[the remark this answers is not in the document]
daira: Or use ArrayVec? It's slightly cleaner I think.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,910 @@
...
+            meta.fixed_column(),
+            meta.fixed_column(),
+            meta.fixed_column(),
+            meta.fixed_column(),
+            meta.fixed_column(),
+            meta.fixed_column(),
+            meta.fixed_column(),
+            meta.fixed_column(),
+        ];
+        let loaded_z = meta.fixed_column();
```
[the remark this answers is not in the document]
daira: I'm a little confused. I thought that the interpolation coefficients were needed as fixed columns. In particular the fixed-base multiplication uses the same gate for each step using each base point, but interpolates over a different 8-element window table.

If these are not real columns, then we're doing a complicated optimization to reduce the number of columns (from 16 to 9) that we didn't need to do.

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +378,168 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+            let bits = meta.advice_column();
+            let P = (meta.advice_column(), meta.advice_column());
+            let lambda = (meta.advice_column(), meta.advice_column());
+            let extras = [
+                meta.advice_column(),
+                meta.advice_column(),
+                meta.advice_column(),
+                meta.advice_column(),
+                meta.advice_column(),
+            ];
```
daira, line 415: Please add documentation of how the extras are used. Perhaps have either constants specifying the offsets into `extras`, or references to them (if that doesn't cause too many borrow checker problems).

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +378,168 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+                let real_added = p_val + q_val;
+                let added_complete = p.add(layouter.namespace(|| "P + Q"), &q)?;
+                if let (Some(x), Some(y)) =
+                    (added_complete.inner.x.value, added_complete.inner.y.value)
+                {
+                    if C::from_xy(x, y).is_some().into() {
+                        assert_eq!(real_added.to_affine(), C::from_xy(x, y).unwrap());
+                    }
+                }
+            }
```
daira, line 448: Also check complete addition of P+P, P+(-P), P+𝒪, 𝒪+P, and 𝒪+𝒪.

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +378,168 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+                let added_incomplete = p.add_incomplete(layouter.namespace(|| "P + Q"), &q)?;
+                if let (Some(x), Some(y)) = (
+                    added_incomplete.inner.x.value,
+                    added_incomplete.inner.y.value,
+                ) {
+                    if C::from_xy(x, y).is_some().into() {
+                        assert_eq!(real_added.to_affine(), C::from_xy(x, y).unwrap());
+                    }
+                }
+            }
```
daira, line 462: Check that trying to perform incomplete addition of P+P, P+(-P), P+𝒪, 𝒪+P, or 𝒪+𝒪 results in errors.

src/circuit/gadget/ecc/chip.rs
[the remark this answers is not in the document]
therealyingtong: When this comment was made, we had two sets of fixed columns: `lagrange_coeffs` and `loaded_lagrange_coeffs`. 

`lagrange_coeffs` are used as you would expect (loaded whenever the circuit performs a fixed-base scalar mul).

`loaded_lagrange_coeffs` were simply "storing" these values and never actually being used in gates, so have been replaced by vectors.

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +378,168 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+                )?;
+                let nullifier_k = super::FixedPoint::get(
+                    chip.clone(),
+                    OrchardFixedBases::NullifierK(nullifier_k),
+                )?;
+                let mul_fixed = nullifier_k.mul(layouter.namespace(|| "mul"), &scalar_fixed)?;
+                if let (Some(x), Some(y)) = (mul_fixed.inner.x.value, mul_fixed.inner.y.value) {
+                    assert_eq!(real_mul_fixed.to_affine(), C::from_xy(x, y).unwrap());
+                }
+            }
```
daira, line 484: Check multiplication by 0.

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +378,168 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+                let scalar_fixed_short = super::ScalarFixedShort::new(
+                    chip.clone(),
+                    layouter.namespace(|| "ScalarFixedShort"),
+                    Some(scalar_fixed_short),
+                )?;
+                let value_commit_v = super::FixedPointShort::get(
+                    chip.clone(),
+                    OrchardFixedBasesShort(value_commit_v),
+                )?;
+                let mul_fixed_short =
```
daira, line 506: Check short multiplication by 0 and by the limit values ±(2<sup>64</sup> - 1).

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +378,168 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+                    value_commit_v.mul(layouter.namespace(|| "mul fixed"), &scalar_fixed_short)?;
+                if let (Some(x), Some(y)) =
+                    (mul_fixed_short.inner.x.value, mul_fixed_short.inner.y.value)
+                {
+                    assert_eq!(real_mul_fixed_short.to_affine(), C::from_xy(x, y).unwrap());
+                }
+            }
```
daira, line 513: Check short multiplication by 0 and by the limit values ±(2<sup>64</sup> - 1).

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +378,168 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+                let scalar = super::ScalarVar::new(
+                    chip,
+                    layouter.namespace(|| "ScalarVar"),
+                    Some(scalar_val),
+                )?;
+                let mul = p.mul(layouter.namespace(|| "mul"), &scalar)?;
+                if let (Some(x), Some(y)) = (mul.inner.x.value, mul.inner.y.value) {
+                    assert_eq!(real_mul.to_affine(), C::from_xy(x, y).unwrap());
+                }
+            }
```
daira, line 530: Check multiplication by 0, and multiplication of a base 𝒪. (I don't think these can occur in the Orchard circuit, but they should either work or have well-defined error behaviour.)

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +378,168 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+
+            Ok(())
+        }
+    }
+
+    #[test]
+    fn ecc() {
+        let k = 11;
```
daira, line 538: Is this the smallest value that will work (to speed up the test)?

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,643 @@
...
+    #[allow(clippy::many_single_char_names)]
+    #[allow(clippy::too_many_arguments)]
+    pub fn configure(
+        meta: &mut ConstraintSystem<C::Base>,
+        bits: Column<Advice>,
+        P: (Column<Advice>, Column<Advice>),
+        lambda: (Column<Advice>, Column<Advice>),
+        extras: [Column<Advice>; 5],
+    ) -> <Self as Chip<C::Base>>::Config {
+        let q_double = meta.selector();
```
daira, line 175: We don't use doubling in the Orchard circuit; how much cost are we paying for this selector?

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -190,14 +189,12 @@ pub(super) fn assign_region<C: CurveAffine>(
             .map(|(((x_p, x_q), y_p), y_q)| -> Result<(), Error> {
                 if x_q == x_p {
                     // x_q = x_p ⟹ A
-                    region.assign_advice(|| "set A", a, offset, || Ok(C::Base::one()))?;
+                    region.assign_advice(|| "set A", a, offset, || Ok(F::one()))?;
 
                     // Doubling case, λ = (y_q − y_p) / (x_q − x_p)
```
daira, line 194: This comment is not correct; give the equation for the tangent slope.

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -1,8 +1,7 @@
 use super::{util, CellValue, EccConfig, EccPoint};
 
-use group::Curve;
 use halo2::{
-    arithmetic::{CurveAffine, CurveExt, Field, FieldExt},
+    arithmetic::FieldExt,
     circuit::Region,
     plonk::{ConstraintSystem, Error, Expression},
 };
```
daira, line 7: On line 75, the label for the gate does not match the comment.

daira: Also the labels on lines 85 and 92 are a bit confusing; I suggest replacing those with the corresponding comments.

daira: Also I think the gate on line 116 may be incorrect.

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -190,14 +189,12 @@ pub(super) fn assign_region<C: CurveAffine>(
...
-                    region.assign_advice(|| "set A", a, offset, || Ok(C::Base::one()))?;
+                    region.assign_advice(|| "set A", a, offset, || Ok(F::one()))?;
 
                     // Doubling case, λ = (y_q − y_p) / (x_q − x_p)
-                    if y_p != C::Base::zero() {
-                        let lambda_val = C::Base::from_u64(3)
-                            * x_p
-                            * x_p
-                            * (C::Base::from_u64(2) * y_p).invert().unwrap();
+                    if y_p != F::zero() {
```
daira, line 195: y_p = 0 should not occur unless we are adding 𝒪+𝒪. It appears we don't assign lambda at all in that case; will this cause a problem?

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -210,17 +207,17 @@ pub(super) fn assign_region<C: CurveAffine>(
                     region.assign_advice(|| "set lambda", lambda, offset, || Ok(lambda_val))?;
                 }
 
-                if x_p == C::Base::zero() {
+                if x_p == F::zero() {
```
daira, line 210: Similarly we don't assign lambda in these cases. Will this cause a problem?

src/circuit/gadget/ecc/chip/add.rs
therealyingtong: Changed to `λ = 3(x_p)^2 / (2 * y_p)`

src/circuit/gadget/ecc.rs
therealyingtong: I've modified the test to expect that these cases produce errors.

src/circuit/gadget/ecc.rs
therealyingtong: Apparently `k = 9` is the smallest value.

src/circuit/gadget/ecc.rs
therealyingtong: `[0]B` returns `(0,0)` since variable-base scalar mul uses complete addition for the final bits. `[a]𝒪` returns an error since it uses incomplete addition for the initial bits.

src/circuit/gadget/ecc.rs
daira: This will change in any case now that we've optimized out A, B, C, and D.

src/circuit/gadget/ecc.rs
daira: That sounds fine. The latter should be documented.

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +378,332 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+                )?;
+
+                // [a]B
+                let mul_fixed = nullifier_k.mul(layouter.namespace(|| "mul"), &scalar_fixed)?;
+                if let (Some(x), Some(y)) = (mul_fixed.inner.x.value, mul_fixed.inner.y.value) {
+                    assert_eq!(real_mul_fixed.to_affine(), C::from_xy(x, y).unwrap());
+                }
+
+                // [0]B should return an error since fixed-base scalar multiplication
+                // uses incomplete addition internally.
```
daira, line 558: This may be a problem, because we require at least the signed 64-bit fixed-base scalar multiplication to work correctly for v = 0.

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +378,332 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+                    )?;
+                    let mul_fixed_short = value_commit_v
+                        .mul(layouter.namespace(|| "mul fixed"), &scalar_fixed_short)?;
+
+                    checks.push((real_mul_fixed_short, mul_fixed_short));
+                }
+
+                // [2^64 - 1]B
+                {
+                    let scalar_fixed_short = C::Scalar::from_u64(1 << 64 - 1);
```
daira, line 617:
```suggestion
                    let scalar_fixed_short = C::Scalar::from_u64(0xFFFFFFFFFFFFFFFFu64);
```

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +378,332 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+                    )?;
+                    let mul_fixed_short = value_commit_v
+                        .mul(layouter.namespace(|| "mul fixed"), &scalar_fixed_short)?;
+
+                    checks.push((real_mul_fixed_short, mul_fixed_short));
+                }
+
+                // [-(2^64 - 1)]B
+                {
+                    let scalar_fixed_short = -C::Scalar::from_u64(1 << 64 - 1);
```
daira, line 633:
```suggestion
                    let scalar_fixed_short = -C::Scalar::from_u64(0xFFFF_FFFF_FFFF_FFFFu64);
```

src/circuit/gadget/ecc/chip/double.rs
```diff
@@ -0,0 +1,103 @@
...
+    q_double: Expression<F>,
+    x_a: Expression<F>,
+    y_a: Expression<F>,
+    x_p: Expression<F>,
+    y_p: Expression<F>,
+) {
+    let x_p_2 = x_p.clone() * x_p.clone();
+    let x_p_4 = x_p_2.clone() * x_p_2.clone();
+
+    // 4⋅(y_p)^2⋅(x_a + 2⋅x_p) − 9⋅(x_p)^4 = 0
```
daira, line 20: Where does this equation come from?

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,550 @@
...
+        EccPoint {
+            x: x.0,
+            y: CellValue::<C::Base>::new(y_a_cell, *y_a),
+        }
+    };
+
+    let mut z_val = z.value;
+    // Complete addition
+    for (iter, k) in k_complete.iter().enumerate() {
+        // Each iteration uses 4 rows (two complete additions)
```
daira, line 228: I believe we can now do a complete addition in one row.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,550 @@
...
+        Y(y_a),
+        ZValue(CellValue::<C::Base>::new(z_cell, z_val)),
+    ))
+}
+
+fn decompose_scalar<C: CurveAffine>(scalar: C::Base) -> Vec<bool> {
+    // Cast into scalar field
+    let scalar = C::Scalar::from_bytes(&scalar.to_bytes()).unwrap();
+
+    // The scalar field `F_q = 2^254 + t_q`
```
daira, line 532:
```suggestion
    // The scalar field `F_q = 2^254 + t_q`.
    // -((2^127)^2) = -(2^254) = t_q (mod q)
```

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,237 @@
...
+                config.lagrange_coeffs[k],
+                w + offset,
+                || Ok(base.lagrange_coeffs.0[w].0[k]),
+            )?;
+        }
+
+        // Assign z-values for each window
+        region.assign_fixed(
+            || format!("z-value for window: {:?}", w),
+            config.fixed_z.into(),
```
daira, line 83:
```suggestion
            config.fixed_z,
```

src/circuit/gadget/ecc/chip/mul_fixed_short.rs
```diff
@@ -0,0 +1,259 @@
...
+                config.lagrange_coeffs[k],
+                w + offset,
+                || Ok(base.lagrange_coeffs_short.0[w].0[k]),
+            )?;
+        }
+
+        // Assign z-values for each window
+        region.assign_fixed(
+            || format!("z-value for window: {:?}", w),
+            config.fixed_z.into(),
```
daira, line 70:
```suggestion
            config.fixed_z,
```

src/circuit/gadget/ecc/chip/double.rs
therealyingtong: It comes from substituting `lambda` in the point doubling formulae. I've added a commit documenting this in the ECC book PR: https://github.com/zcash/orchard/pull/54/commits/58e3da88c274fa0261d61d550dfec224e74311b4

src/circuit/gadget/ecc.rs
therealyingtong: Fixed-base scalar multiplication now uses complete addition on the last step, so it returns `(0,0)` for `[0]B`.

src/circuit/gadget/ecc/chip/add.rs
therealyingtong: Resolving this comment since `add.rs` has undergone significant fixes.

src/circuit/gadget/ecc/chip/add.rs
therealyingtong: Resolving this comment since `add.rs` has undergone significant fixes.

src/circuit/gadget/ecc/chip/add.rs
therealyingtong: Resolving this comment since `add.rs` has undergone significant fixes.

src/circuit/gadget/ecc/chip/mul.rs
therealyingtong: This has now been implemented.

src/circuit/gadget/ecc/chip.rs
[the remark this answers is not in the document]
therealyingtong: I think `Box`-ing the array might be more memory-efficient (using heap allocation instead of stack allocation). I encountered a stack overflow when I tried removing the `Box`es.

src/circuit/gadget/ecc/chip.rs
therealyingtong: The doubling API, module, and selector have now been flagged off with `#[cfg(test)]`.

src/circuit/gadget/ecc.rs
therealyingtong: I've defined a separate `Config` for each operation (e.g. `add::Config, mul_fixed::Config`), each with an `impl From<EccConfig>` showing how it is derived from the top-level config.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,473 @@
...
+                meta.fixed_column(),
+                meta.fixed_column(),
+            ],
+            fixed_z: meta.fixed_column(),
+            #[cfg(test)]
+            q_double: meta.selector(),
+            q_add_incomplete: meta.selector(),
+            q_add: meta.selector(),
+            q_mul_hi: meta.selector(),
+            q_mul_lo: meta.selector(),
```
ebfull, line 210: Let's explore whether or not the Rust compiler guarantees that it will execute these operations in a particular and deterministic order. These calls happen within a single struct initialization expression, whereas in other code like this we usually separate them out into separate expressions which are guaranteed to run in a deterministic order by the language. If, for example, a new version of the Rust compiler is allowed to start initializing deeply-nested expressions before calling others, it could change the order of this and still compile successfully, breaking the mapping between the columns and thus failing to correctly verify and create proofs created by other users.

(I don't know if this guarantee is made by the compiler right now, if it is then we're good.)

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,473 @@
...
+}
+
+/// Configuration for the ECC chip
+#[derive(Clone, Debug)]
+#[allow(non_snake_case)]
+pub struct EccConfig {
+    /// Advice column for scalar decomposition into bits
+    pub bits: Column<Advice>,
+    /// Holds a point (x_p, y_p)
+    pub P: (Column<Advice>, Column<Advice>),
```
ebfull, line 84: We also use these columns (or at least the first one) to store scalars right?

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,473 @@
...
+    ) -> Result<Self::ScalarVar, Error> {
+        layouter.assign_region(
+            || "Witness scalar var",
+            |mut region| {
+                let cell = region.assign_advice(
+                    || "Scalar var",
+                    self.config().P.0,
+                    0,
+                    || value.ok_or(Error::SynthesisError),
+                )?;
```
ebfull, line 339: I'm still confused about this.

We're witnessing a scalar of Pallas (an element of Fq) in our circuit which is built over Fp (the native circuit field and the scalar field for Vesta, the curve we're using for halo in orchard). But q < p! So some scalars (Fq) cannot fit inside of a single cell in Fp. What am I missing?

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,385 @@
...
+
+                let delta = if x_q == x_p {
+                    if y_q != -y_p {
+                        (y_q + y_p).invert().unwrap()
+                    } else {
+                        C::Base::zero()
+                    }
+                } else {
+                    C::Base::zero()
+                };
```
ebfull, line 292: My understanding is that `delta` should be equal to `inv0(y_q + y_p)` when `x_q == x_p` regardless of whether or not `y_q != -y_p`. The code as written is the same behavior (it sorta checks if they're additive inverses of each other which means their sum is zero) but to avoid confusion let's be explicit.

Let's actually bring out `inv0()` as a function elsewhere in the code (this same module is fine) that we call from all the spots we need to use it.

ebfull (changes requested):
I got to `add_incomplete` in this review.

src/circuit/gadget/ecc/chip.rs
daira: This scalar multiplication only needs to work for scalars in [0, p).

src/circuit/gadget/ecc.rs
```diff
@@ -111,6 +113,7 @@ pub trait EccInstructions<C: CurveAffine>: Chip<C::Base> {
         b: &Self::Point,
     ) -> Result<Self::Point, Error>;
 
+    #[cfg(test)]
```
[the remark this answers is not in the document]
therealyingtong: I've removed point doubling from this PR altogether, and opened an issue to potentially reintroduce it in future. https://github.com/zcash/orchard/issues/90

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,473 @@
...
+    loaded: EccLoaded<C>,
+}
+
+impl<C: CurveAffine> PartialEq for EccChip<C> {
+    fn eq(&self, other: &Self) -> bool {
+        self.id == other.id
+    }
+}
+
+impl<C: CurveAffine> Eq for EccChip<C> {}
```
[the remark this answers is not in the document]
therealyingtong: Yes! This depends on an upstream change to derive `Eq, PartialEq` for `Selector`, `Permutation`: https://github.com/zcash/halo2/pull/279

src/circuit/gadget/ecc/chip.rs
therealyingtong: I couldn't find official documentation on whether this is guaranteed now / in future. There's a [StackOverflow thread](https://stackoverflow.com/questions/61984925/is-the-order-in-which-struct-fields-are-initialized-guaranteed-in-rust) which suggests that the order of field initialisation follows the order in which they're written in the initialiser. But failing any official documentation, I'm just going to separate this nested initialiser into single expressions.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,363 @@
...
+            scalar: ecc_config.extras[0],
+            perm: ecc_config.perm.clone(),
+            add_config: ecc_config.into(),
+            hi_config: incomplete::Config::hi_config(ecc_config),
+            lo_config: incomplete::Config::lo_config(ecc_config),
+        }
+    }
+}
+
+trait Mul<C: CurveAffine> {
```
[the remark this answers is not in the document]
therealyingtong: These functions are now floating free :+1:

src/circuit/gadget/ecc/chip.rs
therealyingtong: The doubling API has now been removed. I've opened a ticket to reintroduce it at some later time (see #90)

src/circuit/gadget/ecc/chip/double.rs
therealyingtong: Resolving this conversation since `double.rs` has been removed.

src/circuit/gadget/ecc/chip.rs
therealyingtong: I'm actually inclined to pass in `advices: [Column<Advice>; 10]` instead of naming any columns or "extras" here. The `bits`, `P`, and `lambda` columns are renamed in some operations anyway.

src/circuit/gadget/ecc.rs
```diff
@@ -111,14 +113,8 @@ pub trait EccInstructions<C: CurveAffine>: Chip<C::Base> {
...
 
-    /// Performs point doubling, returning `[2] a`.
-    fn double(
-        &self,
-        layouter: &mut impl Layouter<C::Base>,
-        a: &Self::Point,
-    ) -> Result<Self::Point, Error>;
-
     /// Performs variable-base scalar multiplication, returning `[scalar] base`.
+    /// Multiplication of the identity [a]𝒪 returns an error.
```
daira, line 117:
```suggestion
    /// Multiplication of the identity [a] 𝒪 returns an error.
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,370 @@
...
+            },
+        )?;
+
+        // Assign β = inv0(x_p)
+        region.assign_advice(
+            || "β",
+            self.beta,
+            offset,
+            || {
+                let beta = x_p.map(|x_p| inv0(x_p));
```
daira, line 237: Take clippy's suggestions.

src/circuit/gadget/ecc/chip.rs
therealyingtong: I've made it so that the top-level `EccConfig` only knows about 10 generic advice columns. These columns are renamed by individual components (e.g. `AddConfig`, `MulConfig`). The mapping for each individual config is captured in its `impl From<EccConfig>`.

daira (commented):
Partial review; flushing comments.

src/circuit/gadget/ecc.rs
```diff
@@ -376,3 +372,360 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug + Eq> FixedPoin
...
+                    chip.clone(),
+                    layouter.namespace(|| "point"),
+                    Some(endo_p.to_affine()),
+                )?;
+                let added_complete = p.add(layouter.namespace(|| "P + endo(P)"), &endo_p)?;
+                checks.push((real_added, added_complete));
+
+                for check in checks.into_iter() {
+                    let real_added = check.0;
+                    let added_complete = check.1;
```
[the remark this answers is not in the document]
therealyingtong: I've inlined equality checks into the `assign_region()`s of the relevant operations, flagged off with `[cfg(test)]`s.

src/circuit/gadget/ecc.rs
```diff
@@ -114,7 +114,7 @@ pub trait EccInstructions<C: CurveAffine>: Chip<C::Base> {
     ) -> Result<Self::Point, Error>;
 
     /// Performs variable-base scalar multiplication, returning `[scalar] base`.
-    /// Multiplication of the identity [a]𝒪 returns an error.
+    /// Multiplication of the identity [a] 𝒪 returns an error.
```
daira, line 117: Hmm, this didn't fix the doc check failure. What command is being run for doc checks? None of `cargo doc`, `cargo check`, or `cargo clippy` reproduce this for me.

therealyingtong: Oh, it runs `cargo doc --document-private-items`.

daira:
```suggestion
    /// Multiplication of the identity `[a] 𝒪` returns an error.
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,423 @@
...
+}
+
+/// A full-width scalar used for variable-base scalar multiplication.
+/// This is decomposed in chunks of `window_width` bits in little-endian order.
+/// For example, if `window_width` = 3, we will have [k_0, k_1, ..., k_n]
+/// where `scalar = k_0 + k_1 * (2^3) + ... + k_n * (2^3)^n`.
+#[derive(Clone, Debug)]
+pub struct EccScalarFixed<C: CurveAffine> {
+    value: Option<C::Scalar>,
+    k_bits: Vec<CellValue<C::Base>>,
```
daira, line 264:
```suggestion
    k_limbs: Vec<CellValue<C::Base>>,
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,423 @@
...
+
+/// A signed short scalar used for variable-base scalar multiplication.
+/// This is decomposed in chunks of `window_width` bits in little-endian order.
+/// For example, if `window_width` = 3, we will have [k_0, k_1, ..., k_n]
+/// where `scalar = k_0 + k_1 * (2^3) + ... + k_n * (2^3)^n`.
+#[derive(Clone, Debug)]
+pub struct EccScalarFixedShort<C: CurveAffine> {
+    magnitude: Option<C::Scalar>,
+    sign: CellValue<C::Base>,
+    k_bits: Vec<CellValue<C::Base>>,
```
daira, line 275:
```suggestion
    k_limbs: Vec<CellValue<C::Base>>,
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,423 @@
...
+            commit_ivk_r,
+            note_commit_r,
+            nullifier_k,
+            value_commit_r,
+            value_commit_v,
+        }
+    }
+}
+
+/// A full-width scalar used for variable-base scalar multiplication.
```
daira, line 257:
```suggestion
/// A full-width scalar used for fixed-base scalar multiplication.
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,423 @@
...
+/// This is decomposed in chunks of `window_width` bits in little-endian order.
+/// For example, if `window_width` = 3, we will have [k_0, k_1, ..., k_n]
+/// where `scalar = k_0 + k_1 * (2^3) + ... + k_n * (2^3)^n`.
+#[derive(Clone, Debug)]
+pub struct EccScalarFixed<C: CurveAffine> {
+    value: Option<C::Scalar>,
+    k_bits: Vec<CellValue<C::Base>>,
+}
+
+/// A signed short scalar used for variable-base scalar multiplication.
```
daira, line 267:
```suggestion
/// A signed short scalar used for fixed-base scalar multiplication.
```

src/circuit/gadget/ecc/chip/witness_scalar_fixed.rs
```diff
@@ -0,0 +1,133 @@
...
+        let windows: Option<Vec<u8>> = value.map(|value| {
+            util::decompose_scalar_fixed::<C>(
+                value,
+                Self::SCALAR_NUM_BITS,
+                constants::FIXED_BASE_WINDOW_SIZE,
+            )
+        });
+
+        // Store the scalar decomposition
+        let mut k_bits: Vec<CellValue<C::Base>> = Vec::new();
```
daira, line 110:
```suggestion
        let mut k_digits: Vec<CellValue<C::Base>> = Vec::new();
```

src/circuit/gadget/ecc/chip/witness_scalar_fixed.rs
```diff
@@ -0,0 +1,133 @@
...
+            let k_var = region.assign_advice(
+                || format!("k[{:?}]", offset + idx),
+                self.k(),
+                offset + idx,
+                || window.ok_or(Error::SynthesisError),
+            )?;
+            k_bits.push(CellValue::new(k_var, window));
+        }
+
+        Ok(k_bits)
```
daira, line 131:
```suggestion
            k_digits.push(CellValue::new(k_var, window));
        }

        Ok(k_digits)
```

src/circuit/gadget/ecc/chip/witness_scalar_fixed/full_width.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    fn assign_region(
+        &self,
+        value: Option<C::Scalar>,
+        offset: usize,
+        region: &mut Region<'_, C::Base>,
+    ) -> Result<EccScalarFixed<C>, Error> {
+        let k_bits = self.decompose_scalar_fixed(value, offset, region)?;
+
+        Ok(EccScalarFixed { value, k_bits })
```
daira, line 47:
```suggestion
        let k_digits = self.decompose_scalar_fixed(value, offset, region)?;

        Ok(EccScalarFixed { value, k_digits })
```

src/circuit/gadget/ecc/chip/witness_scalar_fixed/short.rs
```diff
@@ -0,0 +1,106 @@
...
+                -C::Scalar::one()
+            } else {
+                C::Scalar::one()
+            }
+        });
+
+        let magnitude = sign.zip(value).map(|(sign, value)| sign * value);
+
+        // Decompose magnitude into `k`-bit windows
+        let k_bits = self.decompose_scalar_fixed(magnitude, offset, region)?;
```
daira, line 82:
```suggestion
        let k_digits = self.decompose_scalar_fixed(magnitude, offset, region)?;
```

src/circuit/gadget/ecc/chip/witness_scalar_fixed/short.rs
```diff
@@ -0,0 +1,106 @@
...
+            if sign == C::Scalar::one() {
+                C::Base::one()
+            } else {
+                -C::Base::one()
+            }
+        });
+        let sign_cell = region.assign_advice(
+            || "sign",
+            self.k_s,
+            offset + k_bits.len(),
```
daira, line 96:
```suggestion
            offset + k_digits.len(),
```

src/circuit/gadget/ecc/chip/witness_scalar_fixed/short.rs
```diff
@@ -0,0 +1,106 @@
...
+            || sign.ok_or(Error::SynthesisError),
+        )?;
+
+        Ok(EccScalarFixedShort {
+            magnitude,
+            sign: CellValue::<C::Base>::new(sign_cell, sign),
+            k_bits,
```
daira, line 103:
```suggestion
            k_digits,
```

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,446 @@
...
+        // Initialize the accumulator `acc = [2]base`
+        let acc = self
+            .add_config
+            .assign_region(&base, &base, offset, region)?;
+
+        // Increase the offset by 1 after complete addition.
+        let offset = offset + 1;
+
+        // Decompose the scalar bitwise (big-endian bit order).
+        let k_bits = decompose_for_scalar_mul::<C>(scalar.value);
```
daira, line 115:
```suggestion
        let k_digits = decompose_for_scalar_mul::<C>(scalar.value);
```

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,446 @@
...
+        let z_val = C::Base::zero();
+        let z_cell =
+            region.assign_advice(|| "initial z", self.hi_config.z, offset, || Ok(z_val))?;
+        let z = CellValue::new(z_cell, Some(z_val));
+
+        // Increase the offset by 1 after initializing `z`.
+        let offset = offset + 1;
+
+        // Double-and-add (incomplete addition) for the `hi` half of the scalar decomposition
+        let k_incomplete_hi = &k_bits[incomplete_hi_range::<C>()];
```
daira, line 127:
```suggestion
        let k_incomplete_hi = &k_digits[incomplete_hi_range::<C>()];
```

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,446 @@
...
+        let (x, y_a, z) = self.hi_config.double_and_add(
+            region,
+            offset,
+            &base,
+            k_incomplete_hi,
+            (X(acc.x.clone()), Y(acc.y.value), Z(z)),
+        )?;
+
+        // Double-and-add (incomplete addition) for the `lo` half of the scalar decomposition
+        let k_incomplete_lo = &k_bits[incomplete_lo_range::<C>()];
```
daira, line 137:
```suggestion
        let k_incomplete_lo = &k_digits[incomplete_lo_range::<C>()];
```

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,446 @@
...
+
+        // Increase the offset by 1 after complete addition.
+        let offset = offset + 1;
+
+        // Complete addition
+        let (acc, z_val) = {
+            let complete_config: complete::Config<C> = self.into();
+            // Bits used in complete addition. k_{3} to k_{1} inclusive
+            // The LSB k_{0} is handled separately.
+            let k_complete = &k_bits[complete_range::<C>()];
```
daira, line 179:
```suggestion
            let k_complete = &k_digits[complete_range::<C>()];
```

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,446 @@
...
+            complete_config.assign_region(region, offset, k_complete, base, acc, z.value)?
+        };
+
+        let offset = offset + complete_len::<C>() * 2;
+
+        // Process the least significant bit
+        let k_0 = k_bits[C::Scalar::NUM_BITS as usize - 1];
```
daira, line 186:
```suggestion
        let k_0 = k_digits[C::Scalar::NUM_BITS as usize - 1];
```

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,503 @@
...
+    fn from(scalar_fixed: &EccScalarFixedShort<C>) -> Self {
+        Self::Short(scalar_fixed.clone())
+    }
+}
+
+impl<C: CurveAffine> ScalarFixed<C> {
+    fn k_bits(&self) -> &[CellValue<C::Base>] {
+        match self {
+            ScalarFixed::FullWidth(scalar) => &scalar.k_bits,
+            ScalarFixed::Short(scalar) => &scalar.k_bits,
```
daira, line 227:
```suggestion
    fn k_digits(&self) -> &[CellValue<C::Base>] {
        match self {
            ScalarFixed::FullWidth(scalar) => &scalar.k_digits,
            ScalarFixed::Short(scalar) => &scalar.k_digits,
```

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,503 @@
...
+        }
+    }
+
+    // The scalar decomposition was done in the base field. For computation
+    // outside the circuit, we now convert them back into the scalar field.
+    fn k_field(&self) -> Vec<Option<C::Scalar>> {
+        self.k_bits()
```
daira, line 234:
```suggestion
        self.k_digits()
```

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,503 @@
...
+                    .map(|value| C::Scalar::from_bytes(&value.to_bytes()).unwrap())
+            })
+            .collect::<Vec<_>>()
+    }
+
+    // The scalar decomposition is guaranteed to be in three-bit windows,
+    // so we also cast the least significant byte in their serialisation
+    // into usize for convenient indexing into `u`-values
+    fn k_usize(&self) -> Vec<Option<usize>> {
+        self.k_bits()
```
daira, line 247:
```suggestion
        self.k_digits()
```

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,503 @@
...
+    }
+
+    fn copy_scalar(
+        &self,
+        region: &mut Region<'_, C::Base>,
+        offset: usize,
+        scalar: &ScalarFixed<C>,
+    ) -> Result<(), Error> {
+        // Copy the scalar decomposition
+        for (window, k) in scalar.k_bits().iter().enumerate() {
```
daira, line 351:
```suggestion
        for (window, k) in scalar.k_digits().iter().enumerate() {
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,482 @@
...
+        advice_columns.insert(self.x_p);
+        advice_columns.insert(self.y_p);
+        advice_columns.insert(self.x_qr);
+        advice_columns.insert(self.y_qr);
+        advice_columns.insert(self.lambda);
+        advice_columns.insert(self.alpha);
+        advice_columns.insert(self.beta);
+        advice_columns.insert(self.gamma);
+        advice_columns.insert(self.delta);
+        advice_columns
```
daira, line 66:
```suggestion
        IntoIter::new([
            self.x_p,
            self.y_p,
            self.x_qr,
            self.y_qr,
            self.lambda,
            self.alpha,
            self.beta,
            self.gamma,
            self.delta,
        ])
        .collect()
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,482 @@
+use super::{util, CellValue, EccConfig, EccPoint};
+use ff::Field;
+use halo2::{
+    arithmetic::{CurveAffine, FieldExt},
+    circuit::Region,
+    plonk::{Advice, Column, ConstraintSystem, Error, Expression, Permutation, Selector},
+    poly::Rotation,
+};
+use std::collections::HashSet;
+
```
daira, line 10:
```suggestion
use core::array::IntoIter;

```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,482 @@
...
+        // α = inv0(x_q - x_p)
+        let alpha = meta.query_advice(self.alpha, Rotation::cur());
+        // β = inv0(x_p)
+        let beta = meta.query_advice(self.beta, Rotation::cur());
+        // γ = inv0(x_q)
+        let gamma = meta.query_advice(self.gamma, Rotation::cur());
+        // δ = inv0(y_p + y_q) if x_q = x_p, 0 otherwise
+        let delta = meta.query_advice(self.delta, Rotation::cur());
+
+        // Useful composite expressions
```
daira, line 88: I really like this pattern: each of these values are 1 if the if-condition holds, and unconstrained otherwise. We should probably give the pattern a name, although I can't think of one right now.

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,482 @@
...
+            );
+
+            // (1 - (x_q - x_p)⋅α)⋅(2y_p ⋅λ - 3x_p^2) = 0
+            meta.create_gate("(1 - (x_q - x_p)⋅α)⋅(2y_p ⋅λ - 3x_p^2) = 0", |_| {
+                let three_x_p_sq = three * x_p.clone() * x_p.clone(); // 3x_p^2
+                let two_y_p = two.clone() * y_p.clone(); // 2y_p
+                let derivative = two_y_p * lambda.clone() - three_x_p_sq; // (2y_p ⋅λ - 3x_p^2)
+
+                // q_add ⋅(1 - (x_q - x_p)⋅α)⋅(2y_p ⋅λ - 3x_p^2)
+                q_add.clone() * (one.clone() - if_alpha.clone()) * derivative
```
daira, line 126: 2y<sub>p</sub>⋅λ - 3x<sub>p</sub><sup>2</sup> is not the derivative itself; setting it to zero constrains the derivative λ.

```suggestion
                let derivative_equation = two_y_p * lambda.clone() - three_x_p_sq; // (2y_p⋅λ - 3x_p^2)

                // q_add ⋅ (1 - (x_q - x_p)⋅α)⋅(2y_p⋅λ - 3x_p^2)
                q_add.clone() * (one.clone() - if_alpha.clone()) * derivative_equation
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,482 @@
...
+            // x_p⋅x_q⋅(x_q - x_p)⋅(λ^2 - x_p - x_q - x_r) = 0
+            meta.create_gate(
+                "x_p⋅x_q⋅(x_q - x_p)⋅(λ^2 - x_p - x_q - x_r) = 0",
+                |_| {
+                    let x_q_minus_x_p = x_q.clone() - x_p.clone(); // (x_q - x_p)
+                    let incomplete =
+                        lambda.clone() * lambda.clone() - x_p.clone() - x_q.clone() - x_r.clone(); // (λ^2 - x_p - x_q - x_r)
+
+                    // q_add ⋅ x_p⋅x_q⋅(x_q - x_p)⋅(λ^2 - x_p - x_q - x_r)
+                    q_add.clone() * x_p.clone() * x_q.clone() * x_q_minus_x_p * incomplete
```
daira, line 138:
```suggestion
                    let incomplete_constraint =
                        lambda.clone() * lambda.clone() - x_p.clone() - x_q.clone() - x_r.clone(); // (λ^2 - x_p - x_q - x_r)

                    // q_add ⋅ x_p⋅x_q⋅(x_q - x_p)⋅(λ^2 - x_p - x_q - x_r)
                    q_add.clone() * x_p.clone() * x_q.clone() * x_q_minus_x_p * incomplete_constraint
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,482 @@
...
+                    q_add.clone()
+                        * x_p.clone()
+                        * x_q.clone()
+                        * x_q_minus_x_p
+                        * (lambda.clone() * x_p_minus_x_r - y_p.clone() - y_r.clone())
+                },
+            );
+
+            // x_p⋅x_q⋅(y_q + y_p)⋅(λ^2 - x_p - x_q - x_r) = 0
+            meta.create_gate("y_r check", |_| {
```
daira, line 159: This check doesn't involve y<sub>r</sub>.
```suggestion
            meta.create_gate("x_p⋅x_q⋅(y_q + y_p)⋅(λ^2 - x_p - x_q - x_r) = 0", |_| {
```
(will need formatting).

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,482 @@
...
+        let if_delta = (y_q.clone() + y_p.clone()) * delta;
+
+        // Useful constants
+        let one = Expression::Constant(F::one());
+        let two = Expression::Constant(F::from_u64(2));
+        let three = Expression::Constant(F::from_u64(3));
+
+        // Handle cases in incomplete addition
+        {
+            // (x_q − x_p)⋅((x_q − x_p)⋅λ − (y_q−y_p))=0
```
daira, line 105: Remove comments that are identical to gate labels.

src/circuit/gadget/ecc/chip.rs
therealyingtong: I've renamed `k_bits` -> `windows` for fixed-base scalar mul.

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,482 @@
...
+            );
+
+            // x_p⋅x_q⋅(y_q + y_p)⋅(λ^2 - x_p - x_q - x_r) = 0
+            meta.create_gate("y_r check", |_| {
+                let y_q_plus_y_p = y_q.clone() + y_p.clone(); // (y_q + y_p)
+                let incomplete =
+                    lambda.clone() * lambda.clone() - x_p.clone() - x_q.clone() - x_r.clone(); // (λ^2 - x_p - x_q - x_r)
+
+                // q_add ⋅ x_p⋅x_q⋅(y_q + y_p)⋅(λ^2 - x_p - x_q - x_r)
+                q_add.clone() * x_p.clone() * x_q.clone() * y_q_plus_y_p * incomplete
```
daira:
```suggestion
                let output_line_x =
                    lambda.clone() * lambda.clone() - x_p.clone() - x_q.clone() - x_r.clone(); // (λ^2 - x_p - x_q - x_r)

                // q_add ⋅ x_p⋅x_q⋅(y_q + y_p)⋅(λ^2 - x_p - x_q - x_r)
                q_add.clone() * x_p.clone() * x_q.clone() * y_q_plus_y_p * output_line_x
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,386 @@
...
+                    advices[0].into(),
+                    advices[1].into(),
+                    advices[2].into(),
+                    advices[3].into(),
+                    advices[4].into(),
+                    advices[6].into(),
+                    advices[7].into(),
+                    advices[8].into(),
+                    advices[9].into(),
+                ],
```
daira, line 169:
```suggestion
                advices
                    .iter()
                    .map(|a| (*a).into())
                    .collect::<ArrayVec<Column<Any>, 10>>()
                    .as_slice(),
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,386 @@
+use super::EccInstructions;
+use crate::constants::{self, OrchardFixedBasesFull, ValueCommitV};
+use arrayvec::ArrayVec;
+use ff::Field;
+use halo2::{
+    arithmetic::CurveAffine,
+    circuit::{Cell, Chip, Layouter},
+    plonk::{Advice, Column, ConstraintSystem, Error, Fixed, Permutation, Selector},
```
daira, line 8:
```suggestion
    plonk::{Advice, Any, Column, ConstraintSystem, Error, Fixed, Permutation, Selector},
```
(used by a suggestion below).

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,386 @@
...
+    type FixedPoints = OrchardFixedBasesFull<C>;
+    type FixedPointsShort = ValueCommitV<C>;
+
+    fn witness_scalar_var(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        value: Option<C::Base>,
+    ) -> Result<Self::ScalarVar, Error> {
+        layouter.assign_region(
+            || "Witness scalar var",
```
daira, line 267:
```suggestion
            || "witness scalar for variable-base mul",
```
(for consistency with below).

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,386 @@
...
+    }
+
+    fn witness_scalar_fixed_short(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        value: Option<C::Scalar>,
+    ) -> Result<Self::ScalarFixedShort, Error> {
+        let config: witness_scalar_fixed::short::Config<C> = self.config().into();
+        layouter.assign_region(
+            || "witness scalar for fixed-base mul",
```
daira, line 299:
```suggestion
            || "witness short scalar for fixed-base mul",
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    fn add_incomplete(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        a: &Self::Point,
+        b: &Self::Point,
+    ) -> Result<Self::Point, Error> {
+        let config: add_incomplete::Config = self.config().into();
+        layouter.assign_region(
+            || "point addition",
```
daira, line 328:
```suggestion
            || "incomplete point addition",
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    fn add(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        a: &Self::Point,
+        b: &Self::Point,
+    ) -> Result<Self::Point, Error> {
+        let config: add::Config = self.config().into();
+        layouter.assign_region(
+            || "point addition",
```
daira, line 341:
```suggestion
            || "complete point addition",
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,386 @@
...
+    fn mul_fixed(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        scalar: &Self::ScalarFixed,
+        base: &Self::FixedPoints,
+    ) -> Result<Self::Point, Error> {
+        let config: mul_fixed::full_width::Config<C, { constants::NUM_WINDOWS }> =
+            self.config().into();
+        layouter.assign_region(
+            || format!("Multiply {:?}", base),
```
daira, line 368:
```suggestion
            || format!("fixed-base mul of {:?}", base),
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -0,0 +1,386 @@
...
+    fn mul_fixed_short(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        scalar: &Self::ScalarFixedShort,
+        base: &Self::FixedPointsShort,
+    ) -> Result<Self::Point, Error> {
+        let config: mul_fixed::short::Config<C, { constants::NUM_WINDOWS_SHORT }> =
+            self.config().into();
+        layouter.assign_region(
+            || format!("Multiply {:?}", base),
```
daira, line 382:
```suggestion
            || format!("short fixed-base mul of {:?}", base),
```

src/circuit/gadget/ecc/chip/add.rs
daira: Let's call it the "conditional 1 pattern".

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+
+        // Useful constants
+        let one = Expression::Constant(F::one());
+        let two = Expression::Constant(F::from_u64(2));
+        let three = Expression::Constant(F::from_u64(3));
+
+        // Handle cases in incomplete addition
+        {
+            meta.create_gate(
+                "(x_q − x_p)⋅((x_q − x_p)⋅λ − (y_q−y_p))=0",
```
daira, line 107:
```suggestion
                "(x_q − x_p)⋅((x_q − x_p)⋅λ − (y_q−y_p)) = 0",
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+
+                    let y_q_minus_y_p = y_q.clone() - y_p.clone(); // (y_q − y_p)
+                    let incomplete = x_q_minus_x_p.clone() * lambda.clone() - y_q_minus_y_p; // (x_q − x_p)⋅λ − (y_q−y_p)
+
+                    // q_add ⋅(x_q − x_p)⋅((x_q − x_p)⋅λ − (y_q−y_p))
+                    q_add.clone() * x_q_minus_x_p * incomplete
+                },
+            );
+
+            meta.create_gate("(1 - (x_q - x_p)⋅α)⋅(2y_p ⋅λ - 3x_p^2) = 0", |_| {
```
daira, line 119:
```suggestion
            meta.create_gate("q_add ⋅ (1 - (x_q - x_p)⋅α)⋅(2y_p⋅λ - 3x_p^2) = 0", |_| {
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+                let three_x_p_sq = three * x_p.clone() * x_p.clone(); // 3x_p^2
+                let two_y_p = two.clone() * y_p.clone(); // 2y_p
+                let tangent_line = two_y_p * lambda.clone() - three_x_p_sq; // (2y_p ⋅λ - 3x_p^2)
+
+                // q_add ⋅(1 - (x_q - x_p)⋅α)⋅(2y_p ⋅λ - 3x_p^2)
+                q_add.clone() * (one.clone() - if_alpha.clone()) * tangent_line
+            });
+
+            meta.create_gate(
+                "x_p⋅x_q⋅(x_q - x_p)⋅(λ^2 - x_p - x_q - x_r) = 0",
```
daira, line 129:
```suggestion
                "q_add ⋅ x_p⋅x_q⋅(x_q - x_p)⋅(λ^2 - x_p - x_q - x_r) = 0",
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+                    let secant_line =
+                        lambda.clone() * lambda.clone() - x_p.clone() - x_q.clone() - x_r.clone(); // (λ^2 - x_p - x_q - x_r)
+
+                    // q_add ⋅ x_p⋅x_q⋅(x_q - x_p)⋅(λ^2 - x_p - x_q - x_r)
+                    q_add.clone() * x_p.clone() * x_q.clone() * x_q_minus_x_p * secant_line
+                },
+            );
+
+            meta.create_gate(
+                "x_p⋅x_q⋅(x_q - x_p)⋅(λ ⋅(x_p - x_r) - y_p - y_r) = 0",
```
daira, line 141:
```suggestion
                "q_add ⋅ x_p ⋅ x_q ⋅ (x_q - x_p) ⋅ (λ⋅(x_p - x_r) - y_p - y_r) = 0",
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+                    q_add.clone()
+                        * x_p.clone()
+                        * x_q.clone()
+                        * x_q_minus_x_p
+                        * (lambda.clone() * x_p_minus_x_r - y_p.clone() - y_r.clone())
+                },
+            );
+
+            meta.create_gate(
+                "x_p⋅x_q⋅(y_q + y_p)⋅(λ^2 - x_p - x_q - x_r) = 0",
```
daira, line 156:
```suggestion
                "q_add ⋅ x_p ⋅ x_q ⋅ (y_q + y_p) ⋅ (λ^2 - x_p - x_q - x_r) = 0",
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+                    let incomplete =
+                        lambda.clone() * lambda.clone() - x_p.clone() - x_q.clone() - x_r.clone(); // (λ^2 - x_p - x_q - x_r)
+
+                    // q_add ⋅ x_p⋅x_q⋅(y_q + y_p)⋅(λ^2 - x_p - x_q - x_r)
+                    q_add.clone() * x_p.clone() * x_q.clone() * y_q_plus_y_p * incomplete
+                },
+            );
+
+            meta.create_gate(
+                "x_p⋅x_q⋅(y_q + y_p)⋅(λ ⋅(x_p - x_r) - y_p - y_r) = 0",
```
daira, line 168:
```suggestion
                "q_add ⋅ x_p ⋅ x_q ⋅ (y_q + y_p) ⋅ (λ⋅(x_p - x_r) - y_p - y_r) = 0",
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+                    // q_add ⋅ x_p⋅x_q⋅(y_q + y_p)⋅(λ ⋅(x_p - x_r) - y_p - y_r)
+                    q_add.clone()
+                        * x_p.clone()
+                        * x_q.clone()
+                        * y_q_plus_y_p
+                        * (lambda.clone() * x_p_minus_x_r - y_p.clone() - y_r.clone())
+                },
+            );
+
+            meta.create_gate("(1 - x_p * β) * (x_r - x_q) = 0", |_| {
```
daira, line 182:
```suggestion
            meta.create_gate("q_add ⋅ (1 - x_p⋅β) ⋅ (x_r - x_q) = 0", |_| {
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+                q_add.clone() * (one.clone() - if_beta.clone()) * (x_r.clone() - x_q.clone())
+            });
+
+            meta.create_gate("(1 - x_p * β) * (y_r - y_q) = 0", |_| {
```
daira, line 186:
```suggestion
            meta.create_gate("q_add ⋅ (1 - x_p⋅β) ⋅ (y_r - y_q) = 0", |_| {
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+                q_add.clone() * (one.clone() - if_beta) * (y_r.clone() - y_q.clone())
+            });
+
+            meta.create_gate("(1 - x_q * γ) * (x_r - x_p) = 0", |_| {
```
daira, line 190:
```suggestion
            meta.create_gate("q_add ⋅ (1 - x_q⋅γ) ⋅ (x_r - x_p) = 0", |_| {
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+                q_add.clone() * (one.clone() - if_gamma.clone()) * (x_r.clone() - x_p.clone())
+            });
+
+            meta.create_gate("(1 - x_q * γ) * (y_r - y_p) = 0", |_| {
```
daira, line 194:
```suggestion
            meta.create_gate("q_add ⋅ (1 - x_q⋅γ) ⋅ (y_r - y_p) = 0", |_| {
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+                q_add.clone() * (one.clone() - if_gamma) * (y_r.clone() - y_p.clone())
+            });
+
+            meta.create_gate("((1 - (x_q - x_p) * α - (y_q + y_p) * δ)) * x_r", |_| {
```
daira, line 198:
```suggestion
            meta.create_gate("q_add ⋅ ((1 - (x_q - x_p)⋅α - (y_q + y_p)⋅δ)) ⋅ x_r", |_| {
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,478 @@
...
+                q_add.clone() * (one.clone() - if_alpha.clone() - if_delta.clone()) * x_r.clone()
+            });
+
+            meta.create_gate("((1 - (x_q - x_p) * α - (y_q + y_p) * δ)) * y_r", |_| {
```
daira, line 202:
```suggestion
            meta.create_gate("q_add ⋅ ((1 - (x_q - x_p)⋅α - (y_q + y_p)⋅δ)) ⋅ y_r", |_| {
```

src/circuit/gadget/ecc/chip/add_incomplete.rs
```diff
@@ -0,0 +1,197 @@
...
+        let q_add_incomplete = meta.query_selector(self.q_add_incomplete, Rotation::cur());
+        let x_p = meta.query_advice(self.x_p, Rotation::cur());
+        let y_p = meta.query_advice(self.y_p, Rotation::cur());
+        let x_q = meta.query_advice(self.x_qr, Rotation::cur());
+        let y_q = meta.query_advice(self.y_qr, Rotation::cur());
+        let x_r = meta.query_advice(self.x_qr, Rotation::next());
+        let y_r = meta.query_advice(self.y_qr, Rotation::next());
+
+        // (x_r + x_q + x_p)⋅(x_p − x_q)^2 − (y_p − y_q)^2 = 0
+        meta.create_gate("point addition expr1", |_| {
```
daira, line 50:
```suggestion
        // (x_r + x_q + x_p)⋅(x_p − x_q)^2 − (y_p − y_q)^2 = 0
        meta.create_gate("q_add_incomplete ⋅ (x_r + x_q + x_p) ⋅ (x_p - x_q)^2 - (y_p - y_q)^2 = 0", |_| {
```

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,469 @@
...
+            // ((1 - (x_q - x_p) * α - (y_q + y_p) * δ)) * y_r
+            let poly12 = (one - if_alpha - if_delta) * y_r;
+
+            [
+                poly1, poly2, poly3, poly4, poly5, poly6, poly7, poly8, poly9, poly10, poly11,
+                poly12,
+            ]
+            .iter()
+            .map(|poly| q_add.clone() * poly.clone())
+            .collect()
```
daira, line 194: Hmm. With the refactoring to allow gates to specify multiple polynomials, there is a loss of debug information that I hadn't anticipated: we only have a label for the overall gate, not for each polynomial.

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,469 @@
...
+                let y_q = y_q.ok_or(Error::SynthesisError)?;
+
+                let delta = if x_q == x_p {
+                    inv0(y_q + y_p)
+                } else {
+                    C::Base::zero()
+                };
+                Ok(delta)
+            },
+        )?;
```
daira, line 272: Consider using batch inversion. That would reduce the cost of these inversions in witness generation by roughly a factor of 4, even if we don't batch them across gate instances.

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -0,0 +1,469 @@
...
+        #[allow(clippy::collapsible_else_if)]
+        // Assign lambda
+        let lambda = x_p
+            .zip(y_p)
+            .zip(x_q)
+            .zip(y_q)
+            .map(|(((x_p, y_p), x_q), y_q)| {
+                if x_q != x_p {
+                    // λ = (y_q - y_p)/(x_q - x_p)
+                    (y_q - y_p) * (x_q - x_p).invert().unwrap()
```
daira, line 283: `inv0(x_q - x_p)`, which we compute above for `alpha`, would suffice here since we know that `x_q != x_p` in this branch.

src/circuit/gadget/ecc/chip/add.rs
str4d: I'll see what I can do to improve this. I don't think we necessarily need to force a label on each polynomial since in some cases it may not actually help, but it would be nice to enable it if possible. In any case, the debug tooling _can_ reference a polynomial index within a gate; I forgot to add that, but will do so now.

str4d: Added constrain indices to error-reporting: https://github.com/zcash/halo2/pull/285

str4d: Optional static annotations for constraints: https://github.com/zcash/halo2/pull/286

src/circuit/gadget/ecc/chip.rs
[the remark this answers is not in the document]
therealyingtong: Resolving now that the upstream PR has been merged.

src/circuit/gadget/ecc/chip.rs
therealyingtong: I realised the ECC chip `configure()` should take in a `perm: Permutation`, now that we're defining a single top-level permutation.

src/circuit/gadget/ecc/chip/add.rs
therealyingtong: Resolving this (and similar suggested changes below) because we are no longer creating a single gate for each polynomial constraint, as of https://github.com/zcash/halo2/pull/273.

therealyingtong:
I am closing this PR in favour of https://github.com/zcash/orchard/pull/107.

This is because the changes to `constants.rs` and the `constants` module have been factored out into #106. I will link the unresolved comments from this PR on the corresponding positions in the #107.

src/circuit/gadget/ecc/chip/add.rs
therealyingtong: Resolving this since the MockProver improvements to address this have been merged upstream (https://github.com/zcash/halo2/pull/285, https://github.com/zcash/halo2/pull/286)

PR #43, opened 2021-03-12
Add Orchard constants

therealyingtong:
Closes #42

PR #75, opened 2021-05-04
Modify ECC gadget to work with chip refactor

therealyingtong:
Based on #43 (not yet merged)

### DIFF
--- a/src/circuit/gadget/ecc.rs
+++ b/src/circuit/gadget/ecc.rs
@@ -88,6 +88,8 @@ pub trait EccInstructions<C: CurveAffine>: Chip<C::Base> {
     ) -> Result<Self::FixedPointShort, Error>;
 
     /// Performs incomplete point addition, returning `a + b`.
+    ///
+    /// This returns an error in exceptional cases.
     fn add_incomplete(
         &self,
         layouter: &mut impl Layouter<C::Base>,
--- a/src/circuit/gadget/ecc.rs
+++ b/src/circuit/gadget/ecc.rs
@@ -34,6 +34,9 @@ pub trait EccInstructions<C: CurveAffine>: Chip<C::Base> {
     type FixedPoints: Clone + Debug;
     /// Variable representing a fixed elliptic curve point (constant in the circuit).
     type FixedPoint: Clone + Debug;
+    /// Variable representing a fixed elliptic curve point (constant in the circuit)
+    /// to be used in scalar multiplication with a short signed exponent.
+    type FixedPointShort: Clone + Debug;
 
     /// Witnesses the given base field element as a private input to the circuit for variable-base scalar mul.
     fn witness_scalar_var(
@@ -70,6 +73,13 @@ pub trait EccInstructions<C: CurveAffine>: Chip<C::Base> {
     /// The pre-loaded cells are used to set up equality constraints in other
     /// parts of the circuit where the fixed base is used.
     fn get_fixed(&self, fixed_points: Self::FixedPoints) -> Result<Self::FixedPoint, Error>;
+
+    /// Returns a fixed point to be used in scalar multiplication with a signed
+    /// short exponent.
+    fn get_fixed_short(
+        &self,
+        fixed_points: Self::FixedPoints,
+    ) -> Result<Self::FixedPointShort, Error>;
 
     /// Performs incomplete point addition, returning `a + b`.
     fn add_incomplete(
@@ -108,7 +118,7 @@ pub trait EccInstructions<C: CurveAffine>: Chip<C::Base> {
         &self,
         layouter: &mut impl Layouter<C::Base>,
         scalar: &Self::ScalarFixedShort,
-        base: &Self::FixedPoint,
+        base: &Self::FixedPointShort,
     ) -> Result<Self::Point, Error>;
 }
 
@@ -299,13 +309,13 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug> FixedPoint<C, 
 #[derive(Clone, Debug)]
 pub struct FixedPointShort<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug> {
     chip: EccChip,
-    inner: EccChip::FixedPoint,
+    inner: EccChip::FixedPointShort,
 }
 
 impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug> FixedPointShort<C, EccChip> {
     /// Gets a reference to the specified fixed point in the circuit.
     pub fn get(chip: EccChip, point: EccChip::FixedPoints) -> Result<Self, Error> {
-        chip.get_fixed(point)
+        chip.get_fixed_short(point)
             .map(|inner| FixedPointShort { chip, inner })
     }
 
--- a/src/circuit/gadget/ecc.rs
+++ b/src/circuit/gadget/ecc.rs
@@ -207,11 +207,26 @@ impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug> Point<C, EccCh
         Point { chip, inner }
     }
 
-    /// Returns `self + other`.
+    /// Returns `self + other` using complete addition.
     pub fn add(&self, mut layouter: impl Layouter<C::Base>, other: &Self) -> Result<Self, Error> {
         assert_eq!(format!("{:?}", self.chip), format!("{:?}", other.chip));
         self.chip
             .add(&mut layouter, &self.inner, &other.inner)
+            .map(|inner| Point {
+                chip: self.chip.clone(),
+                inner,
+            })
+    }
+
+    /// Returns `self + other` using incomplete addition.
+    pub fn add_incomplete(
+        &self,
+        mut layouter: impl Layouter<C::Base>,
+        other: &Self,
+    ) -> Result<Self, Error> {
+        assert_eq!(format!("{:?}", self.chip), format!("{:?}", other.chip));
+        self.chip
+            .add_incomplete(&mut layouter, &self.inner, &other.inner)
             .map(|inner| Point {
                 chip: self.chip.clone(),
                 inner,
--- a/src/circuit/gadget/ecc.rs
+++ b/src/circuit/gadget/ecc.rs
@@ -12,129 +12,227 @@ use halo2::{
 pub trait FixedPoints<C: CurveAffine>: Clone + fmt::Debug {}
 
 /// The set of circuit instructions required to use the ECC gadgets.
-pub trait EccInstructions<C: CurveAffine>: Chip<Field = C::Base> {
-    /// Variable representing an element of the elliptic curve's scalar field.
-    type Scalar: Clone + fmt::Debug;
+pub trait EccInstructions<C: CurveAffine>: Chip<C::Base> {
+    /// Variable representing a full-width element of the elliptic curve's scalar field, to be used for fixed-base scalar mul.
+    type ScalarFixed: Clone + fmt::Debug;
+    /// Variable representing a signed short element of the elliptic curve's scalar field, to be used for fixed-base scalar mul.
+    type ScalarFixedShort: Clone + fmt::Debug;
     /// Variable representing an elliptic curve point.
     type Point: Clone + fmt::Debug;
+    /// Variable representing the x-coordinate of an elliptic curve point.
+    type X: Clone + fmt::Debug;
     /// Variable representing the set of fixed bases in the circuit.
     type FixedPoints: FixedPoints<C>;
     /// Variable representing a fixed elliptic curve point (constant in the circuit).
     type FixedPoint: Clone + fmt::Debug;
 
-    /// Witnesses the given scalar as a private input to the circuit.
-    fn witness_scalar(
-        layouter: &mut impl Layouter<Self>,
+    /// Witnesses the given full-width scalar as a private input to the circuit for fixed-based scalar mul.
+    fn witness_scalar_fixed(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
         value: Option<C::Scalar>,
-    ) -> Result<Self::Scalar, Error>;
+    ) -> Result<Self::ScalarFixed, Error>;
+
+    /// Witnesses the given signed short scalar as a private input to the circuit for fixed-based scalar mul.
+    fn witness_scalar_fixed_short(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        value: Option<C::Scalar>,
+    ) -> Result<Self::ScalarFixedShort, Error>;
 
     /// Witnesses the given point as a private input to the circuit.
     fn witness_point(
-        layouter: &mut impl Layouter<Self>,
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
         value: Option<C>,
     ) -> Result<Self::Point, Error>;
 
+    /// Extracts the x-coordinate of a point.
+    fn extract_p(point: &Self::Point) -> &Self::X;
+
     /// Gets a fixed point into the circuit.
-    fn get_fixed(
-        layouter: &mut impl Layouter<Self>,
-        fixed_points: Self::FixedPoints,
-    ) -> Result<Self::FixedPoint, Error>;
+    fn get_fixed(&self, fixed_points: Self::FixedPoints) -> Result<Self::FixedPoint, Error>;
 
     /// Performs point addition, returning `a + b`.
     fn add(
-        layouter: &mut impl Layouter<Self>,
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
         a: &Self::Point,
         b: &Self::Point,
     ) -> Result<Self::Point, Error>;
 
-    /// Performs point doubling, returning `[2] a`.
-    fn double(layouter: &mut impl Layouter<Self>, a: &Self::Point) -> Result<Self::Point, Error>;
+    /// Performs complete point addition, returning `a + b`.
+    fn add_complete(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        a: &Self::Point,
+        b: &Self::Point,
+    ) -> Result<Self::Point, Error>;
 
     /// Performs variable-base scalar multiplication, returning `[scalar] base`.
     fn mul(
-        layouter: &mut impl Layouter<Self>,
-        scalar: &Self::Scalar,
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        scalar: C::Scalar,
         base: &Self::Point,
     ) -> Result<Self::Point, Error>;
 
-    /// Performs fixed-base scalar multiplication, returning `[scalar] base`.
+    /// Performs fixed-base scalar multiplication using a full-width scalar, returning `[scalar] base`.
     fn mul_fixed(
-        layouter: &mut impl Layouter<Self>,
-        scalar: &Self::Scalar,
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        scalar: &Self::ScalarFixed,
+        base: &Self::FixedPoint,
+    ) -> Result<Self::Point, Error>;
+
+    /// Performs fixed-base scalar multiplication using a short signed scalar, returning `[scalar] base`.
+    fn mul_fixed_short(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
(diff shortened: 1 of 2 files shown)


src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
+        layouter: &mut impl Layouter<C::Base>,
         value: Option<C>,
     ) -> Result<Self::Point, Error>;
 
+    /// Extracts the x-coordinate of a point.
+    fn extract_p(point: &Self::Point) -> &Self::X;
+
     /// Gets a fixed point into the circuit.
-    fn get_fixed(
-        layouter: &mut impl Layouter<Self>,
```
str4d, line 39: I presume the removal of the layouter from this API means it turned out to not be necessary.

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
         a: &Self::Point,
         b: &Self::Point,
     ) -> Result<Self::Point, Error>;
 
-    /// Performs point doubling, returning `[2] a`.
-    fn double(layouter: &mut impl Layouter<Self>, a: &Self::Point) -> Result<Self::Point, Error>;
+    /// Performs complete point addition, returning `a + b`.
+    fn add_complete(
```
str4d, line 65: Hmm, I think I would prefer that `EccInstructions::add` be complete addition, and `EccInstructions::add_incomplete` be incomplete addition; it's the latter that a user needs to be careful with.

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
+        b: &Self::Point,
+    ) -> Result<Self::Point, Error>;
 
     /// Performs variable-base scalar multiplication, returning `[scalar] base`.
     fn mul(
-        layouter: &mut impl Layouter<Self>,
-        scalar: &Self::Scalar,
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        scalar: C::Scalar,
```
str4d, line 76: This can't work; the scalar needs to exist in the circuit somewhere. Re-introduce the `Scalar` associated type for this (or name it `ScalarVar` or something).

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
+        layouter: &mut impl Layouter<C::Base>,
+        scalar: &Self::ScalarFixedShort,
         base: &Self::FixedPoint,
     ) -> Result<Self::Point, Error>;
 }
 
-/// An element of the given elliptic curve's scalar field.
+/// A full-width element of the given elliptic curve's scalar field, to be used for fixed-base scalar mul.
 #[derive(Debug)]
-pub struct Scalar<C: CurveAffine, EccChip: EccInstructions<C>> {
```
str4d, line 70: Re-introduce `Scalar` with `EccChip::Scalar` as the inner type (or `ScalarVar` etc if that's what it gets named).

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
     pub fn new(
-        mut layouter: impl Layouter<EccChip>,
+        chip: EccChip,
+        mut layouter: impl Layouter<C::Base>,
         value: Option<C::Scalar>,
     ) -> Result<Self, Error> {
-        EccChip::witness_scalar(&mut layouter, value).map(|inner| Scalar { inner })
+        chip.witness_scalar_fixed_short(&mut layouter, value)
+            .map(|inner| ScalarFixedShort {
+                chip: chip.clone(),
```
str4d, line 135: No need to clone here, since `chip` is passed by value (implying the caller has already cloned it if necessary).
```suggestion
                chip,
```

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
+impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone> ScalarFixed<C, EccChip> {
+    /// Constructs a new ScalarFixed with the given value.
+    pub fn new(
+        chip: EccChip,
+        mut layouter: impl Layouter<C::Base>,
+        value: Option<C::Scalar>,
+    ) -> Result<Self, Error> {
+        chip.witness_scalar_fixed(&mut layouter, value)
+            .map(|inner| ScalarFixed {
+                chip: chip.clone(),
```
str4d, line 113:
```suggestion
                chip,
```

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
     inner: EccChip::Point,
 }
 
-impl<C: CurveAffine, EccChip: EccInstructions<C>> Point<C, EccChip> {
+impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone> Point<C, EccChip> {
     /// Constructs a new point with the given value.
-    pub fn new(mut layouter: impl Layouter<EccChip>, value: Option<C>) -> Result<Self, Error> {
-        EccChip::witness_point(&mut layouter, value).map(|inner| Point { inner })
+    pub fn new(
+        &self,
```
str4d, line 151: This is a constructor, so it can't take `&self`:
```suggestion
        chip: EccChip,
```

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
+        mut layouter: impl Layouter<C::Base>,
+        value: Option<C>,
+    ) -> Result<Self, Error> {
+        self.chip
```
str4d, line 155:
```suggestion
        chip
```

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
+            .witness_point(&mut layouter, value)
+            .map(|inner| Point {
+                chip: self.chip.clone(),
```
str4d, line 158:
```suggestion
                chip,
```
You might need to define a temporary with the result of `chip.witness_point()` to make the lifetimes work for this map.

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
     /// Witnesses the given point as a private input to the circuit.
     fn witness_point(
-        layouter: &mut impl Layouter<Self>,
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
         value: Option<C>,
     ) -> Result<Self::Point, Error>;
 
+    /// Extracts the x-coordinate of a point.
+    fn extract_p(point: &Self::Point) -> &Self::X;
```
str4d, line 51: This places a restriction on the `Self::Point` type that it contain `Self::X`. This is probably fine (given that this instruction doesn't take a layouter, a chip implementor is forced to store the x-coordinate directly as part of the point, rather than using an alternate representation, but in the circuit they will likely do that anyway).

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
-    /// Returns `[2] self`.
-    pub fn double(&self, mut layouter: impl Layouter<EccChip>) -> Result<Self, Error> {
-        EccChip::double(&mut layouter, &self.inner).map(|inner| Point { inner })
+    /// Wraps the given point (obtained directly from an instruction) in a gadget.
+    pub fn from_inner(chip: EccChip, inner: EccChip::Point) -> Self {
+        Point { chip, inner }
+    }
+
+    /// Returns `self + other`.
+    pub fn add(&self, mut layouter: impl Layouter<C::Base>, other: &Self) -> Result<Self, Error> {
```
str4d, line 174:
```suggestion
    pub fn add(&self, mut layouter: impl Layouter<C::Base>, other: &Self) -> Result<Self, Error> {
        assert_eq!(self.chip, other.chip);
```

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
     /// Returns `[by] self`.
-    pub fn mul(
-        &self,
-        mut layouter: impl Layouter<EccChip>,
-        by: &Scalar<C, EccChip>,
-    ) -> Result<Self, Error> {
-        EccChip::mul(&mut layouter, &by.inner, &self.inner).map(|inner| Point { inner })
+    pub fn mul(&self, mut layouter: impl Layouter<C::Base>, by: C::Scalar) -> Result<Self, Error> {
+        self.chip
+            .mul(&mut layouter, by, &self.inner)
```
str4d, line 186: After `Scalar` (or `ScalarVar` or whatever) is reintroduced:
```suggestion
    pub fn mul(&self, mut layouter: impl Layouter<C::Base>, by: &Scalar<C, EccChip>) -> Result<Self, Error> {
        assert_eq!(self.chip, by.chip);
        self.chip
            .mul(&mut layouter, &by.inner, &self.inner)
```

src/circuit/gadget/ecc.rs
```diff
@@ -12,129 +12,227 @@ use halo2::{
...
     }
 
     /// Returns `[by] self`.
     pub fn mul(
         &self,
-        mut layouter: impl Layouter<EccChip>,
-        by: &Scalar<C, EccChip>,
+        mut layouter: impl Layouter<C::Base>,
+        by: &ScalarFixed<C, EccChip>,
     ) -> Result<Point<C, EccChip>, Error> {
```
str4d, line 230:
```suggestion
    ) -> Result<Point<C, EccChip>, Error> {
        assert_eq!(self.chip, by.chip);
```

str4d (changes requested):

src/circuit/gadget/ecc.rs
therealyingtong: `get_fixed()` does not need to take a layouter because it doesn't assign any new cells. Instead, it returns references to pre-loaded cells, which can be copied by the caller into whichever part of the circuit they need.

src/circuit/gadget/ecc.rs
therealyingtong: should `Point` use complete addition by default?

str4d: Yea. Now that we expose both, it should have an `add_incomplete` method as well.

src/circuit/gadget/ecc.rs
```diff
@@ -1,140 +1,336 @@
...
         value: Option<C::Scalar>,
-    ) -> Result<Self::Scalar, Error>;
+    ) -> Result<Self::ScalarFixed, Error>;
+
+    /// Witnesses the given signed short scalar as a private input to the circuit for fixed-base scalar mul.
+    fn witness_scalar_fixed_short(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
+        value: Option<C::Scalar>,
+    ) -> Result<Self::ScalarFixedShort, Error>;
```
str4d, line 60: What defines a scalar as "short"? None of the short types or methods indicate any bound on the length of the scalar. This API in particular just takes a `C::Scalar` which could be any width.

This API should specify the precise bound it enforces. Ideally we could type that bound on a const generic parameter, but for now I'd be fine with a documented runtime check.

src/circuit/gadget/ecc.rs
```diff
@@ -1,140 +1,336 @@
...
+    /// parts of the circuit where the fixed base is used.
+    fn get_fixed(&self, fixed_points: Self::FixedPoints) -> Result<Self::FixedPoint, Error>;
+
+    /// Returns a fixed point to be used in scalar multiplication with a signed
+    /// short exponent.
+    fn get_fixed_short(
+        &self,
         fixed_points: Self::FixedPoints,
-    ) -> Result<Self::FixedPoint, Error>;
+    ) -> Result<Self::FixedPointShort, Error>;
```
str4d, line 82: Is the intent that all fixed bases can be fetched in either full-width or short form? That makes sense generically, but given that fixed bases are loaded into the circuit ahead of time, that implies we pay the cost of loading all of them in both full and short form (which could be fine if the short form can just be a subset of the full form). If this is the case, please document this on `EccInstructions::FixedPoints`. If not, then I guess we also need an `EccInstructions::FixedPointsShort`.

src/circuit/gadget/ecc.rs
```diff
@@ -1,140 +1,336 @@
...
 
-    /// Performs point addition, returning `a + b`.
-    fn add(
-        layouter: &mut impl Layouter<Self>,
+    /// Performs incomplete point addition, returning `a + b`.
```
str4d, line 84: Document the expected behaviour on exceptional cases (either "always returns an error", or "deterministic but undefined behaviour"). We could propagate the bottom case out of this API, but that would require tracking it in the circuit which we don't want to do for performance.

str4d (approved):
utACK after some documentation is added.

src/circuit/gadget/ecc.rs
therealyingtong: The short scalar must be in the range [-(2^64 - 1), (2^64-1)]. I've added a range check that the short scalar's magnitude is less than `1<<64`.

src/circuit/gadget/ecc.rs
therealyingtong: The short form is not a subset of the full form because the last `u`- and `z`- values are different. We load `ValueCommitV` as a short fixed base, and the other four bases as full-width. I've added `EccInstructions::FixedPointsShort`.

src/circuit/gadget/ecc.rs
```diff
@@ -1,140 +1,362 @@
...
+    /// scalar field `Fq` is larger than its base field `Fp`.
+    ///
+    /// However, the only use of variable-base scalar mul in the Orchard protocol
+    /// is in deriving diversified addresses `[ivk] g_d`,  and `ivk` is guaranteed
+    /// to be in the base field of the curve. (See non-normative notes in
+    /// https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents.)
+    type ScalarVar: Clone + Debug;
+    /// Variable representing a full-width element of the elliptic curve's scalar field, to be used for fixed-base scalar mul.
+    type ScalarFixed: Clone + Debug;
+    /// Variable representing a signed short element of the elliptic curve's scalar field, to be used for fixed-base scalar mul.
```
daira, line 28: Wrap this line and the comment above.

src/circuit/gadget/ecc.rs
```diff
@@ -1,140 +1,362 @@
...
+    /// with a short signed exponent.
+    type FixedPointsShort: Clone + Debug;
     /// Variable representing a fixed elliptic curve point (constant in the circuit).
-    type FixedPoint: Clone + fmt::Debug;
+    type FixedPoint: Clone + Debug;
+    /// Variable representing a fixed elliptic curve point (constant in the circuit)
+    /// to be used in scalar multiplication with a short signed exponent.
+    type FixedPointShort: Clone + Debug;
+
+    /// Witnesses the given base field element as a private input to the circuit for variable-base scalar mul.
```
daira, line 47: Wrap this line.

src/circuit/gadget/ecc.rs
```diff
@@ -1,140 +1,362 @@
...
+    /// Variable representing a full-width element of the elliptic curve's scalar field, to be used for fixed-base scalar mul.
+    type ScalarFixed: Clone + Debug;
+    /// Variable representing a signed short element of the elliptic curve's scalar field, to be used for fixed-base scalar mul.
+    ///
+    /// A `ScalarFixedShort` must be in the range [-(2^64 - 1), 2^64 - 1].
+    type ScalarFixedShort: Clone + Debug;
     /// Variable representing an elliptic curve point.
-    type Point: Clone + fmt::Debug;
+    type Point: Clone + Debug;
+    /// Variable representing the x-coordinate of an elliptic curve point.
```
daira, line 34:
```suggestion
    /// Variable representing the affine short Weierstrass x-coordinate of an
    /// elliptic curve point.
```

src/circuit/gadget/ecc.rs
daira, line 47: Wrap this line.

src/circuit/gadget/ecc.rs
daira: My preferred behaviour would be to return Error when incomplete addition is defined to return ⊥. (The prover can always check that even if the circuit doesn't.)

src/circuit/gadget/ecc.rs
```diff
@@ -1,140 +1,362 @@
...
+    /// This returns an error in exceptional cases.
+    fn add_incomplete(
+        &self,
+        layouter: &mut impl Layouter<C::Base>,
         a: &Self::Point,
         b: &Self::Point,
     ) -> Result<Self::Point, Error>;
 
-    /// Performs point doubling, returning `[2] a`.
-    fn double(layouter: &mut impl Layouter<Self>, a: &Self::Point) -> Result<Self::Point, Error>;
```
daira, line 51: We don't need doubling for Orchard, but it might be useful in general, when this gadget starts being used for other circuits. Please file a ticket to add it later.

src/circuit/gadget/ecc.rs
```diff
@@ -1,140 +1,362 @@
...
+    pub fn from_inner(chip: EccChip, inner: EccChip::Point) -> Self {
+        Point { chip, inner }
     }
 
-    /// Returns `[2] self`.
-    pub fn double(&self, mut layouter: impl Layouter<EccChip>) -> Result<Self, Error> {
-        EccChip::double(&mut layouter, &self.inner).map(|inner| Point { inner })
+    /// Returns `self + other` using complete addition.
+    pub fn add(&self, mut layouter: impl Layouter<C::Base>, other: &Self) -> Result<Self, Error> {
+        assert_eq!(format!("{:?}", self.chip), format!("{:?}", other.chip));
```
daira, line 248: Add Eq to Chip and EccChip, so that we don't have to explicitly compare debug representations.

src/circuit/gadget/ecc.rs
therealyingtong: We could reintroduce it into this gadget now?

daira: Up to you. We don't want to make the code unnecessarily hard to audit, but a double implementation is probably quite simple and makes sense as part of the API, I think.

src/circuit/gadget/ecc.rs
```diff
@@ -1,140 +1,368 @@
...
+        self.chip
+            .mul(&mut layouter, &by.inner, &self.inner)
+            .map(|inner| Point {
+                chip: self.chip.clone(),
+                inner,
+            })
+    }
+}
+
+/// The x-coordinate of an elliptic curve point over the given curve.
```
daira, line 294:
```suggestion
/// The affine short Weierstrass x-coordinate of an elliptic curve point over the
/// given curve.
```

src/circuit/gadget/ecc.rs
```diff
@@ -1,140 +1,368 @@
...
+
+impl<C: CurveAffine, EccChip: EccInstructions<C> + Clone + Debug> X<C, EccChip> {
+    /// Wraps the given x-coordinate (obtained directly from an instruction) in a gadget.
+    pub fn from_inner(chip: EccChip, inner: EccChip::X) -> Self {
+        X { chip, inner }
     }
 }
 
 /// A constant elliptic curve point over the given curve, for which scalar multiplication
 /// is more efficient.
```
daira, line 319: "for which window tables have been provided to make scalar multiplication more efficient."

src/circuit/gadget/ecc.rs
```diff
@@ -1,140 +1,378 @@
...
     /// Returns `[by] self`.
     pub fn mul(
         &self,
-        mut layouter: impl Layouter<EccChip>,
-        by: &Scalar<C, EccChip>,
+        mut layouter: impl Layouter<C::Base>,
+        by: &ScalarVar<C, EccChip>,
     ) -> Result<Self, Error> {
-        EccChip::mul(&mut layouter, &by.inner, &self.inner).map(|inner| Point { inner })
+        assert_eq!(format!("{:?}", self.chip), format!("{:?}", by.chip));
```
daira, line 293:
```suggestion
        assert_eq!(self.chip, by.chip);
```

daira (approved):
utACK modulo comments.

str4d (approved):

str4d:
Gah, I missed that this was based on #43, sorry!

PR #77, opened 2021-05-08
Remove rand 0.7 usage

str4d:
Upstream `redjubjub` (on which our `reddsa` dependency is based) has migrated to `rand 0.8`.


daira (approved):
utACK

PR #80, opened 2021-05-11
Update implementation to match protocol spec version 2021.2.0

### LINKED ISSUE #79: Update implementation of Orchard keys / primitives to match version 2021.2.0 of the spec
Since we implemented e.g. key components in #40, there have been some changes to the protocol spec (e.g. excluding zero points and scalars from most types). We need to go through and ensure that this implementation matches the spec.

str4d:
The implementation was previously (supposed to be) implementing version 2021.1.20 of the protocol spec.

Closes #79.

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -328,7 +328,7 @@ impl From<&FullViewingKey> for KeyAgreementPrivateKey {
 }
 
 impl KeyAgreementPrivateKey {
-    /// Derives ask from sk. Internal use only, does not enforce all constraints.
+    /// Derives ivk from fvk. Internal use only, does not enforce all constraints.
     fn derive_inner(fvk: &FullViewingKey) -> CtOption<NonZeroPallasBase> {
         let ak = extract_p(&pallas::Point::from_bytes(&(&fvk.ak.0).into()).unwrap());
         commit_ivk(&ak, &fvk.nk.0, &fvk.rivk.0)
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -16,7 +16,7 @@ use crate::{
     primitives::redpallas::{self, SpendAuth},
     spec::{
         commit_ivk, diversify_hash, extract_p, ka_orchard, prf_expand, prf_expand_vec, prf_nf,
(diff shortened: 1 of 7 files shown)


src/spec.rs
```diff
@@ -60,21 +118,23 @@ pub(crate) fn commit_ivk(
...
     let hasher = pallas::Point::hash_to_curve("z.cash:Orchard-gd");
     let pk_d = hasher(d);
-    if pk_d.is_identity().into() {
-        // If the identity occurs, we replace it with a different fixed point.
-        hasher(&[])
-    } else {
-        pk_d
-    }
+    // If the identity occurs, we replace it with a different fixed point.
+    NonIdentityPallasPoint(CtOption::new(pk_d, !pk_d.is_identity()).unwrap_or_else(|| hasher(&[])))
```
str4d, line 137:
```suggestion
    // TODO: Replace the unwrap_or_else with a cached fixed point.
    NonIdentityPallasPoint(CtOption::new(pk_d, !pk_d.is_identity()).unwrap_or_else(|| hasher(&[])))
```

src/keys.rs
```diff
@@ -297,6 +298,49 @@ impl Diversifier {
...
+impl From<&FullViewingKey> for KeyAgreementPrivateKey {
+    fn from(fvk: &FullViewingKey) -> Self {
+        // KeyAgreementPrivateKey cannot be constructed such that this unwrap would fail.
+        let ivk = KeyAgreementPrivateKey::derive_inner(fvk).unwrap();
+        KeyAgreementPrivateKey(ivk.into())
+    }
+}
+
+impl KeyAgreementPrivateKey {
+    /// Derives ask from sk. Internal use only, does not enforce all constraints.
```
therealyingtong, line 331:
```suggestion
    /// Derives ivk from fvk. Internal use only, does not enforce all constraints.
```

therealyingtong (approved):

daira (approved):
utACK with comments (with @therealyingtong). I have *not* checked that this implements all of the changes made to the spec, only that the changes that were made are correct.

PR #86, opened 2021-05-24
Update formula for fixed-base window tables.

therealyingtong:
Review suggestion: the logical changes are in the first commit. The second commit only updates the inlined constants.

Previously, the window table M for fixed-base scalar multiplication
computed `M[w][k] = [(k+1)*(2^3)^w]B` for each window `w`, where `k` is a
3-bit chunk in the scalar decomposition in the range `[0..8)`.

However, in the case `k_0 = 7`, `k_1 = 0`, the window table entries would
evaluate to:
* `M[0][k_0] = [(7+1)*(2^3)^0]B = [8]B`,
* `M[1][k_1] = [(0+1)*(2^3)^1]B = [8]B`,
which means the first addition would require complete addition.

To avoid this, we alter the formula to `M[w][k] = [(k+2)*(2^3)^w]B`.

We make a corresponding change to the formula for the last window
`W`. Previously, we had:
`M[W][k] = [k * (2^3)^W - \sum((2^3)^j)]B`, for `j` in `[0..W-1)`.

Now, we have:
`M[W][k] = [k * (2^3)^W - \sum(2^(3j+1))]B`, for `j` in `[0..W-1)`.


daira (approved):
utACK

src/constants.rs
```diff
@@ -107,14 +107,14 @@ impl<C: CurveAffine> FixedBase<C> for OrchardFixedBase<C> {
         let mut window_table: Vec<[C; H]> = Vec::with_capacity(num_windows);
 
         // Generate window table entries for all windows but the last.
-        // For these first `num_windows - 1` windows, we compute the multiple [(k+1)*(8^w)]B.
+        // For these first `num_windows - 1` windows, we compute the multiple [(k+2)*(2^3)^w]B.
         // Here, w ranges from [0..`num_windows - 1`)
         for w in 0..(num_windows - 1) {
```
str4d, line 112:
```suggestion
        for w in 0..(num_windows - 1) {
            let H_w = C::ScalarExt::from_u64(H as u64).pow(&[w as u64, 0, 0, 0]);
```

src/constants.rs
```diff
@@ -107,14 +107,14 @@ impl<C: CurveAffine> FixedBase<C> for OrchardFixedBase<C> {
...
             window_table.push(
                 (0..H)
                     .map(|k| {
-                        // scalar = (k+1)*(8^w)
-                        let scalar = C::ScalarExt::from_u64(k as u64 + 1)
+                        // scalar = (k+2)*(8^w)
+                        let scalar = C::ScalarExt::from_u64(k as u64 + 2)
                             * C::ScalarExt::from_u64(H as u64).pow(&[w as u64, 0, 0, 0]);
```
str4d, line 118:
```suggestion
                        let scalar = C::ScalarExt::from_u64(k as u64 + 2) * H_w;
```

str4d (approved):
utACK

PR #91, opened 2021-05-28
Add test vectors for key components

str4d:
Source:
- https://github.com/zcash-hackworks/zcash-test-vectors/pull/29
- https://github.com/zcash-hackworks/zcash-test-vectors/pull/32


daira (approved):
utACK. I have *not* checked that these correspond to the output of the Python code in zcash-test-vectors.

PR #93, opened 2021-05-28
Define explicit domains for PRF^expand

str4d:
`prf_expand{_vec}` have been replaced by the `PrfExpand` enum, which has `PrfExpand::{expand, with_ad, with_ad_slices}` methods for use within each domain as necessary.

### DIFF
--- a/src/spec/prf_expand.rs
+++ b/src/spec/prf_expand.rs
@@ -2,23 +2,70 @@ use blake2b_simd::Params;
 
 const PRF_EXPAND_PERSONALIZATION: &[u8; 16] = b"Zcash_ExpandSeed";
 
-/// $PRF^\mathsf{expand}(sk, t) := BLAKE2b-512("Zcash_ExpandSeed", sk || t)$
-///
-/// Defined in [Zcash Protocol Spec § 5.4.2: Pseudo Random Functions][concreteprfs].
-///
-/// [concreteprfs]: https://zips.z.cash/protocol/nu5.pdf#concreteprfs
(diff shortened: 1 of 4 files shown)


str4d:
I think this would be a good candidate for living in a "real" `zcash_primitives` crate - one that contains the fundamental types we reuse across multiple protocols, that need cross-protocol consensus. But for now, it can live here.

src/spec/prf_expand.rs
```diff
@@ -0,0 +1,71 @@
+use blake2b_simd::Params;
+
+const PRF_EXPAND_PERSONALIZATION: &[u8; 16] = b"Zcash_ExpandSeed";
+
+/// The set of domains in which $PRF^\mathsf{expand}$ is defined.
```
daira, line 5:
```suggestion
/// The set of domains in which $PRF^\mathsf{expand}$ is defined.
/// This currently only includes domains used in Orchard.
/// TODO: refactor uses of $PRF^\mathsf{expand}$ in Sapling.
```

daira (approved):
utACK

src/spec/prf_expand.rs
str4d: I don't see a need to reference Sapling inside the `orchard` crate; instead this should be an issue over in the `librustzcash` repo.

str4d: Opened https://github.com/zcash/librustzcash/issues/391.

nuttycom (approved):
utACK

I did not re-verify the constants against the spec, but the behavior is unchanged.

PR #99, opened 2021-06-02
Note encryption

### LINKED ISSUE #32: Implement note encryption for Orchard
Ideally we will build this on a reusable component shared with Sapling.

- https://zips.z.cash/protocol/nu5.pdf#saplingandorchardinband

str4d:
Closes #32.


src/note.rs
```diff
@@ -44,8 +53,18 @@ impl RandomSeed {
...
+
+    /// Defined in [Zcash Protocol Spec § 4.7.3: Sending Notes (Orchard)][orchardsend].
+    ///
+    /// [orchardsend]: https://zips.z.cash/protocol/nu5.pdf#orchardsend
+    fn esk(&self, rho: &Nullifier) -> NonZeroPallasScalar {
+        // We can't construct a RandomSeed for which this unwrap fails.
+        self.esk_inner(rho).unwrap()
     }
 
     /// Defined in [Zcash Protocol Spec § 4.7.3: Sending Notes (Orchard)][orchardsend].
```
therealyingtong, line 70: Could we move `esk()`, `rcm()`, `psi()` onto the `Note impl`? That way they don't have to take an additional `rho` argument.

str4d: I'm not sure about moving them (I want the definitions to be close to each other), but we can also have Note methods (in fact I think this is what we do for at least one of them).

str4d:
[Rebased](https://github.com/zcash/orchard/compare/69a595365c503c58239debfe263db26c1896b57c..3cb8e85f58109c063643d11a120b1c3224d32f27) to fix merge conflict in imports.

nuttycom (approved):
utACK. I've checked this against the NU5 spec and didn't find any divergence.

str4d:
Rebased on main to fix merge conflict with #110.

ebfull (approved):

PR #101, opened 2021-06-04
Action circuit

### LINKED ISSUE #4: Implement the Action circuit
- https://zips.z.cash/protocol/nu5.pdf#actionstatement

### LINKED ISSUE #153: Undo individual non-binary selector optimizations
While implementing the various chips used in the Action circuit, we implemented a number of selector optimisations, of two forms:
- If we needed a non-zero constant, we used an assignment of zero as the disabling selector.
- In Sinsemilla, `q_sinsemilla2` merges two boolean selectors into one non-boolean selector.

We are in the process of implementing https://github.com/zcash/halo2/issues/116, after which we will need to undo all of these optimizations:
- Non-zero constant columns as selectors "encumber" the entire fixed column, preventing us from sharing them with other chips. If we split apart the "selector" and "constant" behaviour, we can share the constant columns, while the selector columns will be automatically merged away.
- Non-boolean selectors cannot be automatically merged with boolean selectors. If we un-merge them on an individual level, we will likely be able to remove their currently-used column entirely.

### LINKED ISSUE #157: Remove Standard PLONK from the Action circuit and rely on selector combining
Alternative to #154.

therealyingtong:
Closes #4. Closes #157. Closes #153.

Canonicity checks for Sinsemilla inputs documented at #134.


src/circuit.rs
```diff
@@ -1,46 +1,666 @@
...
+                layouter.namespace(|| "witness nk"),
+                config.advices[0],
+                self.nk.map(|nk| *nk),
+            )?;
+
+            (v_old, v_new, rho_old, psi_old, cm_old, g_d_old, ak, nk)
+        };
+
+        // Merkle path validity check.
+        // TODO: constrain output to equal public input
```
therealyingtong, line 329: More precisely: require that either `v_old = 0`, or the computed anchor equals the public input anchor.

src/circuit.rs
```diff
@@ -1,46 +1,658 @@
...
+
+            // Add hash output to psi using standard PLONK
+            // `scalar` = poseidon_hash(nk, rho_old) + psi_old.
+            //
+            let scalar = {
+                let scalar_val = nk_rho_old
+                    .value()
+                    .zip(psi_old.value())
+                    .map(|(nk_rho_old, psi_old)| nk_rho_old + psi_old);
+                let scalar = self.load_private(
```
naure, line 438: It seems that this variable is not connected to anything? It should be returned from this block instead of `ScalarFixed`.

therealyingtong: Thanks! This required us to support multiplying a fixed base by a base field element. `scalar` there is a base field element (a result of an addition `mod p`). It is now directly used in

```rust
nullifier_k.mul_base_field_elem(
    layouter.namespace(|| "[poseidon_output + psi_old] NullifierK"),
    scalar,
)?
```

src/circuit.rs
```diff
@@ -2,55 +2,669 @@
...
+        // Constrain v_old - v_new = magnitude * sign
+        let q_v_net = meta.selector();
+        meta.create_gate("v_old - v_new = magnitude * sign", |meta| {
+            let q_v_net = meta.query_selector(q_v_net);
+            let v_old = meta.query_advice(advices[0], Rotation::cur());
+            let v_new = meta.query_advice(advices[1], Rotation::cur());
+            let magnitude = meta.query_advice(advices[2], Rotation::cur());
+            let sign = meta.query_advice(advices[3], Rotation::cur());
+
+            vec![q_v_net * (v_old - v_new - magnitude * sign)]
```
daira, line 149: Can we do this with the standard PLONK gate?

src/circuit.rs
```diff
@@ -2,55 +2,669 @@
...
+        // Configuration to handle decomposition and canonicity checking
+        // for NoteCommit_new.
+        let new_note_commit_config =
+            NoteCommitConfig::configure(meta, advices, sinsemilla_config_2.clone());
+
+        // TODO: Infrastructure to handle public inputs.
+        let q_primary = meta.selector();
+        let primary = meta.instance_column();
 
         // Placeholder gate so there is something for the prover to operate on.
```
daira, line 261: Why do we still need this?

src/circuit.rs
```diff
@@ -2,55 +2,669 @@
...
+                chip_1: config.merkle_chip_1(),
+                chip_2: config.merkle_chip_2(),
+                domain: SinsemillaHashDomains::MerkleCrh,
+                leaf_pos: self.pos,
+                path: self.path,
+            };
+            let leaf = *cm_old.extract_p().inner();
+            merkle_inputs.calculate_root(layouter.namespace(|| "MerkleCRH"), leaf)?
+        };
+
```
daira, line 381: Up to here; breaking for lunch.

src/circuit.rs
str4d: @therealyingtong is still in the process of migrating this PR to use the new `halo2` public input APIs.

src/circuit.rs
```diff
@@ -50,13 +270,401 @@ impl plonk::Circuit<pallas::Base> for Circuit {
...
+        let _cv_net = {
+            // v_net = v_old - v_new
+            let v_net = {
+                let v_net_val = self.v_old.zip(self.v_new).map(|(v_old, v_new)| {
+                    // Do the subtraction in the scalar field.
+                    let v_old = pallas::Scalar::from_u64(v_old.inner());
+                    let v_new = pallas::Scalar::from_u64(v_new.inner());
+                    v_old - v_new
+                });
+                // If v_net_val > (p - 1)/2, its sign is negative.
```
daira, line 393:
```suggestion
                // If v_net_val > (q_V - 1)/2, its sign is negative.
```

src/circuit.rs
```diff
@@ -50,13 +270,401 @@ impl plonk::Circuit<pallas::Base> for Circuit {
...
+            merkle_inputs.calculate_root(layouter.namespace(|| "MerkleCRH"), leaf)?
+        };
+
+        // Value commitment integrity.
+        // TODO: constrain to equal public input cv_net
+        let _cv_net = {
+            // v_net = v_old - v_new
+            let v_net = {
+                let v_net_val = self.v_old.zip(self.v_new).map(|(v_old, v_new)| {
+                    // Do the subtraction in the scalar field.
```
daira, line 388: Because we know that v_old and v_new are in [0, 2<sup>64</sup>), it would be equivalent to do this in the base field, and maybe slightly simpler.

src/circuit.rs
```diff
@@ -50,13 +270,401 @@ impl plonk::Circuit<pallas::Base> for Circuit {
...
+                let rcv = self.rcv.as_ref().map(|rcv| **rcv);
+                let value_commit_r = OrchardFixedBasesFull::ValueCommitR;
+                let value_commit_r = FixedPoint::from_inner(ecc_chip.clone(), value_commit_r);
+
+                // [rcv] ValueCommitR
+                value_commit_r.mul(layouter.namespace(|| "[rcv] ValueCommitR"), rcv)?
+            };
+
+            // [v_net] ValueCommitV + [rcv] ValueCommitR
+            commitment.add(layouter.namespace(|| "cv_net"), &blind)?
```
daira, line 467: Why do we do this at the Action circuit level, rather than having an abstraction for homomorphic Pedersen commitments?

src/circuit.rs
```diff
@@ -50,13 +270,401 @@ impl plonk::Circuit<pallas::Base> for Circuit {
...
+                            )?;
+                            region.constrain_equal(var, message[i].cell())?;
+                            Ok(Word::<_, _, poseidon::OrchardNullifier, 3, 2> {
+                                inner: StateWord::new(var, value),
+                            })
+                        };
+
+                        Ok([message_word(0)?, message_word(1)?])
+                    },
+                )?;
```
daira, line 496: This seems a bit clunky, and should probably be handled by a function in the Poseidon code.

src/circuit.rs
```diff
@@ -50,13 +270,401 @@ impl plonk::Circuit<pallas::Base> for Circuit {
...
+                g_d_old.inner(),
+                pk_d_old.inner(),
+                v_old,
+                rho_old,
+                psi_old,
+                rcm_old,
+            )?
+        };
+
+        // new note commitment integrity.
```
daira, line 613:
```suggestion
        // New note commitment integrity.
        // TODO: constrain to equal public input cm_x
```

src/circuit.rs
```diff
@@ -50,13 +270,401 @@ impl plonk::Circuit<pallas::Base> for Circuit {
...
+                    rivk,
+                )?
+            };
+
+            // [ivk] g_d_old
+            g_d_old.mul(layouter.namespace(|| "[ivk] g_d_old"), ivk.inner())?
+        };
+
+        // Old note commitment integrity.
+        let _cm_old = {
```
daira, line 592: `_cm_old` here has to be constrained equal to `cm_old` used in the Merkle path check.

src/circuit.rs
```diff
@@ -50,13 +270,401 @@ impl plonk::Circuit<pallas::Base> for Circuit {
...
+                let spend_auth_g = FixedPoint::from_inner(ecc_chip.clone(), spend_auth_g);
+                spend_auth_g.mul(layouter.namespace(|| "[alpha] SpendAuthG"), self.alpha)?
+            };
+
+            // [alpha] SpendAuthG + ak
+            alpha_commitment.add(layouter.namespace(|| "rk"), &ak)?
+        };
+
+        // Diversified address integrity.
+        let (pk_d_old, _) = {
```
daira: The discarded value here is ivk, right? I think @str4d mentioned on the ECC PR that it was a bit strange to just pass through the scalar value from the input to the output. At least add a comment.

src/circuit.rs
```diff
@@ -50,13 +270,401 @@ impl plonk::Circuit<pallas::Base> for Circuit {
...
+                pk_d_new.inner(),
+                v_new,
+                *nf_old.inner(),
+                psi_new,
+                rcm_new,
+            )?;
+
+            cm_new.extract_p()
+        };
+
```
daira, line 769: I don't see the **Enable spend flag** or **Enable output flag** checks.

src/circuit.rs
```diff
@@ -72,7 +680,7 @@ impl VerifyingKey {
     /// Builds the verifying key.
     pub fn build() -> Self {
         let params = halo2::poly::commitment::Params::new(K);
-        let circuit = Circuit {}; // TODO
+        let circuit: Circuit = Default::default(); // TODO
```
daira, line 785: Isn't this correct? Why is it TODO? (Similarly below.)

src/circuit.rs
```diff
@@ -215,10 +828,33 @@ mod tests {
...
+                        psi_new: Some(output_note.rseed().psi(&output_note.rho())),
+                        rcm_new: Some(output_note.rseed().rcm(&output_note.rho())),
+                        rcv: Some(ValueCommitTrapdoor::zero()),
+                    },
                     Instance {
-                        anchor: pallas::Base::zero().into(),
+                        anchor,
                         cv_net,
                         nf_old,
                         rk,
```
daira, line 982: This always passes `enable_spend: true`; I gather that means it does not support building coinbase transactions?

src/circuit.rs
```diff
@@ -2,61 +2,771 @@
...
+        commit_ivk::CommitIvkConfig,
+        merkle::{
+            chip::{MerkleChip, MerkleConfig},
+            MerklePath,
+        },
+        note_commit::NoteCommitConfig,
+    },
+    utilities::{
+        copy,
+        enable_flag::{EnableFlagChip, EnableFlagConfig},
```
daira, line 58: Does not compile.

src/circuit.rs
```diff
@@ -2,61 +2,771 @@
...
+const ZERO: usize = 0;
+const ANCHOR: usize = 1;
+const NF_OLD: usize = 2;
+const CV_NET_X: usize = 3;
+const CV_NET_Y: usize = 4;
+const RK_X: usize = 5;
+const RK_Y: usize = 6;
+const CMX: usize = 7;
+const ENABLE_SPEND: usize = 8;
+const ENABLE_OUTPUT: usize = 9;
```
daira, line 82: This doesn't match the spec.
image

ebfull: Hm, it's much simpler (implementation-wise) to bring `enableOutputs` and `enableSpends` in individually as booleans. What is the motivation for packing them together like this?

src/circuit.rs
```diff
@@ -2,61 +2,771 @@
...
+            meta.advice_column(),
+            meta.advice_column(),
+            meta.advice_column(),
+        ];
+
+        // Constrain v_old - v_new = magnitude * sign.
+        // Constrain v_old = 0 or enable_spends = 1.
+        // Constrain v_new = 0 or enable_outputs = 1.
+        let q_value = meta.selector();
+        meta.create_gate("value constrains", |meta| {
```
daira, line 159:
```suggestion
        meta.create_gate("value constraints", |meta| {
```

src/circuit.rs
daira: For Groth16, inputs were quite expensive (they required a scalar multiplication each, for unbatched verification). I don't know to what extent that is still the case for Halo 2. @ebfull ?

daira: If the cost is not significant, I can split enableSpends and enableOutputs into separate field elements. Note that that isn't the only difference between the implementation above and the spec. When the constants API is added, there will be no need for a zero instance cell.

str4d:
I [force-pushed](https://github.com/zcash/orchard/compare/d2d08019c9688fa28a78c31ffc2a56e729ce4105..af3cb1b409148d260953c372d9ec72075469b84c) to fix the last commit (which was missing some deletions after the `utilities::enable_flag` chip was removed.

src/circuit.rs
str4d: I've fixed this: https://github.com/zcash/orchard/pull/101#issuecomment-882814189

src/circuit.rs
str4d: I added that `TODO` when I wrote the circuit stubs, because an empty struct was obviously not correct, and I expected there might be some strucural values that needed preparing. But if `Circuit` implements `Default` then this is definitely correct.

src/circuit.rs
therealyingtong: This has now been implemented:
```
derived_cm_old.constrain_equal(layouter.namespace(|| "cm_old equality"), &cm_old)?;
```

src/circuit.rs
therealyingtong: This has been implemented in https://github.com/zcash/orchard/pull/101/commits/e4eda3a67fc5ed598783990934af57c84471a36a.

src/circuit.rs
therealyingtong: The round_trip test does pass `enable_spend: true` now, but we should test `enable_spend: false` too.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -24,7 +21,7 @@ impl SinsemillaChip {
     pub(super) fn hash_message(
         &self,
         region: &mut Region<'_, pallas::Base>,
-        Q: pallas::Affine,
+        Q: (usize, pallas::Affine), // This tuple contains: (absolute offset of Q_x in public inputs column, Q)
```
str4d, line 24: Hmm, this breaks the reusable chip abstraction: chips should not depend on anything being in particular absolute rows, as that is not portable between different circuits. I'm not sure it's something we can prevent in the `halo2` instance API (since the top-level circuit builder _does_ need to gain access to public inputs somewhere), but its usage should be discouraged for internal chips.

Instead, we should implement https://github.com/zcash/halo2/issues/335, which would make the positioning of constants a layouter concern.

str4d: See also https://github.com/zcash/halo2/pull/341#pullrequestreview-710410868.

therealyingtong: This has been addressed by using https://github.com/zcash/halo2/pull/342.

src/circuit.rs
```diff
@@ -2,61 +2,707 @@
...
 /// Size of the Orchard circuit.
-const K: u32 = 11;
+// FIXME: This circuit should fit within 2^11 rows.
+const K: u32 = 12;
+
+// Absolute offsets for public inputs.
+const ANCHOR: usize = 0;
+const NF_OLD: usize = 1;
+const CV_NET_X: usize = 2;
+const CV_NET_Y: usize = 3;
```
daira, line 75:
```suggestion
const CV_NET_X: usize = 1;
const CV_NET_Y: usize = 2;
const NF_OLD: usize = 3;
```
This change is to match the spec:
image

src/circuit.rs
```diff
@@ -2,61 +2,707 @@
...
+                    "Either v_old = 0, or anchor equals public input",
+                    v_old.clone() * (anchor - pub_input_anchor),
+                ),
+                ("v_old = 0 or enable_spends = 1", v_old * not_enable_spends),
+                (
+                    "v_new = 0 or enable_outputs = 1",
+                    v_new * not_enable_outputs,
+                ),
+            ])
+            .map(move |(name, poly)| (name, q_orchard.clone() * poly))
```
daira, line 186: TODO: replace these by uses of Standard PLONK to reduce the number of gates/selectors.

src/value.rs
```diff
@@ -160,6 +164,14 @@ impl TryFrom<ValueSum> for i64 {
 #[derive(Clone, Debug)]
 pub struct ValueCommitTrapdoor(pallas::Scalar);
 
+impl std::ops::Deref for ValueCommitTrapdoor {
```
str4d, line 167: I'm not sure whether we actually need all these new `Deref` impls (which will be part of the public API); I'd prefer crate-private getters unless we know we will need to unwrap these newtypes outside the crate.

src/value.rs
```diff
@@ -263,6 +275,24 @@ impl ValueCommitment {
     pub fn to_bytes(&self) -> [u8; 32] {
         self.0.to_bytes()
     }
+
+    /// x-coordinate of this value commitment.
+    pub fn x(&self) -> pallas::Base {
```
str4d, line 280: This definitely doesn't need to be public:
```suggestion
    pub(in crate::circuit) fn x(&self) -> pallas::Base {
```

src/value.rs
```diff
@@ -263,6 +275,24 @@ impl ValueCommitment {
...
+        if self.0 == pallas::Point::identity() {
+            pallas::Base::zero()
+        } else {
+            *self.0.to_affine().coordinates().unwrap().x()
+        }
+    }
+
+    /// y-coordinate of this value commitment.
+    pub fn y(&self) -> pallas::Base {
```
str4d, line 289:
```suggestion
    pub(in crate::circuit) fn y(&self) -> pallas::Base {
```

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -135,7 +131,14 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K>
...
 
-                self.range_check(&mut region, z_0, num_words, strict)
+                let zs = self.range_check(&mut region, z_0, num_words)?;
+
+                if strict {
+                    // Constrain the final `z` to be zero.
+                    region.constrain_constant(zs.last().unwrap().cell(), F::zero())?;
+                }
+
+                Ok(zs)
```
str4d, line 141: This can be un-refactored now we're using `Region::constrain_constant` instead of public inputs:
```suggestion
                self.range_check(&mut region, z_0, num_words, strict)
```
(These changes here and below should be squashed into the `constants` commit, after review is complete and ACKed, so we can confirm the squash is a no-op.)

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -148,7 +151,7 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K>
         num_words: usize,
         strict: bool,
     ) -> Result<(CellValue<F>, Vec<CellValue<F>>), Error> {
-        layouter.assign_region(
+        let (z_0, zs) = layouter.assign_region(
```
str4d, line 154:
```suggestion
        layouter.assign_region(
```

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -161,11 +164,18 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K>
                     CellValue::new(cell, value)
                 };
 
-                let zs = self.range_check(&mut region, z_0, num_words, strict)?;
+                let zs = self.range_check(&mut region, z_0, num_words)?;
+
+                if strict {
+                    // Constrain the final `z` to be zero.
+                    region.constrain_constant(zs.last().unwrap().cell(), F::zero())?;
+                }
```
str4d, line 172:
```suggestion
                let zs = self.range_check(&mut region, z_0, num_words, strict)?;
```

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -161,11 +164,18 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K>
...
 
                 Ok((z_0, zs))
             },
-        )
+        )?;
+
+        Ok((z_0, zs))
```
str4d, line 178:
```suggestion
        )
```

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -180,7 +190,6 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K>
         region: &mut Region<'_, F>,
         element: CellValue<F>,
         num_words: usize,
-        strict: bool,
```
str4d, line 183: Undo this deletion.

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -243,13 +252,6 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K>
             zs.push(z);
         }
 
-        if strict {
-            // Constrain the final `z` to be zero.
-            let cell =
-                region.assign_fixed(|| "zero", self.constants, words.len(), || Ok(F::zero()))?;
-            region.constrain_equal(z.cell(), cell)?;
-        }
-
```
str4d, line 252:
```suggestion
        if strict {
            // Constrain the final `z` to be zero.
            region.constrain_constant(zs.last().unwrap().cell(), F::zero())?;
        }
```

src/circuit/gadget/poseidon.rs
```diff
@@ -67,7 +67,7 @@ pub struct Word<
     const T: usize,
     const RATE: usize,
 > {
-    inner: PoseidonChip::Word,
+    pub inner: PoseidonChip::Word,
```
str4d, line 70: Hmm, this should probably be a crate-public getter (which could then be made public later if this chip is extracted).

src/circuit.rs
therealyingtong: We are now doing this in the base field.

str4d (commented):
Flushing comments. I've reviewed everything except `halo2::circuit::{self, gadget::sinsemilla::{commit_ivk, note_commit}}`.

src/circuit/gadget/sinsemilla/commit_ivk.rs
```diff
@@ -0,0 +1,810 @@
...
+};
+
+// <https://zips.z.cash/protocol/nu5.pdf#concretesinsemillacommit>
+// We need to hash `ak || nk` where each of `ak`, `nk` is a field element (255 bits).
+//
+// a = bits 0..=249 of `ak`
+// b = b_0||b_1||b_2` = (bits 250..=253 of `ak`) || (bit 254 of  `ak`) || (bits 0..=4 of  `nk`)
+// c = bits 5..=244 of `nk`
+// d = d_0||d_1` = (bits 245..=253 of `nk`) || (bit 254 of `nk`)
+
```
str4d, line 28: Duplicate comment:
```suggestion
```

Cargo.toml
```diff
@@ -1,3 +1,4 @@
+
```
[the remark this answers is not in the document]
therealyingtong: ¿

src/circuit.rs
therealyingtong: We have decided to remove standard PLONK (as noted in the issue).

src/value.rs
therealyingtong: I have replaced the `Deref` impls I introduced with crate-private getters.

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -135,7 +131,9 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K>
                 // Copy `element` and initialize running sum `z_0 = element` to decompose it.
                 let z_0 = copy(&mut region, || "z_0", self.running_sum, 0, &element)?;
 
-                self.range_check(&mut region, z_0, num_words, strict)
+                let zs = self.range_check(&mut region, z_0, num_words, strict)?;
+
+                Ok(zs)
```
str4d, line 136:
```suggestion
                self.range_check(&mut region, z_0, num_words, strict)
```

src/circuit/gadget/sinsemilla/commit_ivk.rs
```diff
@@ -0,0 +1,802 @@
...
+        // c = bits 5..=244 of `nk`
+        // d = d_0||d_1` = (bits 245..=253 of `nk`) || (bit 254 of `nk`)
+        //
+        // `a`, `b`, `c`, `d` have been constrained by the Sinsemilla hash to be:
+        //   - a: 250 bits,
+        //   - b: 10 bits,
+        //   - c: 240 bits,
+        //   - d: 10 bits
+        //
+        meta.create_gate("CommitIvk canonicity check", |meta| {
```
str4d, line 57: Currently this gate uses the following region layout over 10 advice columns:

image

I would prefer that we use the following layout over 9 advice columns:

image

src/circuit/gadget/sinsemilla/commit_ivk.rs
```diff
@@ -0,0 +1,802 @@
...
+            let domain = CommitDomain::new(
+                sinsemilla_chip,
+                ecc_chip,
+                &SinsemillaCommitDomains::CommitIvk,
+            );
+            domain.short_commit(layouter.namespace(|| "Hash ak||nk"), message, rivk)?
+        };
+
+        let z13_a = zs[0][13];
+        let z14_c = zs[2][14];
```
str4d, line 333: If I'm interpreting the circuit correctly (in particular the comment in `nk_canonicity`), this is a constraint bug:
```suggestion
        let z13_c = zs[2][13];
```
That is, if d<sub>1</sub> = 1, we want to constrain c to 130 bits, and then constrain b<sub>2</sub> + c*2<sup>5</sup> (which will be 135 bits) to be < t<sub>P</sub>. The tests (which use valid inputs) pass because if z<sub>c,13</sub> = 0 then z<sub>c,14</sub> will always be zero.

I suspect the bug occurred because for the b<sub>2</sub> + c*2<sup>5</sup> constraint, we _do_ need to select z<sub>14</sub>.

src/circuit/gadget/sinsemilla/chip/generator_table.rs
```diff
@@ -35,7 +35,7 @@ impl GeneratorTableConfig {
 
         meta.lookup(|meta| {
             let q_s1 = meta.query_selector(config.q_sinsemilla1);
-            let q_s2 = meta.query_fixed(config.q_sinsemilla2, Rotation::cur());
+            let q_s2 = meta.query_selector(config.q_sinsemilla2);
             let q_s3 = {
```
str4d, line 39: This is still deriving a composite value from `q_sinsemilla2`, where it should instead be querying `q_sinsemilla3`.

str4d (changes requested):

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -182,35 +192,30 @@ impl SinsemillaChip {
             // Y_A = (lambda_1 + lambda_2) * (x_a - x_r)
             let Y_A_next = Y_A(meta, Rotation::next());
 
-            // Check that the initial x_A, x_P, lambda_1, lambda_2 are consistent with y_Q.
-            // fixed_y_q * (2 * fixed_y_q - Y_{A,0}) = 0
-            let init_y_q_check = fixed_y_q.clone() * (two.clone() * fixed_y_q - Y_A_cur.clone());
```
str4d, line 206: This change (splitting the combined selector into a selector and an advice cell) adds an extra row to every `hash_to_point` region. Instead, we should split it into a selector and a fixed column, and use the previous gate approach (but gated on the selector instead of itself). The fixed columns for the two Sinsemilla instances can then reuse the first two Lagrange coefficient columns.

str4d (changes requested):

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -342,8 +337,14 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K>
...
-        // Enable lookup for shifted element, to constrain it to 10 bits.
-        self.q_lookup_short.enable(region, 1)?;
+        // Assign 2^{-num_bits} from a fixed column.
+        let inv_two_pow_s = F::from_u64(1 << num_bits).invert().unwrap();
+        region.assign_advice_from_constant(
+            || format!("2^(-{})", num_bits),
+            self.running_sum,
+            2,
+            inv_two_pow_s,
+        )?;
```
str4d, line 347: This makes this region 50% larger (3 rows instead of 2), but I think this is an acceptable trade-off here:
- It changes the gate from using rotations `-1,0` to using `-1,0,1` which other gates already use, so we shouldn't be gaining proof size from it.
- If we instead split the combined selector into a selector and a fixed column, we would need to keep that fixed column around (costing 32 bytes in the proof), because we want these to be layed out next to the fixed-base scalar mul regions (so the Lagrange coefficient columns are occupied).

image

src/circuit/gadget/sinsemilla/merkle/chip.rs
```diff
@@ -270,24 +265,24 @@ impl MerkleInstructions<pallas::Affine, MERKLE_DEPTH_ORCHARD, { sinsemilla::K },
...
                     // Recall that l = MERKLE_DEPTH_ORCHARD - layer - 1.
                     // The layer with 2^n nodes is called "layer n".
-                    let l_plus_1 = (l as u64) + 1;
-                    region.assign_fixed(
-                        || format!("l_plus_1 {}", l_plus_1),
-                        config.l_plus_1,
-                        0,
-                        || Ok(pallas::Base::from_u64(l_plus_1)),
+                    config.q_decompose.enable(&mut region, 0)?;
+                    region.assign_advice_from_constant(
```
str4d, line 281: Given that we had the spare advice cell, I agree with this change. We could have reused a fixed column if necessary (these regions end up layed out in areas where we have plenty of free space), but having a region depend on fewer columns is better in general for layout optimizations, and equality constraints between columns that are already in the global permutation (which is the case for all advice columns and the constants-enabled column) are essentially free.

src/circuit/gadget/utilities/lookup_range_check.rs
str4d: We could alternatively use 2 columns to reduce this back to 2 rows/rotations, and keep more 4-column space free:

image

str4d (changes requested):
Finished reviewing `orchard::circuit::gadget::sinsemilla::commit_ivk`.

src/circuit/gadget/sinsemilla/chip/generator_table.rs
therealyingtong: it works because: since `q_s2` is always 0 or 1 now, `q_s3 = (q_s2) * (1 - q_s2)` is always 0. `q_s3` is used in the expression 
```
m_{i+1} = z_{i} - 2^K * (q_s2 - q_s3) * z_{i + 1}
```
so that reduces to `m_{i+1} = z_{i} - 2^K * (q_s2) * z_{i + 1}` which is actually the correct expression after the selector updates.

therealyingtong: (of course i'm going to just remove `q_s3` though)

src/circuit/gadget/sinsemilla/chip.rs
therealyingtong: This change has been made and squashed. It has also been documented at https://github.com/zcash/orchard/pull/155/commits/78b0ec4e7bc0a712f6cf43817eb7019393ca9bc5.

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -0,0 +1,1499 @@
...
+                .map(move |poly| d_0.clone() * poly);
+
+            // The rho_canonicity_checks are enforced if and only if `g_0` = 1.
+            // rho = e_1 (4 bits) || f (250 bits) || g_0 (1 bit)
+            let rho_canonicity_checks = std::iter::empty()
+                .chain(Some(z13_f))
+                .chain(Some(e1_f_prime_decomposition))
+                .map(move |poly| g_0.clone() * poly);
+
+            // The psi_canonicity_checks are enforced if and only if `i_1` = 1.
```
str4d, line 381:
```suggestion
            // The psi_canonicity_checks are enforced if and only if `h_1` = 1.
```

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -0,0 +1,1499 @@
...
+            // b3_c_prime = b_3 + (2^4)c + 2^140 - t_P
+            let b3_c_prime_check = b_3.clone() + (c.clone() * two_pow_4) + two_pow_140.clone()
+                - t_p.clone()
+                - b3_c_prime;
+
+            // e1_f_prime = e_1 + (2^4)f + 2^140 - t_P
+            let e1_f_prime_check =
+                e_1 + (f * two_pow_4) + two_pow_140.clone() - t_p.clone() - e1_f_prime;
+
+            // g1_g2_prime = g_1 + (2^9)g_2 + 2^140 - t_P
```
str4d, line 231: This can actually be:
```suggestion
            // g1_g2_prime = g_1 + (2^9)g_2 + 2^130 - t_P
```

Unlike the other shifted decompositions (which use multiple message pieces), here we are using a subset of a single message piece, with `g_0` being a 1-bit chunk. Thus constraining `z13_g = 0` means that we've constrained `g_1 + (2^9)g_2 < 2^129 < 2^130`, so we can shorted the `g1_g2_prime` decomposition from 14 words to 13.

src/circuit/gadget/utilities/lookup_range_check.rs
therealyingtong: I guess we don't yet know if 2-, 3-, or 4-column space would be more useful for future applications. For now, I'm in favour of keeping it to a single column.

src/circuit.rs
therealyingtong: Opened https://github.com/zcash/orchard/issues/158 for this.

src/circuit/gadget/sinsemilla/commit_ivk.rs
```diff
@@ -0,0 +1,822 @@
...
+                // output by the 10-bit Sinsemilla decomposition of `c`.
+                let z13_c_check = {
+                    let z13_c = meta.query_advice(config.advices[6], Rotation::cur());
+                    d_1.clone() * z13_c
+                };
+
+                // Check that b2_c_prime = b_2 + c * 2^5 + 2^140 - t_P.
+                // This is checked regardless of the value of d_1.
+                let b2_c_prime_check = {
+                    let two_pow_5 = pallas::Base::from_u64(5);
```
str4d, line 184:
```suggestion
                    let two_pow_5 = pallas::Base::from_u64(1 << 5);
```

src/circuit/gadget/sinsemilla/commit_ivk.rs
```diff
@@ -0,0 +1,822 @@
...
+        //      - 0 ≤ b_2 + c * 2^5 < 2^140
+        //          - b_2 was constrained to be 5 bits.
+        //          - z_13 of SinsemillaHash(c) constrains bits 5..=134 to 130 bits
+        //          - so b_2 + c * 2^5 is constrained to be 135 bits < 2^140.
+        //      - 0 ≤ b_2 + c * 2^5 + 2^140 - t_P < 2^140 (14 ten-bit lookups)
+
+        // Decompose the low 140 bits of b2_c_prime = b_2 + c * 2^5 + 2^140 - t_P, and output
+        // the running sum at the end of it. If b2_c_prime < 2^140, the running sum will be 0.
+        let b2_c_prime = b_2.value().zip(c.value()).map(|(b_2, c)| {
+            let two_pow_5 = pallas::Base::from_u64(5);
```
str4d, line 435:
```suggestion
            let two_pow_5 = pallas::Base::from_u64(1 << 5);
```

str4d (changes requested):
Commit<sup>ivk</sup> constraint bug.

I spotted this because I was working on dev tooling that printed out all the gates for performance-optimising purposes, was spot-checking that my output was correct, and realised that I was seeing `0x5` as a constant in a constraint that was supposed to be a binary decomposition (and thus should only contain powers-of-2 constants).

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -0,0 +1,1579 @@
...
+        let two_pow_8 = two_pow_4.square();
+        let two_pow_9 = two_pow_8 * two;
+        let two_pow_10 = two_pow_9 * two;
+        let two_pow_254 = pallas::Base::from_u128(1 << 127).square();
+        let two_pow_130 = Expression::Constant(pallas::Base::from_u128(1 << 65).square());
+        let two_pow_140 = Expression::Constant(pallas::Base::from_u128(1 << 70).square());
+        let t_p = Expression::Constant(pallas::Base::from_u128(T_P));
+
+        // Check that y = lsb + 2 * y_hi.
+        meta.create_gate("y decomposition check", |meta| {
```
str4d, line 86: I'm pretty sure we also need to check the canonicity here, or else someone could witness a non-canonical `y_hi` in order to flip the `lsb` (since `p` is odd, adding it to a canonical scalar will result in an equivalent but non-canonical scalar with the opposite parity).

str4d (changes requested):

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -0,0 +1,1721 @@
...
+            //                && e1_f_prime_decomposition = 0
+            //      - j_1 = 0 => j_0 = 0
+            //                && g1_g2_prime_decomposition = 0
+
+            let q_canon_2 = meta.query_selector(config.q_canon_2);
+
+            // Offset prev
+            let e_1 = meta.query_advice(config.advices[5], Rotation::prev());
+            let f = meta.query_advice(config.advices[6], Rotation::prev());
+            let g_1 = meta.query_advice(config.advices[8], Rotation::prev());
```
str4d, line 387: We can avoid these rotations entirely if we swap these three cells with the three cells at `Rotation::cur()` that are unused in this gate.

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -0,0 +1,1721 @@
...
+
+                |         A_0         |          A_1           |         A_2            |          A_3            |  A_4  |  A_5  |  A_6  |  A_7   |  A_8  |   A_9   |  q_canon_1  |  q_canon_2  |
+                ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+                |       a_prime       |       b3_c_prime       |       e1_f_prime       |       g1_g2_prime       |   a   |   b   |  b_2  |   b_3  |   c   |    d    |      0      |      0      |
+                |         d_1         |           d_2          |          z1_d          |           e             |  e_0  |  e_1  |   f   |    g   |  g_1  |    h    |      1      |      0      |
+                |         h_0         |           h_1          |         x(g_d)         |        x(pk_d)          | value |  b_0  |  b_1  |   d_0  |  g_0  |   z1_g  |      0      |      1      |
+                |a_prime_decomposition|b3_c_prime_decomposition|e1_f_prime_decomposition|g1_g2_prime_decomposition| z13_a | z13_c | z13_f | z13_g  |  psi  |   rho   |      0      |      0      |
```
str4d, line 364: The simplest change to remove those rotations would be:
```suggestion
                |         A_0         |          A_1           |         A_2            |          A_3            |  A_4  |  A_5  |  A_6  |  A_7   |  A_8  |   A_9   |  q_canon_1  |  q_canon_2  |
                ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
                |       a_prime       |       b3_c_prime       |       e1_f_prime       |       g1_g2_prime       |   a   |   b   |  b_2  |   b_3  |   c   |    d    |      0      |      0      |
                |         d_1         |           d_2          |          z1_d          |           e             |  e_0  |x(g_d) |x(pk_d)|    g   | value |    h    |      1      |      0      |
                |         h_0         |           h_1          |          e_1           |           f             |  g_1  |  b_0  |  b_1  |   d_0  |  g_0  |   z1_g  |      0      |      1      |
                |a_prime_decomposition|b3_c_prime_decomposition|e1_f_prime_decomposition|g1_g2_prime_decomposition| z13_a | z13_c | z13_f | z13_g  |  psi  |   rho   |      0      |      0      |
```
(with equivalent changes to the first gate).

I'll have a think about whether I'd prefer an alternate region layout that more clearly conveys the data dependencies, but this change would suffice for now.

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -0,0 +1,1721 @@
...
+            //   - canonicity:
+            //      - b_1 = 0 => b_0 = 0
+            //                && z13_a = 0
+            //                && a_prime_decomposition = 0
+            //      - d_0 = 0 => z13_c = 0
+            //                && b3_c_prime_decomposition = 0
+            //      - h_0 = 0 => z13_f = 0
+            //                && e1_f_prime_decomposition = 0
+            //      - j_1 = 0 => j_0 = 0
+            //                && g1_g2_prime_decomposition = 0
```
str4d, line 380:
```suggestion
            // q_canon_2 checks that:
            //   - field element decomposition:
            //      - rho = e_1 + (2^4) f + (2^254) g_0
            //      - psi = g_1 + (2^9) g_2 + (2^249) h_0 + (2^254) h_1
            //   - canonicity:
            //      - b_1 = 1 => b_0 = 0
            //                && z13_a = 0
            //                && a_prime_decomposition = 0
            //      - d_0 = 1 => z13_c = 0
            //                && b3_c_prime_decomposition = 0
            //      - g_0 = 1 => z13_f = 0
            //                && e1_f_prime_decomposition = 0
            //      - h_1 = 1 => h_0 = 0
            //                && g1_g2_prime_decomposition = 0
```

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -0,0 +1,1721 @@
...
+            let j = lsb.zip(k_0.value()).zip(k_1).map(|((lsb, k_0), k_1)| {
+                let two = pallas::Base::from_u64(2);
+                let two_pow_10 = pallas::Base::from_u64(1 << 10);
+                lsb + two * k_0 + two_pow_10 * k_1
+            });
+            let zs = self.sinsemilla_config.lookup_config.witness_check(
+                layouter.namespace(|| "Decompose j = LSB + (2)k_0 + (2^10)k_1"),
+                j,
+                25,
+                false,
```
therealyingtong, line 964: I think this has to be a strict decomposition (since it is replacing the decomposition usually done by the SinsemillaHash).

```suggestion
                true,
```

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -0,0 +1,1721 @@
...
+            )?;
+            (zs[0], zs[1], zs[13])
+        };
+
+        // Decompose j_prime = j + 2^130 - t_P using 25 ten-bit lookups.
```
therealyingtong, line 969:
```suggestion
        // Decompose j_prime = j + 2^130 - t_P using 13 ten-bit lookups.
```

str4d (commented):
Flushing comments.

src/circuit/gadget/sinsemilla/note_commit.rs
str4d: I've figured out an alternate region layout that keeps almost all cells close to the ones they are involved in (which would make visualising constraints easier), while also keeping the above rotation reduction:
```suggestion
                |   A_0    |    A_1    |     A_2      | A_3 |  A_4  |       A_5     |    A_6   |     A_7      |  A_8  |    A_9    |  q_canon_1  |  q_canon_2  |
                -----------------------------------------------------------------------------------------------------------------------------------------------
                |    b     |     d     |      e       |  g  |   h   |       d_1     |  x(pk_d) |     b_3      |a_prime|    b_2    |      0      |      0      |
                |e1_f_prime|g1_g2_prime|    value     | d_2 |  z1_d |       e_0     |b3_c_prime|      c       |   a   |  x(g_d)   |      1      |      0      |
                |   e_1    |     f     |     g_0      | g_1 |  z1_g |       h_0     |    h_1   |     d_0      |  b_0  |    b_1    |      0      |      1      |
                |   rho    |   z13_f   |z14_e1_f_prime| psi | z13_g |z14_g1_g2_prime|  z13_c   |z14_b3_c_prime| z13_a |z13_a_prime|      0      |      0      |
```

src/circuit/gadget/sinsemilla/note_commit.rs
therealyingtong: Actually I think we could just do a 13-word lookup decomposition here, and redistribute the pieces as 
```
(bit 0) || (bits 1..=9) || (bits 10..=129) || (bits 130..=253) || bit 254
```

src/circuit.rs
```diff
@@ -2,61 +2,759 @@
...
+        mut layouter: impl Layouter<pallas::Base>,
     ) -> Result<(), plonk::Error> {
+        // Load the Sinsemilla generator lookup table used by the whole circuit.
+        SinsemillaChip::load(config.sinsemilla_config_1.clone(), &mut layouter)?;
+
+        // Construct the ECC chip.
+        let ecc_chip = config.ecc_chip();
+
+        // Witness private inputs that are used across multiple checks.
+        let (rho_old, psi_old, cm_old, g_d_old, ak, nk, v_old, v_new) = {
```
daira, line 332:
```suggestion
        let (psi_old, rho_old, cm_old, g_d_old, ak, nk, v_old, v_new) = {
```
Just putting the tuple in the same order as the values are witnessed.

src/circuit.rs
```diff
@@ -2,61 +2,759 @@
...
+
+            // Witness v_new.
+            let v_new = self.load_private(
+                layouter.namespace(|| "witness v_new"),
+                config.advices[0],
+                self.v_new
+                    .map(|v_new| pallas::Base::from_u64(v_new.inner())),
+            )?;
+
+            (rho_old, psi_old, cm_old, g_d_old, ak, nk, v_old, v_new)
```
daira, line 392:
```suggestion
            (psi_old, rho_old, cm_old, g_d_old, ak, nk, v_old, v_new)
```

src/circuit.rs
```diff
@@ -2,61 +2,759 @@
...
 
-        meta.create_gate("TODO", |meta| {
-            let a = meta.query_advice(advice, Rotation::cur());
-            let s = meta.query_selector(selector);
+        // Addition of two field elements poseidon_hash(nk, rho_old) + psi_old.
+        let q_add = meta.selector();
+        meta.create_gate("poseidon_hash(nk, rho_old) + psi_old", |meta| {
+            let q_add = meta.query_selector(q_add);
+            let sum = meta.query_advice(advices[0], Rotation::cur());
+            let nk_rho_old = meta.query_advice(advices[1], Rotation::cur());
```
daira, line 191:
```suggestion
            let hash_old = meta.query_advice(advices[1], Rotation::cur());
```

src/circuit.rs
```diff
@@ -2,61 +2,759 @@
...
+            let psi_old = meta.query_advice(advices[2], Rotation::cur());
 
-            vec![s * a]
+            vec![q_add * (nk_rho_old + psi_old - sum)]
```
daira, line 194:
```suggestion
            vec![q_add * (hash_old + psi_old - sum)]
```

src/circuit.rs
```diff
@@ -2,61 +2,759 @@
...
+            layouter.constrain_instance(cv_net.inner().x().cell(), config.primary, CV_NET_X)?;
+            layouter.constrain_instance(cv_net.inner().y().cell(), config.primary, CV_NET_Y)?;
+
+            v_net
+        };
+
+        // Nullifier integrity
+        let nf_old = {
+            // nk_rho_old = poseidon_hash(nk, rho_old)
+            let nk_rho_old = {
```
daira, line 479:
```suggestion
            let hash_old = {
```

src/circuit.rs
```diff
@@ -2,61 +2,759 @@
...
+                || " `scalar` = poseidon_hash(nk, rho_old) + psi_old",
+                |mut region| {
+                    config.q_add.enable(&mut region, 0)?;
+
+                    copy(
+                        &mut region,
+                        || "copy nk_rho_old",
+                        config.advices[1],
+                        0,
+                        &nk_rho_old,
```
daira, line 529:
```suggestion
                        || "copy hash_old",
                        config.advices[1],
                        0,
                        &hash_old,
```

src/circuit.rs
```diff
@@ -2,61 +2,759 @@
...
+                        || "copy psi_old",
+                        config.advices[2],
+                        0,
+                        &psi_old,
+                    )?;
+
+                    let scalar_val = nk_rho_old
+                        .value()
+                        .zip(psi_old.value())
+                        .map(|(nk_rho_old, psi_old)| nk_rho_old + psi_old);
```
daira, line 542:
```suggestion
                    let scalar_val = hash_old
                        .value()
                        .zip(psi_old.value())
                        .map(|(hash_old, psi_old)| hash_old + psi_old);
```

src/circuit.rs
```diff
@@ -2,61 +2,759 @@
...
+            };
+
+            // [ivk] g_d_old
+            // The scalar value is passed through and discarded.
+            let (derived_pk_d_old, _ivk) =
+                g_d_old.mul(layouter.namespace(|| "[ivk] g_d_old"), ivk.inner())?;
+
+            // Constrain derived pk_d_old to equal witnessed pk_d_old
+            // This addresses the case where ivk = ⊥ , since ⊥ maps to 0 and variable-base
+            // scalar multiplication maps [0]B to (0, 0).
```
daira, line 617:
```suggestion
```
The comment is maybe a bit misleading, since ivk = ⊥ in the spec is modelling the case where Commit<sup>ivk</sup> has hit an exceptional case, and so we don't know what the circuit will actually compute for ivk (we're relying on this happening with negligible probability).

src/circuit.rs
```diff
@@ -2,61 +2,759 @@
...
+            ])
+            .map(move |(name, poly)| (name, q_orchard.clone() * poly))
+        });
 
-        meta.create_gate("TODO", |meta| {
-            let a = meta.query_advice(advice, Rotation::cur());
-            let s = meta.query_selector(selector);
+        // Addition of two field elements poseidon_hash(nk, rho_old) + psi_old.
+        let q_add = meta.selector();
+        meta.create_gate("poseidon_hash(nk, rho_old) + psi_old", |meta| {
```
therealyingtong, line 188: We could merge this into the same gate as `q_orchard`.

daira (approved):
utACK for the QEDIT review.

I want to go back and do a more thorough review of the canonicity checks, but that shouldn't block merging.

str4d (changes requested):
One blocking comment, otherwise utACK with various comments and suggestions.

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -0,0 +1,1721 @@
...
+        */
+        meta.create_gate("y coordinate checks", |meta| {
+            let q_y_canon = meta.query_selector(q_y_canon);
+            let y = meta.query_advice(advices[0], Rotation::cur());
+            // LSB has been boolean-constrained outside this gate.
+            let lsb = meta.query_advice(advices[1], Rotation::cur());
+            // k_0 has been constrained to 9 bits outside this gate.
+            let k_0 = meta.query_advice(advices[2], Rotation::cur());
+            // This gate constrains k_1 to 240 bits by equating it to z1_j.
+            let k_1 = meta.query_advice(advices[3], Rotation::cur());
```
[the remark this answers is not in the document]
therealyingtong: This change has been made in e4a960d7f1bafaf4e87bb47d95c8cd90ef3252d5 and documented in 920fe643997ac5887f4f1a6e5d9e0b8bc4b8af30.

src/circuit/gadget/sinsemilla/note_commit.rs
therealyingtong: Small change needed to this comment: `z14_g1_g2_prime` -> `z13_g1_g2_prime`.

Otherwise, this has been implemented at  7aa3174880cb7fe26b1e594057675b4d7ff49087 and documented at ac5404a94317623ae713d6f0408d37247d46ea10.

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -0,0 +1,1721 @@
...
+            MessagePiece::from_field_elem(chip.clone(), layouter.namespace(|| "a"), a, 25)?
+        };
+
+        // b = b_0 || b_1 || b_2 || b_3
+        //   = (bits 250..=253 of x(g_d)) || (bit 254 of x(g_d)) || (ỹ bit of g_d) || (bits 0..=3 of pk★_d)
+        let (b_0, b_1, b_2, b_3, b) =
+            {
+                let b_0 = gd_x.map(|gd_x| bitrange_subset(gd_x, 250..254));
+                let b_1 = gd_x.map(|gd_x| bitrange_subset(gd_x, 254..255));
+                let b_2 = gd_y.map(|gd_y| bitrange_subset(gd_y, 0..1));
```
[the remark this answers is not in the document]
therealyingtong: I have removed `point_repr()`.

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -1590,9 +1581,12 @@ mod tests {
 
                 // Witness g_d
                 let g_d = {
-                    let g_d = self.gd_x.map(|x| {
+                    let g_d = self.gd_x.zip(self.gd_y_lsb).map(|(x, y_lsb)| {
                         // Calculate y = (x^3 + 5).sqrt()
-                        let y = (x.square() * x + pallas::Affine::b()).sqrt().unwrap();
+                        let mut y = (x.square() * x + pallas::Affine::b()).sqrt().unwrap();
+                        if y.is_odd() && y_lsb == pallas::Base::zero() {
```
str4d, line 1587: This only covers half of the required sign flips. Easiest way to cover both is with an XOR:
```suggestion
                        if y.is_odd() ^ y_lsb.is_odd() {
```

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -1601,9 +1595,12 @@ mod tests {
 
                 // Witness pk_d
                 let pk_d = {
-                    let pk_d = self.pkd_x.map(|x| {
+                    let pk_d = self.pkd_x.zip(self.pkd_y_lsb).map(|(x, y_lsb)| {
                         // Calculate y = (x^3 + 5).sqrt()
-                        let y = (x.square() * x + pallas::Affine::b()).sqrt().unwrap();
+                        let mut y = (x.square() * x + pallas::Affine::b()).sqrt().unwrap();
+                        if y.is_odd() && y_lsb == pallas::Base::zero() {
```
str4d, line 1601:
```suggestion
                        if y.is_odd() ^ y_lsb.is_odd() {
```

str4d (approved):
utACK modulo test fixes. Circuit layout LGTM.

action-circuit-layout

src/circuit/gadget/sinsemilla/note_commit.rs
daira: If bits 130..=253 are not range-constrained, how do you know that bits 0..=254 are actually a 10-bit-word decomposition of y(g<sub>d</sub>) or y(pk<sub>d</sub>)?

src/circuit/gadget/sinsemilla/note_commit.rs
```diff
@@ -0,0 +1,1721 @@
...
+                    g.clone(),
+                    h.clone(),
+                ],
+            );
+            let domain = CommitDomain::new(chip, ecc_chip, &SinsemillaCommitDomains::NoteCommit);
+            domain.commit(
+                layouter.namespace(|| "Process NoteCommit inputs"),
+                message,
+                rcm,
+            )?
```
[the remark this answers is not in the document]
daira: Yeah, I think it would be possible to automate this kind of processing. That's clearly necessary to make the Sinsemilla gadget accessible to a wider user base.

We also probably didn't do ourselves any favours by specifying bit-packed inputs to Sinsemilla, rather than a more regular input arrangement of FE(x(g<sub>d</sub>)) || FE(x(pk<sub>d</sub>) || FE(ρ) || FE(ψ) || FE(v + 2<sup>64</sup> ỹ(g<sub>d</sub>) + 2<sup>65</sup> ỹ(pk<sub>d</sub>)).

PR #105, opened 2021-06-05
Add SpendAuthG fixed base

therealyingtong:
Used in spend authority randomization where `rk = ak + [alpha]SpendAuthG` (cf. https://zips.z.cash/protocol/nu5.pdf#concretespendauthsig).

image


daira (approved):
ACK

src/constants/spend_auth_g.rs
```diff
@@ -0,0 +1,2963 @@
+use super::{OrchardFixedBase, SpendAuthG};
+use halo2::arithmetic::{CurveAffine, FieldExt};
+
+/// The value commitment is used to check balance between inputs and outputs. The value is
+/// placed over this generator.
+pub const GENERATOR: ([u8; 32], [u8; 32]) = (
```
str4d, line 6: I checked this constant against `orchard_generators.py` from the test vectors.

str4d (approved):
ACK

Issue #102, opened 2021-06-04
Builder: Re-randomize signatures


str4d:
The builder currently uses `ask` directly to sign inputs. It should instead be re-randomizing `ask` with the per-input `alpha` to get `rsk`, and then signing with `rsk`.

PR #108, opened 2021-06-05
builder: Store alpha and use it to derive rsk for signing spends

str4d:
This was missed from zcash/orchard#49, but could not have caused a consensus failure, loss-of-funds, or privacy violation because `alpha` _was_ being sampled and used to derive `rk`, meaning that the signatures would fail to validate.

Closes zcash/orchard#102.


daira (approved):
utACK

PR #110, opened 2021-06-08
Implement tree logic in Builder

therealyingtong:


src/tree.rs
```diff
@@ -101,3 +101,138 @@ fn hash_layer(l_star: usize, pair: Pair) -> pallas::Base {
...
+                            .to_le_bits()
+                            .iter()
+                            .by_val()
+                            .take(pallas::Base::NUM_BITS as usize)
+                        ),
+                )
+                .unwrap();
+                v.push(next);
+            }
+            v
```
str4d, line 154: This can just use `hash_layer`. It could also be implemented with a scan:
```suggestion
            Some(pallas::Base::zero())
                .chain((0..MERKLE_DEPTH_ORCHARD).scan(pallas::Base::zero(), |state, l_star| {
                    *state = hash_layer(l_star, Pair { left: *state, right: *state });
                    Some(*state)
                })
                .collect()
```

src/tree.rs
```diff
@@ -101,3 +101,138 @@ fn hash_layer(l_star: usize, pair: Pair) -> pallas::Base {
...
+            let padded_leaves = {
+                let mut padded_leaves = commitments.clone();
+
+                let pad = (0..((1 << perfect_subtree_depth) - n_notes as usize)).map(
+                    |_| ExtractedNoteCommitment::uncommitted()
+                ).collect::<Vec<_>>();
+
+                padded_leaves.extend_from_slice(&pad);
+                padded_leaves
+            };
```
str4d, line 185: Instead of representing the perfect tree with uncommitted leaves (since that API will be removed), use `Option<pallas::Base>`.

src/tree.rs
```diff
@@ -101,3 +101,138 @@ fn hash_layer(l_star: usize, pair: Pair) -> pallas::Base {
...
+                    let inner_nodes = (0..(perfect_subtree_depth >> height)).map(|pos| {
+                        hash_layer(height, Pair {
+                            left: tree[height - 1][pos * 2],
+                            right: tree[height - 1][pos * 2 + 1],
+                        })
+                    }).collect();
+                    tree.push(inner_nodes);
+                };
+                tree
+            };
```
str4d, line 199: Then in here:
- When both left and right are `None`, push `None` (skipping the `hash_layer`, which was already pre-computed in `EMPTY_ROOTS`).
- When either one left or right is `None`, fetch the corresponding value from `EMPTY_ROOTS`.

In the auth path logic below, values in `tree` can just be unwrapped, since all required siblings in the tree will have been filled in.

str4d (changes requested):

src/tree.rs
```diff
@@ -73,6 +81,13 @@ fn cond_swap(swap: bool, node: pallas::Base, sibling: pallas::Base) -> Pair {
 }
 
 // <https://zips.z.cash/protocol/protocol.pdf#orchardmerklecrh>
+// The layer with 2^n nodes is called "layer n":
+//      - leaves are at layer MERKLE_DEPTH_ORCHARD = 32;
+//      - the root is at layer 0.
+// `l_star` is MERKLE_DEPTH_ORCHARD - layer - 1.
+//      - when hashing two leaves, we produce a node on the layer above the leaves, i.e.
+//        layer = 31, l_star = 0
+//      - when hashing to the final root, we produce the anchor with layer = 0, l_star = 31.
```
str4d: Make these doc comments (`///`), as this can then be rendered with `cargo doc --document-private-items`.

str4d (approved):
utACK with suggestions.

src/primitives/sinsemilla.rs
```diff
@@ -20,6 +21,15 @@ fn lebs2ip_k(bits: &[bool]) -> u32 {
         .fold(0u32, |acc, (i, b)| acc + if *b { 1 << i } else { 0 })
 }
 
+/// The sequence of K bits in little-endian order representing an integer
+/// up to `2^K` - 1.
+pub fn i2lebsp_k(int: usize) -> [bool; K] {
+    assert!(int < (1 << K));
+    gen_const_array(int, |int: &mut usize, mask: usize| {
+        ((*int & (1 << mask)) >> mask) == 1
```
nuttycom, line 29: This looks to me like it's the same as `(*int & (1 << mask)) != 0` - is this version more performant?

src/tree.rs
```diff
@@ -1,15 +1,298 @@
...
+            position: rng.next_u32(),
+            auth_path: gen_const_array(rng, dummy_inner),
+        }
+    }
+
+    /// <https://zips.z.cash/protocol/protocol.pdf#orchardmerklecrh>
+    /// The layer with 2^n nodes is called "layer n":
+    ///      - leaves are at layer MERKLE_DEPTH_ORCHARD = 32;
+    ///      - the root is at layer 0.
+    /// `l_star` is MERKLE_DEPTH_ORCHARD - layer - 1.
```
nuttycom, line 47: This is what I called "level" in https://github.com/ebfull/incrementalmerkletree/pull/1

nuttycom (approved):
utACK. This looks great!

src/tree.rs
```diff
@@ -1,15 +1,298 @@
...
+        (
+            // generate note values that we're certain won't exceed MAX_NOTE_VALUE in total
+            notes in vec(
+                arb_positive_note_value(MAX_NOTE_VALUE / n_notes as u64).prop_flat_map(arb_note),
+                n_notes
+            ),
+        )
+        -> (Vec<(Note, MerklePath)>, Anchor) {
+            // Inefficient algorithm to build a perfect subtree containing all notes.
+            let perfect_subtree_depth = (n_notes as f64).log2().ceil() as usize;
```
nuttycom, line 190: Does the `as` cast work properly here, or should `into()` be used?

src/primitives/sinsemilla.rs
str4d:
```rs
fn bar(a: usize, mask: usize) -> bool {
    ((a & (1 << mask)) >> mask) == 1
}

fn foo(a: usize, mask: usize) -> bool {
    (a & (1 << mask)) != 0
}
```
```
playground::bar:
	movq	%rsi, %rcx
	movl	$1, %eax
	shlq	%cl, %rax
	andq	%rdi, %rax
	shrq	%cl, %rax
	cmpq	$1, %rax
	sete	%al
	retq

playground::foo:
	btq	%rsi, %rdi
	setb	%al
	retq
```
It looks like the `!= 0` test enables LLVM to figure out this is bit-testing, and swap in that opcode. So that is the faster operation to use.

str4d: Made this change alongside my suggestions.

src/tree.rs
```diff
@@ -1,15 +1,294 @@
...
+}
+
+// <https://zips.z.cash/protocol/protocol.pdf#orchardmerklecrh>
+// The layer with 2^n nodes is called "layer n":
+//      - leaves are at layer MERKLE_DEPTH_ORCHARD = 32;
+//      - the root is at layer 0.
+// `l_star` is MERKLE_DEPTH_ORCHARD - layer - 1.
+//      - when hashing two leaves, we produce a node on the layer above the leaves, i.e.
+//        layer = 31, l_star = 0
+//      - when hashing to the final root, we produce the anchor with layer = 0, l_star = 31.
```
str4d, line 96:
```suggestion
/// <https://zips.z.cash/protocol/protocol.pdf#orchardmerklecrh>
/// The layer with 2^n nodes is called "layer n":
///      - leaves are at layer MERKLE_DEPTH_ORCHARD = 32;
///      - the root is at layer 0.
/// `l_star` is MERKLE_DEPTH_ORCHARD - layer - 1.
///      - when hashing two leaves, we produce a node on the layer above the leaves, i.e.
///        layer = 31, l_star = 0
///      - when hashing to the final root, we produce the anchor with layer = 0, l_star = 31.
```

src/tree.rs
str4d: Per https://doc.rust-lang.org/nomicon/casts.html:
> - casting from an integer to float will produce the floating point representation of the integer, rounded if necessary (rounding to nearest, ties to even)

I think this is fine; we cannot have more notes than `u32` given the depth of the tree, so that will fit into an `f64`.

str4d: However, we could also use:
```suggestion
            let perfect_subtree_depth = n_notes.next_power_of_two();
```

nuttycom: That's not the same thing; you want the log2 of the number of notes.

str4d: Aha yep. Whoops!

src/tree.rs
```diff
@@ -1,15 +1,103 @@
...
+use pasta_curves::{arithmetic::FieldExt, pallas};
+
+use ff::{PrimeField, PrimeFieldBits};
 use rand::RngCore;
+use std::{convert::TryInto, iter};
 
 /// The root of an Orchard commitment tree.
-#[derive(Clone, Debug)]
+#[derive(Eq, PartialEq, Clone, Debug)]
 pub struct Anchor(pub [u8; 32]);
```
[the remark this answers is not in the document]
daira: Normally we implement type constraints as non-contextual checks. I would do so here, even if it would also be rejected later by the contextual check against a previous anchor.

PR #107, opened 2021-06-05
[ECC chip] Incomplete and complete addition

therealyingtong:
The variable- and fixed-base scalar multiplication instructions are implemented in #111.

PR #111, opened 2021-06-10
[ECC chip] Fixed- and variable-base scalar multiplication

### LINKED ISSUE #38: Implement generic ECC gadget

therealyingtong:
Closes #38.

### DIFF
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -246,7 +246,7 @@ impl EccChip {
             short_config.create_gate(meta);
         }
 
-        // Create gate ths is only used in fixed-base mul using a base field element.
+        // Create gate that is only used in fixed-base mul using a base field element.
         {
             let base_field_config: mul_fixed::base_field_elem::Config = (&config).into();
             base_field_config.create_gate(meta);
@@ -300,7 +300,8 @@ pub struct EccScalarFixedShort {
 /// Each `a_i` is in the range [0..2^3).
 ///
 /// `windows` = [z_1, ..., z_85], where we expect z_85 = 0.
-/// Since z_0 is initialized as α, we store it as `base_field_elem`.
+/// Since z_0 is initialized as the scalar α, we store it as
+/// `base_field_elem`.
 #[derive(Clone, Debug)]
 struct EccBaseFieldElemFixed {
     base_field_elem: CellValue<pallas::Base>,
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -88,9 +88,9 @@ pub struct EccConfig {
     pub q_add: Selector,
 
     /// Variable-base scalar multiplication (hi half)
-    pub q_mul_hi: Selector,
+    pub q_mul_hi: Column<Fixed>,
     /// Variable-base scalar multiplication (lo half)
-    pub q_mul_lo: Selector,
+    pub q_mul_lo: Column<Fixed>,
     /// Selector used to enforce boolean decomposition in variable-base scalar mul
     pub q_mul_decompose_var: Selector,
     /// Selector used to enforce switching logic on LSB in variable-base scalar mul
@@ -179,8 +179,8 @@ impl EccChip {
             fixed_z: meta.fixed_column(),
             q_add_incomplete: meta.selector(),
             q_add: meta.selector(),
-            q_mul_hi: meta.selector(),
-            q_mul_lo: meta.selector(),
+            q_mul_hi: meta.fixed_column(),
+            q_mul_lo: meta.fixed_column(),
             q_mul_decompose_var: meta.selector(),
             q_mul_overflow: meta.selector(),
             q_mul_lsb: meta.selector(),
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -449,9 +449,10 @@ impl EccInstructions<pallas::Affine> for EccChip {
     ) -> Result<Self::Point, Error> {
         let config: mul_fixed::full_width::Config<{ constants::NUM_WINDOWS }> =
             self.config().into();
-        layouter.assign_region(
-            || format!("fixed-base mul of {:?}", base),
-            |mut region| config.assign_region(scalar, *base, 0, &mut region),
+        config.assign(
+            layouter.namespace(|| format!("fixed-base mul of {:?}", base)),
+            scalar,
+            *base,
         )
     }
 
@@ -463,9 +464,10 @@ impl EccInstructions<pallas::Affine> for EccChip {
     ) -> Result<Self::Point, Error> {
         let config: mul_fixed::short::Config<{ constants::NUM_WINDOWS_SHORT }> =
             self.config().into();
-        layouter.assign_region(
-            || format!("short fixed-base mul of {:?}", base),
-            |mut region| config.assign_region(scalar, base, 0, &mut region),
+        config.assign(
+            layouter.namespace(|| format!("short fixed-base mul of {:?}", base)),
+            scalar,
+            base,
         )
     }
 
@@ -476,9 +478,10 @@ impl EccInstructions<pallas::Affine> for EccChip {
         base: &Self::FixedPoints,
     ) -> Result<Self::Point, Error> {
         let config: mul_fixed::base_field_elem::Config = self.config().into();
-        layouter.assign_region(
-            || format!("base field elem fixed-base mul of {:?}", base),
-            |mut region| config.assign_region(base_field_elem, *base, 0, &mut region),
+        config.assign(
+            layouter.namespace(|| format!("base-field elem fixed-base mul of {:?}", base)),
+            base_field_elem,
+            *base,
         )
     }
 }
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -12,7 +12,7 @@ use pasta_curves::{arithmetic::CurveAffine, pallas};
 
 pub(super) mod add;
 pub(super) mod add_incomplete;
-// pub(super) mod mul;
+pub(super) mod mul;
 // pub(super) mod mul_fixed;
 pub(super) mod witness_point;
 // pub(super) mod witness_scalar_fixed;
@@ -86,6 +86,8 @@ pub struct EccConfig {
     pub q_mul_lo: Selector,
     /// Selector used in scalar decomposition for variable-base scalar mul
     pub q_mul_decompose_var: Selector,
+    /// Selector used in scalar decomposition for variable-base scalar mul
+    pub q_init_z: Selector,
     /// Variable-base scalar multiplication (final scalar)
     pub q_mul_complete: Selector,
     /// Fixed-base full-width scalar multiplication
@@ -150,6 +152,7 @@ impl EccChip {
             q_mul_hi: meta.selector(),
             q_mul_lo: meta.selector(),
             q_mul_decompose_var: meta.selector(),
+            q_init_z: meta.selector(),
             q_mul_complete: meta.selector(),
             q_mul_fixed: meta.selector(),
             q_mul_fixed_short: meta.selector(),
@@ -175,6 +178,12 @@ impl EccChip {
         {
             let add_config: add::Config = (&config).into();
             add_config.create_gate(meta);
+        }
+
+        // Create variable-base scalar mul gates
+        {
+            let mul_config: mul::Config = (&config).into();
+            mul_config.create_gate(meta);
         }
 
         config
@@ -311,11 +320,15 @@ impl EccInstructions<pallas::Affine> for EccChip {
 
     fn mul(
         &self,
-        _layouter: &mut impl Layouter<pallas::Base>,
-        _scalar: &Self::ScalarVar,
-        _base: &Self::Point,
+        layouter: &mut impl Layouter<pallas::Base>,
+        scalar: &Self::ScalarVar,
+        base: &Self::Point,
     ) -> Result<Self::Point, Error> {
-        todo!()
+        let config: mul::Config = self.config().into();
+        layouter.assign_region(
+            || "variable-base scalar mul",
+            |mut region| config.assign_region(scalar, base, 0, &mut region),
+        )
     }
 
     fn mul_fixed(
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -284,8 +284,12 @@ pub struct EccScalarFixed {
 /// A signed short scalar used for fixed-base scalar multiplication.
 /// A short scalar must have magnitude in the range [0..2^64), with
 /// a sign of either 1 or -1.
-/// This is decomposed into 22 3-bit windows in little-endian order,
-/// i.e. `windows` = [k_0, k_1, ..., k_21] (for a 64-bit magnitude)
+/// This is decomposed into 3-bit windows in little-endian order
+/// using a running sum `z`, where z_{i+1} = (z_i - a_i) / (2^3)
+/// for element α = a_0 + (2^3) a_1 + ... + (2^{3(n-1)}) a_{n-1}.
+/// Each `a_i` is in the range [0..2^3).
+///
+/// `windows` = [k_0, k_1, ..., k_21] (for a 64-bit magnitude)
 /// where `scalar = k_0 + k_1 * (2^3) + ... + k_84 * (2^3)^84` and
 /// each `k_i` is in the range [0..2^3).
 /// k_21 must be a single bit, i.e. 0 or 1.
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -91,12 +91,12 @@ pub struct EccConfig {
     pub q_mul_hi: Selector,
     /// Variable-base scalar multiplication (lo half)
     pub q_mul_lo: Selector,
-    /// Selector used in scalar decomposition for variable-base scalar mul
-    pub q_mul_decompose_var: Selector,
-    /// Selector used in scalar decomposition for variable-base scalar mul
+    /// Selector used to initialize running sum to zero in variable-base scalar mul
     pub q_init_z: Selector,
-    /// Variable-base scalar multiplication (final scalar)
-    pub q_mul_z: Selector,
+    /// Selector used to enforce boolean decomposition in variable-base scalar mul
+    pub q_mul_decompose_var: Selector,
+    /// Selector used to enforce switching logic on LSB in variable-base scalar mul
+    pub q_mul_lsb: Selector,
     /// Variable-base scalar multiplication (overflow check)
     pub q_mul_overflow: Selector,
 
@@ -171,10 +171,10 @@ impl EccChip {
             q_add: meta.selector(),
             q_mul_hi: meta.selector(),
             q_mul_lo: meta.selector(),
-            q_mul_decompose_var: meta.selector(),
             q_init_z: meta.selector(),
-            q_mul_z: meta.selector(),
+            q_mul_decompose_var: meta.selector(),
             q_mul_overflow: meta.selector(),
+            q_mul_lsb: meta.selector(),
             mul_fixed: meta.fixed_column(),
             q_mul_fixed_short: meta.selector(),
             base_field_fixed: meta.fixed_column(),
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -91,8 +91,6 @@ pub struct EccConfig {
     pub q_mul_hi: Selector,
     /// Variable-base scalar multiplication (lo half)
     pub q_mul_lo: Selector,
-    /// Selector used to initialize running sum to zero in variable-base scalar mul
-    pub q_init_z: Selector,
     /// Selector used to enforce boolean decomposition in variable-base scalar mul
     pub q_mul_decompose_var: Selector,
     /// Selector used to enforce switching logic on LSB in variable-base scalar mul
@@ -109,14 +107,19 @@ pub struct EccConfig {
 
     /// Witness point
     pub q_point: Selector,
+
     /// Witness full-width scalar for fixed-base scalar mul
     pub q_scalar_fixed: Selector,
     /// Witness signed short scalar for full-width fixed-base scalar mul
     pub q_scalar_fixed_short: Selector,
 
-    /// Permutation
+    /// Shared fixed column used for loading constants. This is included in
+    /// the permutation so that cells in advice columns can be constrained to
+    /// equal cells in this fixed column.
+    pub constants: Column<Fixed>,
+    /// Permutation over all advice columns and the `constants` fixed column.
     pub perm: Permutation,
-    /// 10-bit lookup
+    /// 10-bit lookup table
     pub lookup_config: LookupRangeCheckConfig<pallas::Base, { sinsemilla::K }>,
 }
 
@@ -149,10 +152,17 @@ impl EccChip {
         meta: &mut ConstraintSystem<pallas::Base>,
         advices: [Column<Advice>; 10],
         lookup_table: Column<Fixed>,
+        // TODO: Replace with public inputs API
+        constants: [Column<Fixed>; 2],
         perm: Permutation,
     ) -> <Self as Chip<pallas::Base>>::Config {
-        let lookup_config =
-            LookupRangeCheckConfig::configure(meta, advices[9], lookup_table, perm.clone());
+        let lookup_config = LookupRangeCheckConfig::configure(
+            meta,
+            advices[9],
+            constants[0],
+            lookup_table,
+            perm.clone(),
+        );
 
         let config = EccConfig {
             advices,
@@ -171,7 +181,6 @@ impl EccChip {
             q_add: meta.selector(),
             q_mul_hi: meta.selector(),
             q_mul_lo: meta.selector(),
-            q_init_z: meta.selector(),
             q_mul_decompose_var: meta.selector(),
             q_mul_overflow: meta.selector(),
             q_mul_lsb: meta.selector(),
@@ -181,6 +190,7 @@ impl EccChip {
             q_point: meta.selector(),
             q_scalar_fixed: meta.selector(),
             q_scalar_fixed_short: meta.selector(),
+            constants: constants[1],
             perm,
             lookup_config,
         };
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -1,6 +1,7 @@
 use super::EccInstructions;
 use crate::circuit::gadget::utilities::{copy, CellValue, Var};
-use crate::constants;
+use crate::constants::{self, OrchardFixedBasesFull, ValueCommitV};
+use arrayvec::ArrayVec;
 
 use group::prime::PrimeCurveAffine;
 use halo2::{
@@ -180,14 +181,37 @@ impl EccChip {
     }
 }
 
+/// A full-width scalar used for fixed-base scalar multiplication.
+/// This is decomposed in chunks of `window_width` bits in little-endian order.
+/// For example, if `window_width` = 3, we will have [k_0, k_1, ..., k_n]
+/// where `scalar = k_0 + k_1 * (2^3) + ... + k_n * (2^3)^n` and each `k_i` is
+/// in the range [0..2^3).
+#[derive(Clone, Debug)]
+pub struct EccScalarFixed {
+    value: Option<pallas::Scalar>,
+    windows: ArrayVec<CellValue<pallas::Base>, { constants::NUM_WINDOWS }>,
+}
+
+/// A signed short scalar used for fixed-base scalar multiplication.
+/// This is decomposed in chunks of `window_width` bits in little-endian order.
+/// For example, if `window_width` = 3, we will have [k_0, k_1, ..., k_n]
+/// where `scalar = k_0 + k_1 * (2^3) + ... + k_n * (2^3)^n` and each `k_i` is
+/// in the range [0..2^3).
+#[derive(Clone, Debug)]
+pub struct EccScalarFixedShort {
+    magnitude: Option<pallas::Scalar>,
+    sign: CellValue<pallas::Base>,
+    windows: ArrayVec<CellValue<pallas::Base>, { constants::NUM_WINDOWS_SHORT }>,
+}
+
 impl EccInstructions<pallas::Affine> for EccChip {
-    type ScalarFixed = (); // TODO
-    type ScalarFixedShort = (); // TODO
-    type ScalarVar = (); // TODO
+    type ScalarFixed = EccScalarFixed;
+    type ScalarFixedShort = EccScalarFixedShort;
+    type ScalarVar = CellValue<pallas::Base>;
     type Point = EccPoint;
     type X = CellValue<pallas::Base>;
-    type FixedPoints = (); // TODO
-    type FixedPointsShort = (); // TODO
+    type FixedPoints = OrchardFixedBasesFull;
+    type FixedPointsShort = ValueCommitV;
 
     fn constrain_equal(
         &self,
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -1,6 +1,11 @@
 use super::EccInstructions;
-use crate::circuit::gadget::utilities::{copy, CellValue, Var};
-use crate::constants::{self, OrchardFixedBasesFull, ValueCommitV};
+use crate::{
+    circuit::gadget::utilities::{
+        copy, lookup_range_check::LookupRangeCheckConfig, CellValue, Var,
+    },
+    constants::{self, OrchardFixedBasesFull, ValueCommitV},
+    primitives::sinsemilla,
+};
 use arrayvec::ArrayVec;
 
 use group::prime::PrimeCurveAffine;
@@ -78,8 +83,10 @@ pub struct EccConfig {
 
     /// Incomplete addition
     pub q_add_incomplete: Selector,
+
     /// Complete addition
     pub q_add: Selector,
+
     /// Variable-base scalar multiplication (hi half)
     pub q_mul_hi: Selector,
     /// Variable-base scalar multiplication (lo half)
@@ -89,19 +96,27 @@ pub struct EccConfig {
     /// Selector used in scalar decomposition for variable-base scalar mul
     pub q_init_z: Selector,
     /// Variable-base scalar multiplication (final scalar)
-    pub q_mul_complete: Selector,
+    pub q_mul_z: Selector,
+    /// Variable-base scalar multiplication (overflow check)
+    pub q_mul_overflow: Selector,
+
     /// Fixed-base full-width scalar multiplication
     pub q_mul_fixed: Selector,
+
     /// Fixed-base signed short scalar multiplication
     pub q_mul_fixed_short: Selector,
+
     /// Witness point
     pub q_point: Selector,
     /// Witness full-width scalar for fixed-base scalar mul
     pub q_scalar_fixed: Selector,
     /// Witness signed short scalar for full-width fixed-base scalar mul
     pub q_scalar_fixed_short: Selector,
+
     /// Permutation
     pub perm: Permutation,
+    /// 10-bit lookup
+    pub lookup_config: LookupRangeCheckConfig<pallas::Base, { sinsemilla::K }>,
 }
 
 /// A chip implementing EccInstructions
@@ -132,8 +147,12 @@ impl EccChip {
     pub fn configure(
         meta: &mut ConstraintSystem<pallas::Base>,
         advices: [Column<Advice>; 10],
+        lookup_table: Column<Fixed>,
         perm: Permutation,
     ) -> <Self as Chip<pallas::Base>>::Config {
+        let lookup_config =
+            LookupRangeCheckConfig::configure(meta, advices[9], lookup_table, perm.clone());
+
         let config = EccConfig {
             advices,
             lagrange_coeffs: [
@@ -153,13 +172,15 @@ impl EccChip {
             q_mul_lo: meta.selector(),
             q_mul_decompose_var: meta.selector(),
             q_init_z: meta.selector(),
-            q_mul_complete: meta.selector(),
+            q_mul_z: meta.selector(),
+            q_mul_overflow: meta.selector(),
             q_mul_fixed: meta.selector(),
             q_mul_fixed_short: meta.selector(),
             q_point: meta.selector(),
             q_scalar_fixed: meta.selector(),
             q_scalar_fixed_short: meta.selector(),
             perm,
+            lookup_config,
         };
 
         // Create witness point gate
@@ -193,13 +214,13 @@ impl EccChip {
             config.create_gate(meta);
         }
 
-        // Create witness scalar_fixed gate that only apploes to short scalars
+        // Create witness scalar_fixed gate that only applies to short scalars
         {
             let config: witness_scalar_fixed::short::Config = (&config).into();
             config.create_gate(meta);
         }
 
-        // Create fixed-base scalar mul gate that os used in both full-width
+        // Create fixed-base scalar mul gate that is used in both full-width
         // and short multiplication.
         {
             let mul_fixed_config: mul_fixed::Config<{ constants::NUM_WINDOWS }> = (&config).into();
@@ -214,6 +235,17 @@ impl EccChip {
         }
 
         config
+    }
+}
+
+/// A base-field element used as the scalar in variable-base scalar multiplication.
+#[derive(Copy, Clone, Debug)]
+pub struct EccScalarVar(CellValue<pallas::Base>);
+impl std::ops::Deref for EccScalarVar {
+    type Target = CellValue<pallas::Base>;
+
+    fn deref(&self) -> &CellValue<pallas::Base> {
+        &self.0
     }
 }
 
@@ -243,7 +275,7 @@ pub struct EccScalarFixedShort {
 impl EccInstructions<pallas::Affine> for EccChip {
     type ScalarFixed = EccScalarFixed;
     type ScalarFixedShort = EccScalarFixedShort;
-    type ScalarVar = CellValue<pallas::Base>;
+    type ScalarVar = EccScalarVar;
     type Point = EccPoint;
     type X = CellValue<pallas::Base>;
     type FixedPoints = OrchardFixedBasesFull;
@@ -277,12 +309,12 @@ impl EccInstructions<pallas::Affine> for EccChip {
             || "Witness scalar for variable-base mul",
             |mut region| {
                 let cell = region.assign_advice(
-                    || "Witness scalar var",
+                    || "witness scalar var",
                     config.advices[0],
                     0,
                     || value.ok_or(Error::SynthesisError),
                 )?;
-                Ok(CellValue::new(cell, value))
+                Ok(EccScalarVar(CellValue::new(cell, value)))
             },
         )
     }
@@ -360,9 +392,10 @@ impl EccInstructions<pallas::Affine> for EccChip {
         base: &Self::Point,
     ) -> Result<Self::Point, Error> {
         let config: mul::Config = self.config().into();
-        layouter.assign_region(
-            || "variable-base scalar mul",
-            |mut region| config.assign_region(scalar, base, 0, &mut region),
+        config.assign(
+            layouter.namespace(|| "variable-base scalar mul"),
+            *scalar,
+            base,
         )
     }
 
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -233,10 +233,22 @@ impl EccInstructions<pallas::Affine> for EccChip {
 
     fn witness_scalar_var(
         &self,
-        _layouter: &mut impl Layouter<pallas::Base>,
-        _value: Option<pallas::Base>,
+        layouter: &mut impl Layouter<pallas::Base>,
+        value: Option<pallas::Base>,
     ) -> Result<Self::ScalarVar, Error> {
-        todo!()
+        let config = self.config().clone();
+        layouter.assign_region(
+            || "Witness scalar for variable-base mul",
+            |mut region| {
+                let cell = region.assign_advice(
+                    || "Witness scalar var",
+                    config.advices[0],
+                    0,
+                    || value.ok_or(Error::SynthesisError),
+                )?;
+                Ok(CellValue::new(cell, value))
+            },
+        )
     }
 
     fn witness_scalar_fixed(
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -99,11 +99,13 @@ pub struct EccConfig {
     pub q_mul_overflow: Selector,
 
     /// Fixed-base full-width scalar multiplication
-    pub mul_fixed: Column<Fixed>,
+    pub q_mul_fixed: Selector,
     /// Fixed-base signed short scalar multiplication
     pub q_mul_fixed_short: Selector,
     /// Fixed-base multiplication using a base field element as the scalar
-    pub base_field_fixed: Column<Fixed>,
+    pub base_field_fixed_mul: Selector,
+    /// Canonicity checks on base field element used as scalar in fixed-base mul
+    pub base_field_fixed_canon: Selector,
 
     /// Witness point
     pub q_point: Selector,
@@ -184,9 +186,10 @@ impl EccChip {
             q_mul_decompose_var: meta.selector(),
             q_mul_overflow: meta.selector(),
             q_mul_lsb: meta.selector(),
-            mul_fixed: meta.fixed_column(),
+            q_mul_fixed: meta.selector(),
             q_mul_fixed_short: meta.selector(),
-            base_field_fixed: meta.fixed_column(),
+            base_field_fixed_mul: meta.selector(),
+            base_field_fixed_canon: meta.selector(),
             q_point: meta.selector(),
             q_scalar_fixed: meta.selector(),
             q_scalar_fixed_short: meta.selector(),
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -13,7 +13,7 @@ use pasta_curves::{arithmetic::CurveAffine, pallas};
 pub(super) mod add;
 pub(super) mod add_incomplete;
 pub(super) mod mul;
-// pub(super) mod mul_fixed;
+pub(super) mod mul_fixed;
 pub(super) mod witness_point;
 pub(super) mod witness_scalar_fixed;
 
@@ -199,6 +199,20 @@ impl EccChip {
             config.create_gate(meta);
         }
 
+        // Create fixed-base scalar mul gate that os used in both full-width
+        // and short multiplication.
+        {
+            let mul_fixed_config: mul_fixed::Config<{ constants::NUM_WINDOWS }> = (&config).into();
+            mul_fixed_config.create_gate(meta);
+        }
+
+        // Create gate that is only used in short fixed-base scalar mul.
+        {
+            let short_config: mul_fixed::short::Config<{ constants::NUM_WINDOWS_SHORT }> =
+                (&config).into();
+            short_config.create_gate(meta);
+        }
+
         config
     }
 }
@@ -354,19 +368,29 @@ impl EccInstructions<pallas::Affine> for EccChip {
 
     fn mul_fixed(
         &self,
-        _layouter: &mut impl Layouter<pallas::Base>,
-        _scalar: &Self::ScalarFixed,
-        _base: &Self::FixedPoints,
+        layouter: &mut impl Layouter<pallas::Base>,
+        scalar: &Self::ScalarFixed,
+        base: &Self::FixedPoints,
     ) -> Result<Self::Point, Error> {
-        todo!()
+        let config: mul_fixed::full_width::Config<{ constants::NUM_WINDOWS }> =
+            self.config().into();
+        layouter.assign_region(
+            || format!("fixed-base mul of {:?}", base),
+            |mut region| config.assign_region(scalar, *base, 0, &mut region),
+        )
     }
 
     fn mul_fixed_short(
         &self,
-        _layouter: &mut impl Layouter<pallas::Base>,
-        _scalar: &Self::ScalarFixedShort,
-        _base: &Self::FixedPointsShort,
+        layouter: &mut impl Layouter<pallas::Base>,
+        scalar: &Self::ScalarFixedShort,
+        base: &Self::FixedPointsShort,
     ) -> Result<Self::Point, Error> {
-        todo!()
+        let config: mul_fixed::short::Config<{ constants::NUM_WINDOWS_SHORT }> =
+            self.config().into();
+        layouter.assign_region(
+            || format!("short fixed-base mul of {:?}", base),
+            |mut region| config.assign_region(scalar, base, 0, &mut region),
+        )
     }
 }
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -15,7 +15,7 @@ pub(super) mod add_incomplete;
 pub(super) mod mul;
 // pub(super) mod mul_fixed;
 pub(super) mod witness_point;
-// pub(super) mod witness_scalar_fixed;
+pub(super) mod witness_scalar_fixed;
 
 /// A curve point represented in affine (x, y) coordinates. Each coordinate is
 /// assigned to a cell.
@@ -186,6 +186,19 @@ impl EccChip {
             mul_config.create_gate(meta);
         }
 
+        // Create witness scalar_fixed gate that applies to both full-width and
+        // short scalars
+        {
+            let config: witness_scalar_fixed::Config = (&config).into();
+            config.create_gate(meta);
+        }
+
+        // Create witness scalar_fixed gate that only apploes to short scalars
+        {
+            let config: witness_scalar_fixed::short::Config = (&config).into();
+            config.create_gate(meta);
+        }
+
         config
     }
 }
@@ -262,18 +275,26 @@ impl EccInstructions<pallas::Affine> for EccChip {
 
     fn witness_scalar_fixed(
         &self,
-        _layouter: &mut impl Layouter<pallas::Base>,
-        _value: Option<pallas::Scalar>,
+        layouter: &mut impl Layouter<pallas::Base>,
+        value: Option<pallas::Scalar>,
     ) -> Result<Self::ScalarFixed, Error> {
-        todo!()
+        let config: witness_scalar_fixed::full_width::Config = self.config().into();
+        layouter.assign_region(
+            || "witness scalar for fixed-base mul",
+            |mut region| config.assign_region(value, 0, &mut region),
+        )
     }
 
     fn witness_scalar_fixed_short(
         &self,
-        _layouter: &mut impl Layouter<pallas::Base>,
-        _value: Option<pallas::Scalar>,
+        layouter: &mut impl Layouter<pallas::Base>,
+        value: Option<pallas::Scalar>,
     ) -> Result<Self::ScalarFixedShort, Error> {
-        todo!()
+        let config: witness_scalar_fixed::short::Config = self.config().into();
+        layouter.assign_region(
+            || "witness short scalar for fixed-base mul",
+            |mut region| config.assign_region(value, 0, &mut region),
+        )
     }
 
     fn witness_point(
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -101,10 +101,11 @@ pub struct EccConfig {
     pub q_mul_overflow: Selector,
 
     /// Fixed-base full-width scalar multiplication
-    pub q_mul_fixed: Selector,
-
+    pub mul_fixed: Column<Fixed>,
     /// Fixed-base signed short scalar multiplication
     pub q_mul_fixed_short: Selector,
+    /// Fixed-base multiplication using a base field element as the scalar
+    pub base_field_fixed: Column<Fixed>,
 
     /// Witness point
     pub q_point: Selector,
@@ -174,8 +175,9 @@ impl EccChip {
             q_init_z: meta.selector(),
             q_mul_z: meta.selector(),
             q_mul_overflow: meta.selector(),
-            q_mul_fixed: meta.selector(),
+            mul_fixed: meta.fixed_column(),
             q_mul_fixed_short: meta.selector(),
+            base_field_fixed: meta.fixed_column(),
             q_point: meta.selector(),
             q_scalar_fixed: meta.selector(),
             q_scalar_fixed_short: meta.selector(),
@@ -224,7 +226,7 @@ impl EccChip {
         // and short multiplication.
         {
             let mul_fixed_config: mul_fixed::Config<{ constants::NUM_WINDOWS }> = (&config).into();
-            mul_fixed_config.create_gate(meta);
+            mul_fixed_config.create_gate_scalar(meta);
         }
 
         // Create gate that is only used in short fixed-base scalar mul.
@@ -232,6 +234,12 @@ impl EccChip {
             let short_config: mul_fixed::short::Config<{ constants::NUM_WINDOWS_SHORT }> =
                 (&config).into();
             short_config.create_gate(meta);
+        }
+
+        // Create gate ths is only used in fixed-base mul using a base field element.
+        {
+            let base_field_config: mul_fixed::base_field_elem::Config = (&config).into();
+            base_field_config.create_gate(meta);
         }
 
         config
@@ -250,10 +258,10 @@ impl std::ops::Deref for EccScalarVar {
 }
 
 /// A full-width scalar used for fixed-base scalar multiplication.
-/// This is decomposed in chunks of `window_width` bits in little-endian order.
-/// For example, if `window_width` = 3, we will have [k_0, k_1, ..., k_n]
-/// where `scalar = k_0 + k_1 * (2^3) + ... + k_n * (2^3)^n` and each `k_i` is
-/// in the range [0..2^3).
+/// This is decomposed into 85 3-bit windows in little-endian order,
+/// i.e. `windows` = [k_0, k_1, ..., k_84] (for a 255-bit scalar)
+/// where `scalar = k_0 + k_1 * (2^3) + ... + k_84 * (2^3)^84` and
+/// each `k_i` is in the range [0..2^3).
 #[derive(Clone, Debug)]
 pub struct EccScalarFixed {
     value: Option<pallas::Scalar>,
@@ -261,15 +269,38 @@ pub struct EccScalarFixed {
 }
 
 /// A signed short scalar used for fixed-base scalar multiplication.
-/// This is decomposed in chunks of `window_width` bits in little-endian order.
-/// For example, if `window_width` = 3, we will have [k_0, k_1, ..., k_n]
-/// where `scalar = k_0 + k_1 * (2^3) + ... + k_n * (2^3)^n` and each `k_i` is
-/// in the range [0..2^3).
+/// A short scalar must have magnitude in the range [0..2^64), with
+/// a sign of either 1 or -1.
+/// This is decomposed into 22 3-bit windows in little-endian order,
+/// i.e. `windows` = [k_0, k_1, ..., k_21] (for a 64-bit magnitude)
+/// where `scalar = k_0 + k_1 * (2^3) + ... + k_84 * (2^3)^84` and
+/// each `k_i` is in the range [0..2^3).
+/// k_21 must be a single bit, i.e. 0 or 1.
 #[derive(Clone, Debug)]
 pub struct EccScalarFixedShort {
     magnitude: Option<pallas::Scalar>,
     sign: CellValue<pallas::Base>,
     windows: ArrayVec<CellValue<pallas::Base>, { constants::NUM_WINDOWS_SHORT }>,
+}
+
+/// A base field element used for fixed-base scalar multiplication.
+/// This is decomposed into 3-bit windows in little-endian order
+/// using a running sum `z`, where z_{i+1} = (z_i - a_i) / (2^3)
+/// for element α = a_0 + (2^3) a_1 + ... + (2^{3(n-1)}) a_{n-1}.
+/// Each `a_i` is in the range [0..2^3).
+///
+/// `windows` = [z_1, ..., z_85], where we expect z_85 = 0.
+/// Since z_0 is initialized as α, we store it as `base_field_elem`.
+#[derive(Clone, Debug)]
+struct EccBaseFieldElemFixed {
+    base_field_elem: CellValue<pallas::Base>,
+    running_sum: ArrayVec<CellValue<pallas::Base>, { constants::NUM_WINDOWS }>,
+}
+
+impl EccBaseFieldElemFixed {
+    fn base_field_elem(&self) -> CellValue<pallas::Base> {
+        self.base_field_elem
+    }
 }
 
 impl EccInstructions<pallas::Affine> for EccChip {
@@ -424,6 +455,19 @@ impl EccInstructions<pallas::Affine> for EccChip {
         layouter.assign_region(
             || format!("short fixed-base mul of {:?}", base),
             |mut region| config.assign_region(scalar, base, 0, &mut region),
+        )
+    }
+
+    fn mul_fixed_base_field_elem(
+        &self,
+        layouter: &mut impl Layouter<pallas::Base>,
+        base_field_elem: CellValue<pallas::Base>,
+        base: &Self::FixedPoints,
+    ) -> Result<Self::Point, Error> {
+        let config: mul_fixed::base_field_elem::Config = self.config().into();
+        layouter.assign_region(
+            || format!("base field elem fixed-base mul of {:?}", base),
+            |mut region| config.assign_region(base_field_elem, *base, 0, &mut region),
         )
     }
 }
--- a/src/circuit/gadget/ecc/chip/mul.rs
+++ b/src/circuit/gadget/ecc/chip/mul.rs
@@ -1,5 +1,8 @@
 use super::{add, CellValue, EccConfig, EccPoint, EccScalarVar, Var};
-use crate::constants::{NUM_COMPLETE_BITS, T_Q};
+use crate::{
+    circuit::gadget::utilities::copy,
+    constants::{NUM_COMPLETE_BITS, T_Q},
+};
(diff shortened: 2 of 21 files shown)


src/circuit/gadget/ecc/chip/witness_scalar_fixed/short.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub fn create_gate(&self, meta: &mut ConstraintSystem<C::Base>) {
+        // Check that sign is either 1 or -1.
+        meta.create_gate("check sign", |meta| {
+            let q_scalar_fixed_short = meta.query_selector(self.q_scalar_fixed_short);
+            let sign = meta.query_advice(self.super_config.window, Rotation::cur());
+
+            vec![
+                q_scalar_fixed_short
+                    * (sign.clone() + Expression::Constant(C::Base::one()))
+                    * (sign - Expression::Constant(C::Base::one())),
```
therealyingtong, line 35: Simpler might be `q_scalar_fixed_short * (sign.clone() * sign.clone() - Expression::Constant(C::Base::one()))` (aka check that the sign bit squared equals one)

_Originally posted by @ebfull in https://github.com/zcash/orchard/pull/107#discussion_r647654106_

src/circuit/gadget/ecc/chip.rs
```diff
@@ -202,26 +239,45 @@ where
...
+            |mut region| {
+                let cell = region.assign_advice(
+                    || "Scalar var",
+                    self.config().advices[0],
+                    0,
+                    || value.ok_or(Error::SynthesisError),
+                )?;
+                Ok(CellValue::new(cell, value))
+            },
+        )
```
therealyingtong, line 256: Unresolved comment: https://github.com/zcash/orchard/pull/73#discussion_r640939832

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,472 @@
...
+    fn create_final_scalar_gate(&self, meta: &mut ConstraintSystem<C::Base>) {
+        meta.create_gate("Decompose scalar for variable-base mul", |meta| {
+            let q_mul_decompose_var = meta.query_selector(self.q_mul_decompose_var);
+            let scalar = meta.query_advice(self.scalar, Rotation::cur());
+            let z_cur = meta.query_advice(self.z_complete, Rotation::cur());
+
+            // The scalar field `F_q = 2^254 + t_q`.
+            // -((2^127)^2) = -(2^254) = t_q (mod q)
+            let t_q = -(C::Scalar::from_u128(1u128 << 127).square());
+            let t_q = C::Base::from_bytes(&t_q.to_bytes()).unwrap();
```
therealyingtong, line 109: TODO: add `t_q` (and corresponding `t_p` for Pallas) as constants.

_Originally posted by @therealyingtong in https://github.com/zcash/orchard/pull/107#discussion_r646072661_

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,529 @@
...
+        region: &mut Region<'_, pallas::Base>,
+        offset: usize,
+        mut acc: EccPoint,
+        base: OrchardFixedBases,
+        scalar: &ScalarFixed,
+    ) -> Result<EccPoint, Error> {
+        // This is 2^w, where w is the window width
+        let h = pallas::Scalar::from_u64(constants::H as u64);
+
+        let base_value = base.generator();
```
str4d, line 361: Likewise here:
image

This is because... (see next comment)

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,529 @@
...
+    }
+}
+
+impl OrchardFixedBases {
+    pub fn generator(self) -> pallas::Affine {
+        match self {
+            Self::ValueCommitV => constants::value_commit_v::generator(),
+            Self::Full(base) => {
+                let base: OrchardFixedBase = base.into();
+                base.generator
```
str4d, line 52: The `base.into()` here computes the lagrange coefficients inline:
https://github.com/zcash/orchard/blob/37b1b7f357cd34d93f9c55bb96efd05ab4e84408/src/constants/load.rs#L24-L61
But we then immediately discard them and just return the generator itself. Instead, add a `generator()` method to `OrchardFixedBasesFull` which can sidestep deriving the entire `OrchardFixedBase`.

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,529 @@
...
+            )?;
+            let y = CellValue::new(y_cell, y);
+
+            EccPoint { x, y }
+        };
+
+        // Assign u = (y_p + z_w).sqrt() for `m0`
+        {
+            let k0 = scalar.windows_usize()[0];
+            let u0 = &base.u()[0];
```
str4d, line 323: Same thing happens twice here (inside `base.generator()` and `base.u()`):
image

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,529 @@
...
+            }
+        }
+    }
+
+    pub fn u(self) -> Vec<WindowUs> {
+        match self {
+            Self::ValueCommitV => ValueCommitV::get().u_short.0.as_ref().to_vec(),
+            Self::Full(base) => {
+                let base: OrchardFixedBase = base.into();
+                base.u.0.as_ref().to_vec()
```
str4d, line 62: So let's also add an `OrchardFixedBasesFull::u` method to use here.

str4d (commented):
Found some easy performance wins!

src/circuit/gadget/ecc/chip.rs
```diff
@@ -66,28 +72,44 @@ pub struct EccConfig {
...
+    /// Selector used in scalar decomposition for variable-base scalar mul
+    pub q_init_z: Selector,
     /// Variable-base scalar multiplication (final scalar)
-    pub q_mul_complete: Selector,
+    pub q_mul_z: Selector,
+    /// Variable-base scalar multiplication (overflow check)
+    pub q_mul_overflow: Selector,
+    // Fixed column to constrain s_{120..=125} of s = alpha + k_{254} * 2^127
+    pub q_three_bit: Selector,
+    // Fixed column to constrain s_{126} of s = alpha + k_{254} * 2^127
```
daira, line 93: It's simpler to treat the value as as being 130 bits rather than 127 bits, so that we don't need to special-case the top element. I've described this at https://hackmd.io/o9EzZBwxSWSi08kQ_fMIOw?both#Overflow-check-for-variable-base-scalar-multiplication .

src/circuit/gadget/ecc/chip.rs
```diff
@@ -66,28 +72,44 @@ pub struct EccConfig {
...
+    pub q_one_bit: Selector,
```
daira, line 94: It's simpler to treat the value as as being 130 bits rather than 127 bits, so that we don't need to special-case the top element. I've described this at https://hackmd.io/o9EzZBwxSWSi08kQ_fMIOw?both#Overflow-check-for-variable-base-scalar-multiplication .

```suggestion
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -138,13 +164,18 @@ impl EccChip {
             q_mul_hi: meta.selector(),
             q_mul_lo: meta.selector(),
             q_mul_decompose_var: meta.selector(),
-            q_mul_complete: meta.selector(),
+            q_init_z: meta.selector(),
+            q_mul_z: meta.selector(),
+            q_mul_overflow: meta.selector(),
+            q_three_bit: meta.selector(),
+            q_one_bit: meta.selector(),
```
daira, line 171:
```suggestion
```

src/circuit/gadget/ecc/chip/mul/overflow.rs
```diff
@@ -0,0 +1,401 @@
...
+
+pub struct Config {
+    // Selector to check decomposition of lsb
+    q_mul_z: Selector,
+    // Selector to check z_0 = alpha + t_q (mod p)
+    q_mul_overflow: Selector,
+    // Selector to constrain s_{120..=125} of s = alpha + k_{254} * 2^127
+    q_three_bit: Selector,
+    // Selector to constrain s_{126} of s = alpha + k_{254} * 2^127
+    q_one_bit: Selector,
```
daira, line 26:
```suggestion
```

src/circuit/gadget/ecc/chip/mul/overflow.rs
```diff
@@ -0,0 +1,401 @@
...
+    perm: Permutation,
+}
+
+impl From<&EccConfig> for Config {
+    fn from(ecc_config: &EccConfig) -> Self {
+        Self {
+            q_mul_z: ecc_config.q_mul_z,
+            q_mul_overflow: ecc_config.q_mul_overflow,
+            q_three_bit: ecc_config.q_three_bit,
+            q_one_bit: ecc_config.q_one_bit,
```
daira, line 41:
```suggestion
```

src/circuit/gadget/ecc/chip/mul/overflow.rs
```diff
@@ -0,0 +1,401 @@
...
+            let z_prev = meta.query_advice(self.lookup_config.running_sum, Rotation::prev());
+            let z_cur = meta.query_advice(self.lookup_config.running_sum, Rotation::cur());
+
+            // z_cur = (z_prev - word) / (2)
+            // word = z_prev - z_cur * 2
+            let one_bit_value = z_prev - z_cur * pallas::Base::from_u64(2);
+
+            let one = Expression::Constant(pallas::Base::one());
+            vec![q_one_bit * one_bit_value.clone() * (one - one_bit_value)]
+        });
```
daira, line 133:
```suggestion
```

src/circuit/gadget/ecc/chip/mul/overflow.rs
```diff
@@ -0,0 +1,401 @@
...
+                        // s_{126}
+                        let s_126 = pallas::Base::from_u64(leftover_bits[6] as u64);
+
+                        (Some(lo_three), Some(hi_three), Some(s_126))
+                    } else {
+                        (None, None, None)
+                    }
+                };
+
+                // Subtract s_{120..=122}
```
daira, line 328: All of this goes away (and we check 130 bits).

src/circuit/gadget/ecc/chip.rs
```diff
@@ -202,26 +239,45 @@ where
 
     fn witness_scalar_var(
         &self,
-        _layouter: &mut impl Layouter<C::Base>,
-        _value: Option<C::Base>,
+        layouter: &mut impl Layouter<C::Base>,
+        value: Option<C::Base>,
```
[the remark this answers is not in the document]
daira: Yes, the correctness argument would need to be re-checked if it were applied to Vesta.

src/circuit/gadget/ecc/chip/mul/overflow.rs
```diff
@@ -0,0 +1,244 @@
...
+        // s = alpha + k_254 ⋅ 2^130 is witnessed here, and then copied into
+        // the decomposition as well as the overflow check gate.
+        // In the overflow check gate, we check that s is properly derived
+        // from alpha and k_254.
+        let s = {
+            let k_254 = *zs[254];
+            let s_val = alpha
+                .value()
+                .zip(k_254.value())
+                .map(|(alpha, k_254)| alpha + k_254 * pallas::Base::from_u128(1 << 65).square());
```
daira, line 114: https://github.com/zcash/pasta_curves/issues/14

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,426 @@
...
+const INCOMPLETE_LO_RANGE: Range<usize> = (INCOMPLETE_LEN / 2)..INCOMPLETE_LEN;
+
+// Bits k_{3} to k_{1} inclusive are used in complete addition.
+// Bit k_{0} is handled separately.
+const COMPLETE_RANGE: Range<usize> = INCOMPLETE_LEN..(INCOMPLETE_LEN + NUM_COMPLETE_BITS);
+
+pub struct Config {
+    // Selector used to constrain the initialization of the running sum to be zero.
+    q_init_z: Selector,
+    // Selector used to check z_0 = 2*z_1 + k_0
```
ebfull, line 39: Should be `z_i = 2*z_{i+1} + k_i`

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,426 @@
...
+        };
+        let y_p = if let Some(lsb) = lsb {
+            if !lsb {
+                base.y.value().map(|y_p| -y_p)
+            } else {
+                Some(pallas::Base::zero())
+            }
+        } else {
+            None
+        };
```
ebfull, line 285: Nothing is constraining that this value (witnessed by the prover) is the correct value -- which should be Acc - P or Acc depending on the value of `lsb`, so there's a missing gate here.

ebfull (changes requested):

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,426 @@
...
+                let acc = self
+                    .add_config
+                    .assign_region(&base, &base, offset, &mut region)?;
+
+                // Increase the offset by 1 after complete addition.
+                let offset = offset + 1;
+
+                // Initialize the running sum for scalar decomposition to zero
+                let z_init = {
+                    // Constrain the initialization of `z` to equal zero.
```
ebfull, line 131: Let's make a TODO to enforce this to equal zero via public inputs or something (using permutation argument).

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,427 @@
...
+        // Increase the offset by 1 after initializing `z`.
+        let offset = offset + 1;
+
+        // Double-and-add (incomplete addition) for the `hi` half of the scalar decomposition
+        let (x_a, y_a, zs_incomplete_hi) = self.hi_config.double_and_add(
+            region,
+            offset,
+            &base,
+            bits_incomplete_hi,
+            (X(acc.x), Y(acc.y.value()), z_init),
```
str4d: We don't use `acc.y.cell()` anywhere, so we aren't constraining the output of the initial complete addition to be the input to the incomplete addition. We need to pass `acc.y` in entirely, and then inside `double_and_add` have the initial values of lambda_1 etc. be appropriately constrained.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,427 @@
...
+        )?;
+
+        // Double-and-add (incomplete addition) for the `lo` half of the scalar decomposition
+        let z = &zs_incomplete_hi[zs_incomplete_hi.len() - 1];
+        let (x_a, y_a, zs_incomplete_lo) = self.lo_config.double_and_add(
+            region,
+            offset,
+            &base,
+            bits_incomplete_lo,
+            (x_a, y_a, *z),
```
str4d: Similarly, `hi_config.double_and_add` is only returning the value of the intermediate `y_a`, not an assigned cell that can be constrained into `lo_config.double_and_add` to connect them together. If `double_and_add` assigns its final `y_a`, then we can just pass that in here as we would with `acc.y` above.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,427 @@
...
+        let offset = offset + INCOMPLETE_LO_RANGE.len() + 2;
+
+        // Complete addition
+        let (acc, zs_complete) = {
+            let z = &zs_incomplete_lo[zs_incomplete_lo.len() - 1];
+            let complete_config: complete::Config = self.into();
+            // Bits used in complete addition. k_{3} to k_{1} inclusive
+            // The LSB k_{0} is handled separately.
+            let bits_complete = &bits[COMPLETE_RANGE];
+            complete_config.assign_region(region, offset, bits_complete, base, x_a, y_a, *z)?
```
str4d: This is assigning `y_a` to a cell internally, but isn't connecting it to the incomplete addition output (because as above, `double_and_add` only returns a value). Once `double_and_add` returns a `CellValue`, we can just pass that in here and copy it into the complete addition.

src/circuit/gadget/ecc/chip/mul/incomplete.rs
```diff
@@ -0,0 +1,296 @@
...
+                self.x_p,
+                row + offset,
+                || x_p.ok_or(Error::SynthesisError),
+            )?;
+            region.assign_advice(
+                || "y_p",
+                self.y_p,
+                row + offset,
+                || y_p.ok_or(Error::SynthesisError),
+            )?;
```
str4d, line 254: This causes a double-assignment in the circuit, due to sharing these columns between `hi_config` and `lo_config`. This is fine, as we pass the same `base` into both `incomplete_addition` invocations.

src/circuit/gadget/ecc/chip/mul/incomplete.rs
```diff
@@ -0,0 +1,296 @@
...
+                    lambda1_cur.clone() * lambda1_cur.clone() - x_a_cur.clone() - x_p_cur.clone();
+                lambda_neg * (x_a_cur.clone() - lambda1_expr)
+                    - Expression::Constant(pallas::Base::from_u64(2)) * y_a_cur.clone()
+            };
+
+            // λ_{2,i}^2 − x_{A,i+1} −(λ_{1,i}^2 − x_{A,i} − x_{P,i}) − x_{A,i} = 0
+            let poly3 = lambda2_cur.clone() * lambda2_cur.clone()
+                - x_a_next.clone()
+                - (lambda1_cur.clone() * lambda1_cur)
+                + x_p_cur;
```
str4d, line 154:
```suggestion
            let secant_line =
                lambda2_cur.clone().square() - x_a_next.clone() - lambda1_cur.square() + x_p_cur;
```

Non-blocking: We should also make the `x_a_cur` elimination simplification in the Sinsemilla gate (which I think is what would remove the need for an `x_r` helper). This should happen in a separate PR.

src/circuit/gadget/ecc/chip/mul/incomplete.rs
```diff
@@ -0,0 +1,296 @@
...
+            // Q is point at infinity
+            || (x_a == pallas::Base::zero() && y_a == pallas::Base::zero())
+            // x_p = x_a
+            || (x_p == x_a)
+            {
+                return Err(Error::SynthesisError);
+            }
+        }
+
+        // Enable `q_mul` on all but the last row of the incomplete range.
```
str4d, line 198: Yep, disabling this on the last row means we aren't constraining the `y_A` output of this `double_and_add`.

str4d (commented):
Partial review of first 3 commits (I haven't looked at `complete.rs` yet).

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,427 @@
...
+
+        // Initialize the accumulator `acc = [2]base`
+        let acc = self
+            .add_config
+            .assign_region(&base, &base, offset, region)?;
+
+        // Increase the offset by 1 after complete addition.
+        let offset = offset + 1;
+
+        // Initialize the running sum for scalar decomposition to zero
```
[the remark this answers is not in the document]
therealyingtong: Now that I'm checking the initialised `y_a`, I'll need to use an extra cell on this row.

src/circuit/gadget/ecc/chip/mul/incomplete.rs
```diff
@@ -0,0 +1,418 @@
...
+
+                // λ_{1,i}⋅(x_{A,i} − x_{P,i}) − y_{A,i} + (2k_i - 1) y_{P,i} = 0
+                let poly1 = lambda1_cur * (x_a_cur.clone() - x_p_cur) - y_a_cur.clone()
+                    + (k * pallas::Base::from_u64(2) - one.clone()) * y_p_cur;
+
+                // λ_{2,i}^2 − x_{A,i-1} − x_{R,i} − x_{A,i} = 0
+                let secant_line = lambda2_cur.clone().square()
+                    - x_a_next.clone()
+                    - x_r(meta, Rotation::cur())
+                    - x_a_cur.clone();
```
str4d, line 178: @ebfull is there any benefit to simplifying custom gates manually (eliminating `x_a_cur` here)? And/or can that simplification be done later if necessary, or automated, without changing `PinnedVerificationKey`?

str4d (commented):
Flushing comments.

src/circuit/gadget/ecc/chip/mul/incomplete.rs
daira: I will make a book PR for that.

daira: Is this what you meant? https://hackmd.io/iOw7-HpFQY6dPF1aFY8pAw?both#Optimized-Sinsemilla-gate

Note that there is still a reference to x_{A,i} in each step.

daira: We established that this doesn't improve performance because x<sub>R,i</sub> still needs to be computed for Y<sub>A,i</sub>. (It's possible to save one addition by defining a temporary as λ<sub>1,i</sub><sup>2</sup> - x<sub>P,i</sub> but that's neither here nor there; we should take the simpler option and avoid changes at this stage.)

src/circuit/gadget/ecc/chip.rs
```diff
@@ -176,18 +213,114 @@ impl EccChip {
...
+}
+
+/// A base field element used for fixed-base scalar multiplication.
+/// This is decomposed into 3-bit windows in little-endian order
+/// using a running sum `z`, where z_{i+1} = (z_i - a_i) / (2^3)
+/// for element α = a_0 + (2^3) a_1 + ... + (2^{3(n-1)}) a_{n-1}.
+/// Each `a_i` is in the range [0..2^3).
+///
+/// `windows` = [z_1, ..., z_85], where we expect z_85 = 0.
+/// Since z_0 is initialized as α, we store it as `base_field_elem`.
```
daira, line 303:
```suggestion
/// Since z_0 is initialized as the scalar α, we store it as `base_field_elem`.
```

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -0,0 +1,251 @@
...
+        // [0]B should return (0,0) since it uses complete addition
+        // on the last step.
+        {
+            let scalar_fixed = pallas::Scalar::zero();
+            let scalar_fixed = ScalarFixedShort::new(
+                chip.clone(),
+                layouter.namespace(|| "ScalarFixedShort"),
+                Some(scalar_fixed),
+            )?;
+            value_commit_v.mul(layouter.namespace(|| "mul"), &scalar_fixed)?;
```
str4d, line 191: Check the result.

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -0,0 +1,251 @@
...
+            .enable(region, offset + NUM_WINDOWS)?;
+
+        // Assign final `x, y` to `x_p, y_p` columns and return final point
+        let x_val = magnitude_mul.x.value();
+        let x_var = region.assign_advice(
+            || "x_var",
+            self.super_config.x_p,
+            offset + NUM_WINDOWS,
+            || x_val.ok_or(Error::SynthesisError),
+        )?;
```
str4d, line 116: Blocking: This cell isn't constrained to equal `magnitude_mul.x`. However, AFAICT we don't actually need to assign it, since we can just return `magnitude_mul.x` (this is the end of the overall region, we don't need the output to be in a specific column).
```suggestion
        // Assign final `y` to `y_p` column, and return final point.
```

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -0,0 +1,251 @@
...
+        let y_var = region.assign_advice(
+            || "y_var",
+            self.super_config.y_p,
+            offset + NUM_WINDOWS,
+            || y_val.ok_or(Error::SynthesisError),
+        )?;
+
+        let result = EccPoint {
+            x: CellValue::new(x_var, x_val),
```
str4d, line 125:
```suggestion
            x: magnitude_mul.x,
```

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,580 @@
...
+                acc + (window_pow.clone() * meta.query_fixed(*coeff, Rotation::cur()))
+            },
+        );
+
+        // Check interpolation of x-coordinate
+        let x_check = interpolated_x - x_p;
+        // Check that `y + z = u^2`, where `z` is fixed and `u`, `y` are witnessed
+        let y_check = u.clone() * u - y_p - z;
+
+        vec![toggle.clone() * x_check, toggle * y_check]
```
str4d, line 180: Blocking: Unless it's somewhere else, this is missing the on-curve check:
```suggestion
        // Check interpolation of x-coordinate
        let x_check = interpolated_x - x_p.clone();
        // Check that `y + z = u^2`, where `z` is fixed and `u`, `y` are witnessed
        let y_check = u.square() - y_p.clone() - z;
        // Check that (x, y) is on the curve
        let on_curve = y_p.square() - x_p.clone().square() * x_p - 5;

        std::array::IntoIter::new([
            ("check x", x_check),
            ("check y", y_check),
            ("on-curve", on_curve),
        ]).map(move |(name, poly)| (name, toggle.clone() * poly))
```
(I don't remember if there's a constant for `b` in `pasta_curves`.)

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,580 @@
...
+        // Process all windows excluding least and most significant windows
+        let acc = self.add_incomplete(region, offset, acc, base, scalar)?;
+
+        // Process most significant window using complete addition
+        let mul_b = self.process_msb(region, offset, base, scalar)?;
+
+        Ok((acc, mul_b))
+    }
+
+    fn assign_fixed_constants(
```
str4d, line 207: I got up to here.

str4d (commented):
The only files I haven't reviewed are:
- `src/circuit/gadget/ecc/chip/mul_fixed.rs` (started)
- `src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs`

src/circuit/gadget/ecc/chip/mul_fixed/full_width.rs
```diff
@@ -0,0 +1,185 @@
...
+        base: FixedPoint<pallas::Affine, EccChip>,
+    ) -> Result<(), Error>
+    where
+        pallas::Scalar: PrimeFieldBits,
+    {
+        // [a]B
+        {
+            let scalar_fixed = pallas::Scalar::rand();
+
+            let scalar_fixed = ScalarFixed::new(
```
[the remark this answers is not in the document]
therealyingtong: Let's do this rename in the Action circuit PR.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -176,18 +213,115 @@ impl EccChip {
...
+    magnitude: Option<pallas::Scalar>,
+    sign: CellValue<pallas::Base>,
+    windows: ArrayVec<CellValue<pallas::Base>, { constants::NUM_WINDOWS_SHORT }>,
+}
+
+/// A base field element used for fixed-base scalar multiplication.
+/// This is decomposed into 3-bit windows in little-endian order
+/// using a running sum `z`, where z_{i+1} = (z_i - a_i) / (2^3)
+/// for element α = a_0 + (2^3) a_1 + ... + (2^{3(n-1)}) a_{n-1}.
+/// Each `a_i` is in the range [0..2^3).
```
daira, line 300: This also applies to the comment above (which is confusing because that's the comment that will be read first).

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,484 @@
+use super::{add, CellValue, EccConfig, EccPoint, EccScalarVar, Var};
+use crate::{
+    circuit::gadget::utilities::copy,
+    constants::{NUM_COMPLETE_BITS, T_Q},
```
daira, line 4: If `NUM_COMPLETE_BITS` is not used in other files, just move it to this one.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,484 @@
...
+const INCOMPLETE_LEN: usize = pallas::Scalar::NUM_BITS as usize - 1 - NUM_COMPLETE_BITS;
+const INCOMPLETE_RANGE: Range<usize> = 0..INCOMPLETE_LEN;
+
+// Bits k_{254} to k_{4} inclusive are used in incomplete addition.
+// The `hi` half is k_{254} to k_{130} inclusive (length 125 bits).
+const INCOMPLETE_HI_RANGE: Range<usize> = 0..(INCOMPLETE_LEN / 2);
+
+// Bits k_{254} to k_{4} inclusive are used in incomplete addition.
+// The `lo` half is k_{129} to k_{4} inclusive (length 126 bits).
+const INCOMPLETE_LO_RANGE: Range<usize> = (INCOMPLETE_LEN / 2)..INCOMPLETE_LEN;
```
daira, line 33: Add a static assertion that the boundary is at 130, i.e. that `INCOMPLETE_LO_RANGE.len() + 1 + NUM_COMPLETE_BITS == 130`.

daira: Add a link to the book here.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,484 @@
...
+mod complete;
+mod incomplete;
+mod overflow;
+
+// Bits used in incomplete addition. k_{254} to k_{4} inclusive
+const INCOMPLETE_LEN: usize = pallas::Scalar::NUM_BITS as usize - 1 - NUM_COMPLETE_BITS;
+const INCOMPLETE_RANGE: Range<usize> = 0..INCOMPLETE_LEN;
+
+// Bits k_{254} to k_{4} inclusive are used in incomplete addition.
+// The `hi` half is k_{254} to k_{130} inclusive (length 125 bits).
```
daira, line 28:
```suggestion
// The `hi` half is k_{254} to k_{130} inclusive (length 125 bits).
// (It is a coincidence that k_{130} matches the boundary of the
// overflow check described in <ref book>.)
```

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,484 @@
...
+            //    z_0 = 2 * z_1 + k_0
+            // => k_0 = z_0 - 2 * z_1
+            let lsb = z_0 - z_1 * pallas::Base::from_u64(2);
+            let one_minus_lsb = Expression::Constant(pallas::Base::one()) - lsb.clone();
+
+            let bool_check = lsb.clone() * one_minus_lsb.clone();
+
+            // `lsb` = 0 => (x_p, y_p) = (x, -y)
+            // `lsb` = 1 => (x_p, y_p) = (0,0)
+            let lsb_x = (lsb.clone() * x_p.clone()) + one_minus_lsb.clone() * (x_p - base_x);
```
daira, line 116: We can expand this as ```
  lsb * x_p + (1 - lsb) * x_p - (1 - lsb) * base_x
= x_p - (1 - lsb) * base_x
```

File: src/circuit/gadget/ecc/chip/mul.rs [hunk 40]
```diff
@@ -0,0 +1,484 @@
...
+            let lsb_y = (lsb * y_p.clone()) + one_minus_lsb * (y_p + base_y);
```
daira, line 117:
```
  lsb * y_p + (1 - lsb) * y_p + (1 - lsb) * base_y
= y_p + (1 - lsb) * base_y
```

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -188,7 +182,8 @@ pub mod tests {
                 layouter.namespace(|| "ScalarFixedShort"),
                 Some(scalar_fixed),
             )?;
-            value_commit_v.mul(layouter.namespace(|| "mul"), &scalar_fixed)?;
+            let result = value_commit_v.mul(layouter.namespace(|| "mul by zero"), &scalar_fixed)?;
+            result.constrain_equal(layouter.namespace(|| "[0]B = 𝒪"), &zero)?;
```
str4d, line 186:
```suggestion
            result.constrain_equal(layouter.namespace(|| "[0]B = 𝒪"), zero)?;
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -269,6 +270,7 @@ pub mod tests {
             chip.clone(),
             layouter.namespace(|| "commit_ivk_r"),
             commit_ivk_r,
+            &zero,
```
str4d, line 273:
```suggestion
            zero,
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -278,6 +280,7 @@ pub mod tests {
             chip.clone(),
             layouter.namespace(|| "note_commit_r"),
             note_commit_r,
+            &zero,
```
str4d, line 283:
```suggestion
            zero,
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -287,6 +290,7 @@ pub mod tests {
             chip.clone(),
             layouter.namespace(|| "nullifier_k"),
             nullifier_k,
+            &zero,
```
str4d, line 293:
```suggestion
            zero,
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -296,12 +300,18 @@ pub mod tests {
             chip.clone(),
             layouter.namespace(|| "value_commit_r"),
             value_commit_r,
+            &zero,
```
str4d, line 303:
```suggestion
            zero,
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -296,12 +300,18 @@ pub mod tests {
...
 
         // spend_auth_g
         let spend_auth_g = OrchardFixedBasesFull::SpendAuthG;
         let spend_auth_g = FixedPoint::from_inner(chip.clone(), spend_auth_g);
-        test_single_base(chip, layouter.namespace(|| "spend_auth_g"), spend_auth_g)?;
+        test_single_base(
+            chip,
+            layouter.namespace(|| "spend_auth_g"),
+            spend_auth_g,
+            &zero,
```
str4d, line 313:
```suggestion
            zero,
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -356,7 +367,9 @@ pub mod tests {
                 column,
                 Some(scalar_fixed),
             )?;
-            base.mul_base_field_elem(layouter.namespace(|| "mul by zero"), scalar_fixed)?;
+            let result =
+                base.mul_base_field_elem(layouter.namespace(|| "mul by zero"), scalar_fixed)?;
+            result.constrain_equal(layouter.namespace(|| "[0]B = 𝒪"), &zero)?;
```
str4d, line 372:
```suggestion
            result.constrain_equal(layouter.namespace(|| "[0]B = 𝒪"), zero)?;
```

src/circuit/gadget/ecc/chip/mul_fixed/full_width.rs
```diff
@@ -77,6 +78,7 @@ pub mod tests {
             chip.clone(),
             layouter.namespace(|| "commit_ivk_r"),
             commit_ivk_r,
+            &zero,
```
str4d, line 81:
```suggestion
            zero,
```

src/circuit/gadget/ecc/chip/mul_fixed/full_width.rs
```diff
@@ -86,6 +88,7 @@ pub mod tests {
             chip.clone(),
             layouter.namespace(|| "note_commit_r"),
             note_commit_r,
+            &zero,
```
str4d, line 91:
```suggestion
            zero,
```

src/circuit/gadget/ecc/chip/mul_fixed/full_width.rs
```diff
@@ -95,6 +98,7 @@ pub mod tests {
             chip.clone(),
             layouter.namespace(|| "nullifier_k"),
             nullifier_k,
+            &zero,
```
str4d, line 101:
```suggestion
            zero,
```

src/circuit/gadget/ecc/chip/mul_fixed/full_width.rs
```diff
@@ -104,12 +108,18 @@ pub mod tests {
             chip.clone(),
             layouter.namespace(|| "value_commit_r"),
             value_commit_r,
+            &zero,
```
str4d, line 111:
```suggestion
            zero,
```

src/circuit/gadget/ecc/chip/mul_fixed/full_width.rs
```diff
@@ -104,12 +108,18 @@ pub mod tests {
...
 
         // spend_auth_g
         let spend_auth_g = OrchardFixedBasesFull::SpendAuthG;
         let spend_auth_g = FixedPoint::from_inner(chip.clone(), spend_auth_g);
-        test_single_base(chip, layouter.namespace(|| "spend_auth_g"), spend_auth_g)?;
+        test_single_base(
+            chip,
+            layouter.namespace(|| "spend_auth_g"),
+            spend_auth_g,
+            &zero,
```
str4d, line 121:
```suggestion
            zero,
```

src/circuit/gadget/ecc/chip/mul_fixed/full_width.rs
```diff
@@ -166,7 +177,8 @@ pub mod tests {
                 layouter.namespace(|| "ScalarFixed"),
                 Some(scalar_fixed),
             )?;
-            base.mul(layouter.namespace(|| "mul by zero"), &scalar_fixed)?;
+            let result = base.mul(layouter.namespace(|| "mul by zero"), &scalar_fixed)?;
+            result.constrain_equal(layouter.namespace(|| "[0]B = 𝒪"), &zero)?;
```
str4d, line 181:
```suggestion
            result.constrain_equal(layouter.namespace(|| "[0]B = 𝒪"), zero)?;
```

str4d (commented):
Clippy lints

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,644 @@
...
+
+        // Decompose the base field element α into three-bit windows
+        // using a running sum `z`, where z_{i+1} = (z_i - a_i) / (2^3)
+        // for α = a_0 + (2^3) a_1 + ... + (2^{3(84)}) a_{84}.
+        //
+        // We set z_0 = α, which implies:
+        //      z_1 = (α - a_0) / 2^3, (subtract the lowest 3 bits)
+        //          = a_1 + (2^3) a_2 + ... + 2^{3(83)} a_{84},
+        //      z_2 = (z_1 - a_1) / 2^3
+        //          = a_2 + (2^3) a_3 + ... + 2^{3(82)} a_{84},
```
daira, line 67:
```suggestion
        // for α = a_0 + (2^3) a_1 + ... + (2^{3*84}) a_{84}.
        //
        // We set z_0 = α, which implies:
        //      z_1 = (α - a_0) / 2^3, (subtract the lowest 3 bits)
        //          = a_1 + (2^3) a_2 + ... + 2^{3*83} a_{84},
        //      z_2 = (z_1 - a_1) / 2^3
        //          = a_2 + (2^3) a_3 + ... + 2^{3*82} a_{84},
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,644 @@
...
+            );
+        }
+
+        config
+    }
+}
+
+impl Config {
+    pub fn create_gate(&self, meta: &mut ConstraintSystem<pallas::Base>) {
+        // Check that an expression is in the range [0..range),
```
daira, line 51:
```suggestion
        // Check that an expression is in the small range [0..range),
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,644 @@
...
+        // This gate also checks that this window uses the correct y_p and
+        // interpolated x_p.
+        meta.create_gate("Decompose base field element", |meta| {
+            let base_field_fixed_mul = meta.query_selector(self.base_field_fixed_mul);
+
+            let z_cur = meta.query_advice(self.super_config.window, Rotation::cur());
+            let z_next = meta.query_advice(self.super_config.window, Rotation::next());
+
+            //    z_{i+1} = (z_i - a_i) / (2^3)
+            // => a_i = z_i - (z_{i+1} * (2^3))
```
daira, line 85:
```suggestion
            //    z_{i+1} = (z_i - a_i) / (2^3)
            // => a_i = z_i - z_{i+1} * 2^3
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,644 @@
...
+        meta.create_gate("Canonicity checks", |meta| {
+            let base_field_fixed_canon = meta.query_selector(self.base_field_fixed_canon);
+
+            let alpha = meta.query_advice(self.canon_advices[0], Rotation::prev());
+            // z_85_alpha is constrained to be zero in this gate.
+            let z_85_alpha = meta.query_advice(self.canon_advices[1], Rotation::prev());
+            // The last three bits of α.
+            let z_84_alpha = meta.query_advice(self.canon_advices[2], Rotation::prev());
+
+            // Decompose α into three pieces,
```
daira, line 111:
```suggestion
            // Decompose α into three pieces, in little-endian order:
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,644 @@
...
+            };
+
+            // We want to enforce canonicity of a 255-bit base field element, α.
+            // That is, we want to check that 0 ≤ α < p, where p is Pallas base
+            // field modulus p = 2^254 + t_p
+            //                 = 2^254 + 45560315531419706090280762371685220353.
+            // Note that t_p < 2^130.
+            //
+            // α has been decomposed into three pieces,
+            //            α = α_0 (252 bits)  || α_1 (2 bits) || α_2 (1 bit).
```
daira, line 156:
```suggestion
            // α has been decomposed into three pieces in little-endian order:
            //            α = α_0 (252 bits) || α_1 (2 bits) || α_2 (1 bit)
            //              = α_0 + 2^{252} α_1 + 2^{254} α_2
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,644 @@
...
+            let z_13_alpha_0 = meta.query_advice(self.canon_advices[2], Rotation::next());
+
+            let decomposition_checks = {
+                // Range-constrain α_1 to be 2 bits
+                let alpha_1_range_check = range_check(alpha_1.clone(), 1 << 2);
+                // Boolean-constrain α_2
+                let alpha_2_range_check = range_check(alpha_2.clone(), 1 << 1);
+                // Check that α_1 + 2^2 α_2 = z_84_alpha
+                let z_84_alpha_check = z_84_alpha.clone()
+                    - (alpha_1 + alpha_2.clone() * pallas::Base::from_u64(1 << 2));
```
daira, line 129:
```suggestion
                    - (alpha_1.clone() + alpha_2.clone() * pallas::Base::from_u64(1 << 2));
```
(we end up needing `alpha_1` again below).

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,644 @@
...
+            //      - α_2 = 1 => α_1 = 0, and
+            //      - α_2 = 1 => α_0 < t_p. To enforce this:
+            //          - α_2 = 1 => 0 ≤ α_0 < 2^130
+            //                    => 13 ten-bit lookups of α_0
+            //          - α_2 = 1 => 0 ≤ α_0 + 2^130 - t_p < 2^130
+            //                    => 13 ten-bit lookups of α_0 + 2^130 - t_p
+            //                    => z_13_alpha_0_prime = 0
+            //
+            let canon_checks = {
+                std::iter::empty()
```
daira, line 168:
```suggestion
                std::iter::empty()
                    .chain(Some((
                        "MSB = 1 => alpha_1 = 0",
                        alpha_2.clone() * alpha_1,
                    )))
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,645 @@
...
+        )?;
+
+        // We want to enforce canonicity of a 255-bit base field element, α.
+        // That is, we want to check that 0 ≤ α < p, where p is Pallas base
+        // field modulus p = 2^254 + t_p
+        //                 = 2^254 + 45560315531419706090280762371685220353.
+        // Note that t_p < 2^130.
+        //
+        // α has been decomposed into three pieces,
+        //            α = α_0 (252 bits)  || α_1 (2 bits) || α_2 (1 bit).
```
daira, line 251:
```suggestion
            // α has been decomposed into three pieces in little-endian order:
            //            α = α_0 (252 bits) || α_1 (2 bits) || α_2 (1 bit)
            //              = α_0 + 2^{252} α_1 + 2^{254} α_2.
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,645 @@
...
+        //      z_1 = (α - a_0) / 2^3, (subtract the lowest 3 bits)
+        //          = a_1 + (2^3) a_2 + ... + 2^{3(83)} a_{84},
+        //      z_2 = (z_1 - a_1) / 2^3
+        //          = a_2 + (2^3) a_3 + ... + 2^{3(82)} a_{84},
+        //      ...,
+        //      z_{84} = a_{84}
+        //      z_n = (z_{84} - a_{84}) / 2^3
+        //          = 0.
+        //
+        // This gate checks that each a_i = z_i - z_{i+1} * (2^3) is within
```
daira, line 73:
```suggestion
        // for α = a_0 + 2^3 a_1 + ... + 2^{3*84} a_84.
        //
        // We set z_0 = α, which implies:
        //      z_1 = (α - a_0) / 2^3, (subtract the lowest 3 bits)
        //          = a_1 + 2^3 a_2 + ... + 2^{3*83} a_84,
        //      z_2 = (z_1 - a_1) / 2^3
        //          = a_2 + 2^3 a_3 + ... + 2^{3*82} a_84,
        //      ...,
        //      z_84 = a_84
        //      z_n = (z_84 - a_84) / 2^3
        //          = 0.
        //
        // This gate checks that each a_i = z_i - z_{i+1} * 2^3 is within
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,645 @@
...
+            let alpha_0 = alpha
+                .value()
+                .zip(z_84_alpha.value())
+                .map(|(alpha, z_84_alpha)| {
+                    let two_pow_252 = pallas::Base::from_u128(1 << 126).square();
+                    alpha - z_84_alpha * two_pow_252
+                });
+
+            let (alpha_0, zs) = self.lookup_config.witness_check(
+                layouter.namespace(|| "Lookup range check alpha + 2^130 - t_p"),
```
daira, line 277:
```suggestion
                layouter.namespace(|| "Lookup range check alpha"),
```
However, we should be able to use a similar trick to https://hackmd.io/o9EzZBwxSWSi08kQ_fMIOw?view#Overflow-check-for-variable-base-scalar-multiplication to enforce α ∊ [0, 2<sup>130</sup>) without an explicit 10-bit decomposition:

Note that 132 = 44 · 3. Also we have z<sub>44</sub> = a<sub>44</sub> + 2<sup>3</sup> a<sub>45</sub> + ... + 2<sup>3·(84-44)</sup> a<sub>84</sub>.
So the low 132 bits of α are zero exactly when α - 2<sup>44</sup> z<sub>44</sub> = 0. And then we can check that α ∊ [0, 2<sup>130</sup>) by checking α - 2<sup>44</sup> z<sub>44</sub> = 0 and a<sub>43</sub> ∊ {0, 1}.

src/circuit/gadget/utilities.rs
```diff
@@ -84,3 +86,27 @@ where
 
     Ok(CellValue::new(cell, copy.value))
 }
+
+/// Subsets a field element to a specified bitrange (little-endian)
+pub fn bitrange_subset<F: FieldExt + PrimeFieldBits>(
+    field_elem: F,
+    bitrange: std::ops::Range<usize>,
+) -> F {
```
str4d, line 94: Add some test cases for this.

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,645 @@
...
+                false,
+            )?;
+
+            (alpha_0, zs[13])
+        };
+
+        let (alpha_0_prime, z_13_alpha_0_prime) = {
+            // alpha_0_prime = alpha + 2^130 - t_p.
+            let alpha_0_prime = alpha_0.value().map(|alpha_0| {
+                let two_pow_130 = pallas::Base::from_u128(65).square();
```
str4d, line 289: Ah, this is probably why tests pass: the bug is duplicated here.
```suggestion
                let two_pow_130 = pallas::Base::from_u128(1 << 65).square();
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,645 @@
...
+    ) -> Result<EccPoint, Error> {
+        let (result, scalar) = layouter.assign_region(
+            || "Base-field elem fixed-base mul",
+            |mut region| {
+                let offset = 0;
+
+                // Decompose scalar
+                let scalar = self.decompose_base_field_elem(scalar, offset, &mut region)?;
+
+                let (acc, mul_b) = self.super_config.assign_region_inner(
```
str4d, line 202: How does this work? `decompose_base_field_elem` assigns a running sum into the `window` column, but `super_config.assign_region_inner` requires that this column contain the 3-bit words themselves, as it uses the cells directly to interpolate `x`.

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,645 @@
...
+        for (idx, word) in words.iter().enumerate() {
+            // z_next = (z_cur - word) / (2^3)
+            let z_next = {
+                let word = word.map(|word| pallas::Base::from_u64(word as u64));
+                let z_next_val = z
+                    .value()
+                    .zip(word)
+                    .map(|(z_cur_val, word)| (z_cur_val - word) * eight_inv);
+                let cell = region.assign_advice(
+                    || format!("word {:?}", idx),
```
str4d, line 477: This is assigning the running sum, not the words:
```suggestion
                    || format!("z_{:?}", idx + 1),
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,645 @@
...
+        // [a]B
+        {
+            let scalar_fixed = pallas::Base::rand();
+            let scalar_fixed = chip.load_private(
+                layouter.namespace(|| "witness random base field element"),
+                column,
+                Some(scalar_fixed),
+            )?;
+
+            base.mul_base_field_elem(layouter.namespace(|| "mul"), scalar_fixed)?;
```
str4d, line 594: Check the result of this (by witnessing the expected result).

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,645 @@
...
+                            acc * &h + &pallas::Base::from_u64(c.to_digit(8).unwrap().into())
+                        });
+
+            let scalar_fixed = chip.load_private(
+                layouter.namespace(|| "mul with double"),
+                column,
+                Some(scalar_fixed),
+            )?;
+
+            base.mul_base_field_elem(layouter.namespace(|| "mul with double"), scalar_fixed)?;
```
str4d, line 615: Check the result of this (by witnessing the expected result).

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,645 @@
...
+
+        // [-1]B is the largest base field element
+        {
+            let scalar_fixed = -pallas::Base::one();
+            let scalar_fixed = chip.load_private(
+                layouter.namespace(|| "mul by -1"),
+                column,
+                Some(scalar_fixed),
+            )?;
+            base.mul_base_field_elem(layouter.namespace(|| "mul by -1"), scalar_fixed)?;
```
str4d, line 640: Check the result of this (by witnessing the expected result).

str4d (changes requested):

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,583 @@
...
+            .iter()
+            .enumerate()
+        {
+            // Offset window index by 1 since we are starting on k_1
+            let w = w + 1;
+
+            // Compute [(k_w + 2) ⋅ 8^w]B
+            let mul_b = {
+                let mul_b = k.map(|k| {
+                    base_value * (k + pallas::Scalar::from_u64(2)) * h.pow(&[w as u64, 0, 0, 0])
```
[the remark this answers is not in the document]
daira: Alternatively, see zcash/pasta_curves#14

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -0,0 +1,253 @@
...
+                self.super_config
+                    .copy_scalar(&mut region, offset, &scalar.into())?;
+
+                let (acc, mul_b) = self.super_config.assign_region_inner(
+                    &mut region,
+                    offset,
+                    &scalar.into(),
+                    base.clone().into(),
+                    self.super_config.q_mul_fixed,
+                )?;
```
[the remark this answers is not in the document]
daira: I suspect also that making the 6 columns be on the right would help with layout, since we have a lot of regions that use columns on the left. But we can tweak things like that after we've merged the Action circuit PR.

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,645 @@
...
+}
+
+impl Config {
+    pub fn create_gate(&self, meta: &mut ConstraintSystem<pallas::Base>) {
+        // Check that an expression is in the range [0..range),
+        // i.e. 0 ≤ word < range.
+        let range_check = |word: Expression<pallas::Base>, range: usize| {
+            (0..range).fold(Expression::Constant(pallas::Base::one()), |acc, i| {
+                acc * (word.clone() - Expression::Constant(pallas::Base::from_u64(i as u64)))
+            })
```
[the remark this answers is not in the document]
daira: I saw that, but I thought that the Expression evaluator would merge constants and suppress multiplication by 1. If it doesn't, perhaps it should.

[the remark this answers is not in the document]
str4d: It does not.

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
therealyingtong: This works because the `mul_fixed::Config::coords_check()` creates the gate that checks interpolation, and this helper takes in an expression for `window`.

In `base_field_elem::Config::create_gate()`, we input the expression 
```
word = z_cur - z_next * pallas::Base::from_u64(constants::H as u64);
```
as the "window" to the `coords_check()` helper.

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,645 @@
...
+                let alpha_1_range_check = range_check(alpha_1.clone(), 1 << 2);
+                // Boolean-constrain α_2
+                let alpha_2_range_check = range_check(alpha_2.clone(), 1 << 1);
+                // Check that α_1 + 2^2 α_2 = z_84_alpha
+                let z_84_alpha_check = z_84_alpha.clone()
+                    - (alpha_1.clone() + alpha_2.clone() * pallas::Base::from_u64(1 << 2));
+
+                // Check that the witnessed α_0 fulfils the constraint α_0 = α - z_84_alpha * 2^252
+                let two_pow_252 = pallas::Base::from_u128(1 << 126).square();
+                let expected_alpha_0 = alpha - (z_84_alpha * two_pow_252);
```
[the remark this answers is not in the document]
therealyingtong: I've now documented this constraint in the book: 4af23e62dca4fbd28c67adc3b1248f94584ea868

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
therealyingtong: This optimisation makes sense to me, but I understand the details as being slightly different: the range check is on α<sub>0</sub> = α - 2<sup>3·(84)</sup> a<sub>84</sub> ∊ [0, 2<sup>130</sup>). 

To do this, we check:
- the high 120 bits of α<sub>0</sub> are all-zero, i.e. z<sub>44</sub> - 2<sup>3·(84 - 44)</sup> a<sub>84</sub> = 0. This constrains α<sub>0</sub> to be a 132-bit value.
- bits 130..=131 of α<sub>0</sub>, i.e. a<sub>43</sub> ∊ {0,1}. This constrains α<sub>0</sub> to be a 130-bit value.

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,583 @@
...
+        let base_u = base.u();
+        let scalar_windows_field = scalar.windows_field();
+        let scalar_windows_usize = scalar.windows_usize();
+
+        for (w, k) in scalar_windows_field[1..(scalar_windows_field.len() - 1)]
+            .iter()
+            .enumerate()
+        {
+            // Offset window index by 1 since we are starting on k_1
+            let w = w + 1;
```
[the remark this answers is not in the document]
therealyingtong: I've chosen the first alternative.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -449,35 +453,97 @@ pub mod tests {
...
+        fn constrain_equal<
+            EccChip: EccInstructions<pallas::Affine> + Clone + Eq + std::fmt::Debug,
+        >(
+            chip: EccChip,
+            mut layouter: impl Layouter<pallas::Base>,
+            base_val: pallas::Affine,
+            scalar_val: pallas::Base,
+            result: Point<pallas::Affine, EccChip>,
+        ) -> Result<(), Error> {
+            // Case scalar from base field into scalar field
```
str4d, line 467:
```suggestion
            // Move scalar from base field into scalar field (which always fits for Pallas).
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,679 @@
...
+            self.super_config.window,
+            offset,
+            &base_field_elem,
+            &self.super_config.perm,
+        )?;
+
+        for idx in 0..words.len() {
+            self.base_field_fixed_mul.enable(region, offset + idx)?;
+        }
+
```
str4d, line 462: This causes the selector to be double-enabled, as it is also activated by the call to `self.super_config.assign_region_inner` above:

image
```suggestion
```

str4d (approved):
utACK ae4e54dce8b789c43ce3170f83bac8ebc734e12c. Test circuit LGTM other than the unnecessary double-assignment of `base_field_fixed_mul`.

[EDIT: Removed old chip layout as I can't seem to hide review comments. See below for latest.]

src/circuit/gadget/ecc/chip/mul.rs
daira:
```suggestion
                    // Constrain the initialization of `z` to equal zero.
                    // TODO: use an equality constraint instead.
```

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,551 @@
...
+        let bitstring: Vec<bool> = {
+            let mut le_bytes = [0u8; 32];
+            k.to_little_endian(&mut le_bytes);
+            le_bytes.iter().fold(Vec::new(), |mut bitstring, byte| {
+                let bits = (0..8)
+                    .map(|shift| (byte >> shift) % 2 == 1)
+                    .collect::<Vec<_>>();
+                bitstring.extend_from_slice(&bits);
+                bitstring
+            })
```
daira, line 427: This should probably be a utility function. (Does not block.)

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -0,0 +1,551 @@
...
+                chip,
+                layouter.namespace(|| "[-1]B"),
+                p_val,
+                scalar_val,
+                result,
+            )?;
+        }
+
+        Ok(())
+    }
```
daira, line 550: Test the case where the last step (the complete addition) is a doubling. I think this occurs for the scalar 0x1b6db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6e; please check this.

src/circuit/gadget/ecc/chip/mul/complete.rs
```diff
@@ -0,0 +1,186 @@
...
+            |meta| {
+                let q_mul_decompose_var = meta.query_selector(self.q_mul_decompose_var);
+                // z_{i + 1}
+                let z_prev = meta.query_advice(self.z_complete, Rotation::prev());
+                // z_i
+                let z_next = meta.query_advice(self.z_complete, Rotation::next());
+
+                // k_{i} = z_{i} - 2⋅z_{i+1}
+                let k = z_next - Expression::Constant(pallas::Base::from_u64(2)) * z_prev;
+                let k_minus_one = k.clone() - Expression::Constant(pallas::Base::one());
```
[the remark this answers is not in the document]
daira: https://github.com/zcash/orchard/issues/143

src/circuit/gadget/ecc/chip/mul/complete.rs
```diff
@@ -0,0 +1,186 @@
...
+                // (k_i) ⋅ (k_i - 1) = 0
+                let bool_check = k.clone() * k_minus_one.clone();
```
daira, line 61:
```suggestion
                // (k_i)^2 - k_i = 0
                let bool_check = k.clone().square() - k.clone();
```

src/circuit/gadget/ecc/chip/mul/complete.rs
```diff
@@ -0,0 +1,186 @@
...
+
+                std::array::IntoIter::new([("bool_check", bool_check), ("y_switch", y_switch)])
+                    .map(move |(name, poly)| (name, q_mul_decompose_var.clone() * poly))
+            },
+        );
+    }
+
+    #[allow(clippy::type_complexity)]
+    #[allow(non_snake_case)]
+    #[allow(clippy::too_many_arguments)]
```
daira, line 80: Opinion: the default threshold of at most 7 arguments to avoid this lint is silly. We should raise that globally.

src/circuit/gadget/ecc/chip/mul/incomplete.rs
```diff
@@ -0,0 +1,419 @@
...
+    fn from(ecc_config: &EccConfig) -> Self {
+        let config = Config {
+            num_bits: INCOMPLETE_HI_RANGE.len(),
+            q_mul: ecc_config.q_mul_hi,
+            x_p: ecc_config.advices[0],
+            y_p: ecc_config.advices[1],
+            z: ecc_config.advices[9],
+            x_a: ecc_config.advices[3],
+            lambda1: ecc_config.advices[4],
+            lambda2: ecc_config.advices[5],
```
daira, line 50: What's the rationale for this ordering?

therealyingtong: The main rationale is that "hi_config z cannot overlap with complete addition columns", which is checked in an assertion in `mul.rs` (L86). Were it not for this requirement, we could have used a more natural ordering (e.g. [2], [3], [4], [5]).

therealyingtong: The rationale is to avoid an overlap with the output columns of complete addition, which are used to initialise the accumulator at 2[T]. This has now been explained at https://github.com/zcash/orchard/pull/111/commits/b696163e318d541f1327d851e2e9cb8f1b33287a.

src/circuit/gadget/ecc/chip/mul.rs
therealyingtong: I have checked that the last step is a doubling for 0x16db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6db6dc.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -80,15 +80,20 @@ impl From<&EccConfig> for Config {
...
+        // z and lambda1 are assigned on the same row as the add_config output.
+        // Therefore, z and lambda1 must not overlap with add_config.x_qr, add_config.y_qr.
+        let add_config_outputs = config.add_config.output_columns();
+        for config in [&(*config.hi_config), &(*config.lo_config)].iter() {
+            assert!(
+                !add_config_outputs.contains(&config.z),
+                "incomplete config z cannot overlap with complete addition columns."
+            );
+            assert!(
+                !add_config_outputs.contains(&config.lambda1),
```
str4d, line 93: This was previously checking `complete_config.z_complete`, and now it instead checks `{hi|lo}_config.lambda1`. Why is `z_complete` not necessary to check?

str4d (approved):
re-utACK b696163e318d541f1327d851e2e9cb8f1b33287a

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,529 @@
...
+                assert_eq!(NUM_WINDOWS, constants::NUM_WINDOWS_SHORT);
+                let base = ValueCommitV::get();
+                (
+                    base.lagrange_coeffs_short.0.as_ref().to_vec(),
+                    base.z_short.0.as_ref().to_vec(),
+                )
+            }
+            OrchardFixedBases::Full(base) => {
+                assert_eq!(NUM_WINDOWS, constants::NUM_WINDOWS);
+                let base: OrchardFixedBase = base.into();
```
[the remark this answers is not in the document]
daira: Why aren't we building the constants in a lazy_static?

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,550 @@
...
+                scalar.windows_usize()[NUM_WINDOWS - 1].map(|k| base.u()[NUM_WINDOWS - 1].0[k]);
+            region.assign_advice(
+                || "u",
+                self.u,
+                offset + NUM_WINDOWS - 1,
+                || u_val.ok_or(Error::SynthesisError),
+            )?;
+        }
+
+        // offset_acc = \sum_{j = 0}^{NUM_WINDOWS - 2} 2^{FIXED_BASE_WINDOW_SIZE * j+1}
```
daira, line 427:
```suggestion
        // offset_acc = \sum_{j = 0}^{NUM_WINDOWS - 2} 2^{FIXED_BASE_WINDOW_SIZE*j + 1}
```

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,550 @@
...
+        let offset_acc = (0..(NUM_WINDOWS - 1)).fold(pallas::Scalar::zero(), |acc, w| {
+            acc + (*TWO_SCALAR).pow(&[
+                constants::FIXED_BASE_WINDOW_SIZE as u64 * w as u64 + 1,
+                0,
+                0,
+                0,
+            ])
+        });
+
+        // `scalar = [k * 8^84 - offset_acc]`, where `offset_acc = \sum_{j = 0}^{83} 2^{FIXED_BASE_WINDOW_SIZE * j + 1}`.
```
daira, line 437:
```suggestion
        // `scalar = [k * 8^84 - offset_acc]`, where `offset_acc = \sum_{j = 0}^{83} 2^{FIXED_BASE_WINDOW_SIZE*j + 1}`.
```

src/circuit/gadget/ecc/chip/mul.rs
therealyingtong: The `z_complete` check was being duplicated. It was (and still is) checked in `mul::complete.rs`.

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -0,0 +1,550 @@
...
+
+    // The scalar decomposition is guaranteed to be in three-bit windows,
+    // so we also cast the least significant byte in their serialisation
+    // into usize for convenient indexing into `u`-values
+    fn windows_usize(&self) -> Vec<Option<usize>> {
+        self.windows_field()
+            .iter()
+            .map(|window| {
+                if let Some(window) = window {
+                    let window = window.to_bytes()[0] as usize;
```
daira, line 541:
```suggestion
                    let window = window.get_lower_32() as usize;
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -0,0 +1,676 @@
...
+        Ok(())
+    }
+
+    #[allow(clippy::op_ref)]
+    fn test_single_base(
+        chip: EccChip,
+        mut layouter: impl Layouter<pallas::Base>,
+        base: FixedPoint<pallas::Affine, EccChip>,
+        base_val: pallas::Affine,
+    ) -> Result<(), Error> {
```
daira, line 572: Why do these tests seem to be duplicated here and in `mul_fixed/full_width.rs`?

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -0,0 +1,317 @@
...
+                layouter.namespace(|| "[-2^64 - 1]B"),
+                base_val,
+                scalar_fixed_short,
+                result,
+            )?;
+        }
+
+        // There is a single canonical sequence of window values for which a doubling occurs on the last step:
+        // 1333333333333333333334 in octal.
+        // [0xB6DB_6DB6_DB6D_B6DC] B
```
daira, line 291: Test this value with both signs.

src/circuit/gadget/utilities.rs
```diff
@@ -97,3 +98,195 @@ pub fn transpose_option_array<T: Copy + std::fmt::Debug, const LEN: usize>(
     }
     ret
 }
+
+/// Subsets a field element to a specified bitrange (little-endian)
```
daira, line 102: It was unclear to me what this meant without looking at the implementation. I suggest something like:
```suggestion
/// Takes a specified subsequence of the little-endian bit representation of a field element.
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
therealyingtong: They're duplicated because the `mul_fixed()` and `mul_fixed_base_field_elem()` are both usable with the same set of fixed points. Perhaps we could abstract these tests out?

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
therealyingtong: Since these tests are refactored in #145, I have noted this comment there and will address it in that PR: https://github.com/zcash/orchard/pull/145/files#r670057072

src/constants.rs
```diff
@@ -20,6 +20,14 @@ pub mod util;
 
 pub use load::{OrchardFixedBase, OrchardFixedBasesFull, ValueCommitV};
 
+/// The Pallas scalar field modulus is $q = 2^{254} + \mathsf{t_q}$.
+/// <https://github.com/zcash/pasta>
+pub(crate) const T_Q: u128 = 45560315531506369815346746415080538113;
+
+/// The Pallas base field modulus is $p = 2^{254} + \mathsf{t_p}$.
+/// <https://github.com/zcash/pasta>
+pub(crate) const T_P: u128 = 45560315531419706090280762371685220353;
```
daira, line 29: I checked these values.

daira (approved):
utACK with minor comments.

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -531,14 +531,14 @@ impl ScalarFixed {
...
-    // so we also cast the least significant byte in their serialisation
+    // so we also cast the least significant 4 bytes in their serialisation
     // into usize for convenient indexing into `u`-values
     fn windows_usize(&self) -> Vec<Option<usize>> {
         self.windows_field()
             .iter()
             .map(|window| {
                 if let Some(window) = window {
-                    let window = window.to_bytes()[0] as usize;
+                    let window = window.get_lower_32() as usize;
```
str4d, line 541: @daira left a comment inside the `get_lower_32()` impls:
> // TODO: don't reduce, just hash the Montgomery form. (Requires rebuilding perfect hash table.)

We'll need to take care over there to ensure we don't alter this API, now that we are relying on it outside the perfect hash table.

str4d (approved):
re-utACK 425ee6e038d4104cbbb0804c76714ab067b1d8f5. Test chip LGTM.

ecc-chip-layout

PR #112, opened 2021-06-11
[book] commitment-tree.md: Update Uncommitted^Orchard from 0 -> 2.

therealyingtong:
cf. https://zips.z.cash/protocol/protocol.pdf#thmuncommittedorchard

### DIFF
--- a/book/src/design/commitment-tree.md
+++ b/book/src/design/commitment-tree.md
@@ -26,8 +26,8 @@ appended to the tree as a regular leaf.
 
 Orchard note commitments are the $x$-coordinates of Pallas points; thus we take the same
 approach as Sapling, using a value that is not the $x$-coordinate of any Pallas point as the
-uncommitted leaf value. It happens that $0$ is the smallest such value for both Pallas and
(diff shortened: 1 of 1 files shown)


str4d (approved):
ACK. I verified the Sage output is still correct.

book/src/design/commitment-tree.md
```diff
@@ -36,12 +36,17 @@ sage: EllipticCurve(GF(p), [0, 5]).count_points() == q
...
+sage: Mod(13, p).is_square()
 False
-sage: Mod(5, q).is_square()
+sage: Mod(13, q).is_square()
 False
 ```
 
+> Note: There are also no Pallas points with $x$-coordinate $0$, but we map the identity to
+> $(0, 0)$ within the circuit. Although $\mathsf{SinsemillaCommit}$ cannot return the identity
+> (the incomplete addition would return $\perp$ instead), it would arguably be confusing to
```
daira, line 47: Why $\perp$ rather than $\bot$? (The latter is what the spec uses, and reflects the meaning of the symbol as denoting the bottom element of a lattice, rather than its shape.)

PR #114, opened 2021-06-12
utilities::lookup_range_check: Add LookupRangeCheck helper

therealyingtong:
This decomposes a field element into K-bit words and constrains each word's range by looking it up in a K-bit lookup table.

The field element is broken down using a running sum. All interstitial values of the running sum are returned.

### DIFF
--- a/src/circuit/gadget/utilities/lookup_range_check.rs
+++ b/src/circuit/gadget/utilities/lookup_range_check.rs
@@ -23,7 +23,7 @@ pub struct LookupRangeCheckConfig<F: FieldExt + PrimeFieldBits, const K: usize> 
 }
 
 impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K> {
-    /// The `q_lookup` column toggles the lookup on or off. It can be assigned
+    /// The `q_lookup` column toggles the lookup on or off. It MUST be assigned
     /// outside of this helper at the appropriate offsets.
     ///
     /// The `running_sum` advice column breaks the field element into `K`-bit
@@ -85,13 +85,13 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K> 
         )
     }
 
-    // Only the lower `num_words * K` bits of the field element are constrained
-    // by this function. If the field element does not fit into this range, then
-    // the final cumulative sum `z_{num_words}` will be nonzero.
+    /// Only the lower `num_words * K` bits of the field element are constrained
+    /// by this function. If the field element does not fit into this range, then
+    /// the final cumulative sum `z_{num_words}` will be nonzero.
     //
-    // It is up to the caller to constrain `z_{num_words}` == 0` outside this
-    // helper, or otherwise constrain upper bits not covered within the
-    // `num_words * K` range.
+    /// It is up to the caller to constrain `z_{num_words}` == 0` outside this
+    /// helper, or otherwise constrain upper bits not covered within the
+    /// `num_words * K` range.
     pub fn lookup_range_check(
         &self,
         region: &mut Region<'_, F>,
@@ -100,29 +100,30 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K> 
         num_words: usize,
     ) -> Result<Vec<CellValue<F>>, Error> {
         // `num_words` must fit into a single field element.
-        assert!(num_words <= F::NUM_BITS as usize / K);
+        assert!(num_words * K <= F::CAPACITY as usize);
         let num_bits = num_words * K;
 
-        // Take first num_bits bits of `element`.
-        let bits = element.value().map(|element| {
-            element
-                .to_le_bits()
-                .into_iter()
-                .take(num_bits)
-                .collect::<Vec<_>>()
-        });
-
         // Chunk the first num_bits bits into K-bit words.
-        let bits: Option<Vec<F>> = bits.map(|bits| {
-            bits.chunks_exact(K)
-                .map(|word| F::from_u64(lebs2ip::<K>(&(word.try_into().unwrap()))))
-                .collect::<Vec<_>>()
-        });
+        let words = {
+            // Take first num_bits bits of `element`.
+            let bits = element.value().map(|element| {
+                element
+                    .to_le_bits()
+                    .into_iter()
+                    .take(num_bits)
+                    .collect::<Vec<_>>()
+            });
 
-        let bits = if let Some(bits) = bits {
-            bits.into_iter().map(Some).collect()
-        } else {
-            vec![None; num_words]
+            let words: Option<Vec<F>> = bits.map(|bits| {
+                bits.chunks_exact(K)
+                    .map(|word| F::from_u64(lebs2ip::<K>(&(word.try_into().unwrap()))))
+                    .collect::<Vec<_>>()
+            });
+            if let Some(words) = words {
+                words.into_iter().map(Some).collect()
+            } else {
+                vec![None; num_words]
+            }
         };
 
         // Copy `element` and initialize running sum `z_0 = element` to decompose it.
@@ -144,14 +145,14 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K> 
         // For `element` = a_0 + 2^10 a_1 + ... + 2^{120} a_{12}}, initialize z_0 = `element`.
         // If `element` fits in 130 bits, we end up with z_{13} = 0.
         let mut z = z_0;
-        let inv_2_pow_k = F::from_u64(1u64 << K).invert().unwrap();
-        for (idx, word) in bits.into_iter().enumerate() {
+        let inv_two_pow_k = F::from_u64(1u64 << K).invert().unwrap();
+        for (idx, word) in words.into_iter().enumerate() {
             // z_next = (z_cur - m_cur) / 2^K
             z = {
                 let z_val = z
                     .value()
                     .zip(word)
-                    .map(|(z, word)| (z - word) * inv_2_pow_k);
+                    .map(|(z, word)| (z - word) * inv_two_pow_k);
 
                 // Assign z_next
                 let z_cell = region.assign_advice(
--- a/src/circuit/gadget/utilities/lookup_range_check.rs
+++ b/src/circuit/gadget/utilities/lookup_range_check.rs
@@ -3,8 +3,8 @@
 
 use crate::primitives::sinsemilla::lebs2ip_k;
 use halo2::{
-    circuit::{Chip, Layouter},
-    plonk::{Advice, Column, ConstraintSystem, Error, Fixed, Permutation, Selector},
+    circuit::{Layouter, Region},
+    plonk::{Advice, Column, ConstraintSystem, Error, Fixed, Permutation},
     poly::Rotation,
 };
 use std::marker::PhantomData;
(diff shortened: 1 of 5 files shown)


src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,297 @@
...
+        let mut zs = vec![z_0];
+
+        // Assign cumulative sum such that
+        //          z_i = 2^{K}⋅z_{i + 1} + a_i
+        // => z_{i + 1} = (z_i - a_i) / (2^K)
+        //
+        // For `words` = a_0 + 2^10 a_1 + ... + 2^{120} a_{12}}, initialize z_0 = `words`.
+        // If `words` fits in 130 bits, we end up with z_{13} = 0.
+        let mut z = z_0;
+        let inv_2_pow_k = F::from_u64(1u64 << K).invert().unwrap();
```
str4d, line 126: This single inversion per decomposed element is probably negligible, but is there a reason why the region decomposes the element in little-endian word order? The relation would be functionally identical if the words were constrained in reverse, but I don't know how you intend to use the returned `zs`. The intended usage pattern should be documented.

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,297 @@
...
+        )?;
+
+        let mut zs = vec![z_0];
+
+        // Assign cumulative sum such that
+        //          z_i = 2^{K}⋅z_{i + 1} + a_i
+        // => z_{i + 1} = (z_i - a_i) / (2^K)
+        //
+        // For `words` = a_0 + 2^10 a_1 + ... + 2^{120} a_{12}}, initialize z_0 = `words`.
+        // If `words` fits in 130 bits, we end up with z_{13} = 0.
```
str4d, line 124: Document the `lookup_range_check` function to include this API limitation: only the lower `num_words * K` bits are constrained by this function, and it is up to the caller to constrain `z_{num_words} == 0` (or otherwise constrain the upper bits).

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,297 @@
...
+        let bits = if let Some(bits) = bits {
+            bits.into_iter().map(Some).collect()
+        } else {
+            vec![None; num_words]
+        };
+
+        // Copy `words` and initialize running sum `z_0 = words` to decompose it.
+        let z_0 = copy(
+            region,
+            || "copy words",
```
str4d, line 110:
```suggestion
            || "z_0",
```

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,297 @@
...
+                        |mut region| {
+                            for idx in 0..num_words {
+                                // Assign fixed column to activate lookup.
+                                region.assign_fixed(
+                                    || format!("lookup on row {}", idx),
+                                    config.q_lookup,
+                                    idx,
+                                    || Ok(F::one()),
+                                )?;
+                            }
```
str4d, line 237: Why can't this be activated inside `lookup_range_check` (inside the `for (idx, word) in` loop, activated at `offset + idx`)? If it can't be, document that behaviour on `lookup_range_check`.

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,297 @@
...
+        let chunks = {
+            words
+                .to_le_bits()
+                .iter()
+                .by_val()
+                .take(num_words * K)
+                .collect::<Vec<_>>()
+                .chunks_exact(K)
+                .collect::<Vec<_>>()
+                .iter()
```
str4d, line 279: I think this intermediate collection is unnecessary, as `chunks_exact` returns an iterator of slices that can be passed to `lebs2ip`:
```suggestion
```

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,297 @@
...
+                                    idx,
+                                    || Ok(F::one()),
+                                )?;
+                            }
+
+                            config.lookup_range_check(&mut region, 0, words, num_words)
+                        },
+                    )?;
+
+                    assert_eq!(expected_zs[expected_zs.len() - 1], *expected_final_z);
```
str4d, line 243:
```suggestion
                    assert_eq!(expected_zs.last().unwrap(), *expected_final_z);
```

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,297 @@
...
+                cs: &mut impl Assignment<F>,
+                config: Self::Config,
+            ) -> Result<(), Error> {
+                let mut layouter = SingleChipLayouter::new(cs)?;
+
+                // Load table_idx
+                config.load(&mut layouter)?;
+
+                let num_words = 6;
+                let words_and_expected_final_zs = [
```
str4d, line 211: Similarly rename these throughout to refer to "elements" instead of "words" for clarity:
```suggestion
                let elements_and_expected_final_zs = [
```

str4d (changes requested):

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,297 @@
...
+            let word = z_cur - z_next * F::from_u64(1 << K);
+            let table = meta.query_fixed(config.table_idx, Rotation::cur());
+
+            vec![(q_lookup * word, table)]
+        });
+
+        config
+    }
+
+    pub fn load(&self, layouter: &mut impl Layouter<F>) -> Result<(), Error> {
```
therealyingtong, line 56: This lookup table is provided by the Sinsemilla chip in the context of Orchard, but may not be in general.

```suggestion
    #[cfg(test)]
    fn load(&self, layouter: &mut impl Layouter<F>) -> Result<(), Error> {
```

src/circuit/gadget/utilities/lookup_range_check.rs
therealyingtong: We could constrain the decomposition in reverse, but I prefer little-endian order for conceptual consistency. This helper subtracts the lowest `K` bits from the element with each iteration. e.g. This is used in the overflow check for variable-base scalar mul: to decompose as 127-bit element, we use this helper for the first 120 bits, and then pick up where it left off to manually subtract and range-check the remaining 7 bits.

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,319 @@
...
+pub struct LookupRangeCheckConfig<F: FieldExt + PrimeFieldBits, const K: usize> {
+    q_lookup: Column<Fixed>,
+    running_sum: Column<Advice>,
+    table_idx: Column<Fixed>,
+    perm: Permutation,
+    _marker: PhantomData<F>,
+}
+
+impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K> {
+    /// The `q_lookup` column toggles the lookup on or off. It can be assigned
```
str4d, line 26:
```suggestion
    /// The `q_lookup` column toggles the lookup on or off. It MUST be assigned
```

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,319 @@
...
+        let bits = element.value().map(|element| {
+            element
+                .to_le_bits()
+                .into_iter()
+                .take(num_bits)
+                .collect::<Vec<_>>()
+        });
+
+        // Chunk the first num_bits bits into K-bit words.
+        let bits: Option<Vec<F>> = bits.map(|bits| {
```
str4d, line 116:
```suggestion
        let words: Option<Vec<F>> = bits.map(|bits| {
```

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,319 @@
...
+            bits.chunks_exact(K)
+                .map(|word| F::from_u64(lebs2ip::<K>(&(word.try_into().unwrap()))))
+                .collect::<Vec<_>>()
+        });
+
+        let bits = if let Some(bits) = bits {
+            bits.into_iter().map(Some).collect()
```
str4d, line 123:
```suggestion
        let words = if let Some(words) = words {
            words.into_iter().map(Some).collect()
```

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,319 @@
...
+
+        // Assign cumulative sum such that
+        //          z_i = 2^{K}⋅z_{i + 1} + a_i
+        // => z_{i + 1} = (z_i - a_i) / (2^K)
+        //
+        // For `element` = a_0 + 2^10 a_1 + ... + 2^{120} a_{12}}, initialize z_0 = `element`.
+        // If `element` fits in 130 bits, we end up with z_{13} = 0.
+        let mut z = z_0;
+        let inv_2_pow_k = F::from_u64(1u64 << K).invert().unwrap();
+        for (idx, word) in bits.into_iter().enumerate() {
```
str4d, line 148:
```suggestion
        for (idx, word) in words.into_iter().enumerate() {
```

str4d (approved):
ACK. Layout for the test circuit LGTM.

lookup-range-check

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,319 @@
...
+        )
+    }
+
+    // Only the lower `num_words * K` bits of the field element are constrained
+    // by this function. If the field element does not fit into this range, then
+    // the final cumulative sum `z_{num_words}` will be nonzero.
+    //
+    // It is up to the caller to constrain `z_{num_words}` == 0` outside this
+    // helper, or otherwise constrain upper bits not covered within the
+    // `num_words * K` range.
```
ebfull, line 94: These should be doc comments (`///`)

src/circuit/gadget/utilities/lookup_range_check.rs
```diff
@@ -0,0 +1,319 @@
...
+    pub fn lookup_range_check(
+        &self,
+        region: &mut Region<'_, F>,
+        offset: usize,
+        element: CellValue<F>,
+        num_words: usize,
+    ) -> Result<Vec<CellValue<F>>, Error> {
+        // `num_words` must fit into a single field element.
+        assert!(num_words <= F::NUM_BITS as usize / K);
```
ebfull, line 103: `num_words * K <= F::NUM_BITS`?

ebfull: Actually, it should be _strictly less than_ right? aka `F::CAPACITY`?

src/spec.rs
```diff
@@ -179,6 +179,18 @@ pub(crate) fn extract_p_bottom(point: CtOption<pallas::Point>) -> CtOption<palla
     point.map(|p| extract_p(&p))
 }
 
+/// The u64 integer represented by an L-bit little-endian bitstring.
+///
+/// Panics
```
ebfull, line 184:
```suggestion
/// # Panics
```

src/circuit/gadget/utilities/lookup_range_check.rs
ebfull: Let's make an issue for revisiting making this inversion into a constant somewhere.

ebfull (changes requested):

src/circuit/gadget/utilities/lookup_range_check.rs
therealyingtong: my intention is: `K = 10`, `F::NUM_BITS = 255`, `num_words <= 255 / 10 = 25`

src/circuit/gadget/utilities/lookup_range_check.rs
therealyingtong: I've added `INV_TWO_POW_K` to `primitives::sinsemilla::constants.rs`.

ebfull (approved):

PR #115, opened 2021-06-12
Store anchors as pallas::Base instead of [u8; 32]

str4d:
This matches what we store in `MerklePath`, and better enforces the required type.


daira (approved):
utACK

PR #117, opened 2021-06-14
Use patches for halo2 and zcash_note_encryption dependencies

str4d:
While these two are in flux, it's hard to keep these revisions consistent
(e.g. https://github.com/zcash/zcash/pull/5217 currently depends on two
different versions of zcash_note_encryption). Using patches allows the
downstream users (i.e. zcashd) to define its own set of patches, and keep
everything in sync. This works fine now because we aren't actively making
changes to the public APIs, only additions.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,27 +26,21 @@ blake2b_simd = "0.5"
 ff = "0.10"
 fpe = "0.4"
 group = "0.10"
+halo2 = "0.0"
 lazy_static = "1"
 pasta_curves = "0.1"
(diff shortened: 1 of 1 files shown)


str4d:
Turns out this is required, because we can't vendor both `zcash_note_encryption` versions without a bunch of workaround: https://ci.z.cash/#builders/1/builds/669

str4d:
Doesn't touch the code, doesn't alter the dependency versions. Self-ACK to unblock the zcashd PR.

ebfull:
post-hoc ACK

PR #118, opened 2021-06-15
Sinsemilla chip with Commit Domain

therealyingtong:


src/circuit/gadget/sinsemilla.rs
```diff
@@ -222,6 +226,7 @@ pub struct HashDomain<
     EccChip: EccInstructions<
             C,
             Point = <SinsemillaChip as SinsemillaInstructions<C, K, MAX_WORDS>>::Point,
+            FixedPoints = <SinsemillaChip as SinsemillaInstructions<C, K, MAX_WORDS>>::FixedPoints,
```
str4d, line 229: Do we want these to be identical? I agree that `EccChip::FixedPoints` needs to contain all fixed points we want to use with `CommitDomain`, but ideally it could be a superset.

Non-blocking because we do have the `CommitDomains` trait which means someone can't accidentally use a non-SinsemillaCommit fixed point, and I can imagine the type restrictions to make this a superset could get somewhat ugly.

src/circuit/gadget/sinsemilla.rs
```diff
@@ -573,6 +572,29 @@ mod tests {
...
+                    Some(pallas::Scalar::rand()),
+                )?;
+                let message: Vec<Option<bool>> =
+                    (0..500).map(|_| Some(rand::random::<bool>())).collect();
+                let message = Message::from_bitstring(
+                    chip2,
+                    layouter.namespace(|| "witness message"),
+                    message,
+                )?;
+                commit_ivk.commit(layouter.namespace(|| "commit"), message, r)?;
```
str4d, line 595: Check the result with a constraint.

str4d (approved):
utACK after the test is updated.

str4d (changes requested):
`ecc-chip-layout.png` was accidentally committed.

str4d:
Squash the last commit so the binary PNG doesn't exist in the git history.

src/circuit/gadget/sinsemilla.rs
```diff
@@ -284,12 +290,99 @@ where
...
+    ) -> Self {
+        CommitDomain {
+            M: HashDomain::new(sinsemilla_chip, ecc_chip.clone(), &domain.hash_domain()),
+            R: ecc::FixedPoint::from_inner(ecc_chip, domain.r()),
+        }
+    }
+
+    /// $\mathsf{SinsemillaCommit}$ from [§ 5.4.8.4][concretesinsemillacommit].
+    ///
+    /// [concretesinsemillacommit]: https://zips.z.cash/protocol/nu5.pdf#concretesinsemillacommit
```
str4d, line 357: This can be renamed to `protocol.pdf`, but don't make that change in this PR (we can do a repo-wide rename once everything has stabilised).

src/circuit/gadget/sinsemilla.rs
```diff
@@ -284,37 +290,123 @@ where
     }
 }
 
+/// Trait allowing circuit's Sinsemilla CommitDomains to be enumerated.
+pub trait CommitDomains<C: CurveAffine, F: Clone + Debug, H: HashDomains<C>>:
+    Clone + Debug
+{
+    /// Returns the fixed point corresponding to the R constant for this CommitDomain.
```
daira, line 297:
```suggestion
    /// Returns the fixed point used for randomization in this CommitDomain.
```

src/circuit/gadget/sinsemilla.rs
```diff
@@ -284,37 +290,123 @@ where
...
+    ) -> Self {
+        CommitDomain {
+            M: HashDomain::new(sinsemilla_chip, ecc_chip.clone(), &domain.hash_domain()),
+            R: ecc::FixedPoint::from_inner(ecc_chip, domain.r()),
+        }
+    }
+
+    /// $\mathsf{SinsemillaCommit}$ from [§ 5.4.8.4][concretesinsemillacommit].
+    ///
+    /// [concretesinsemillacommit]: https://zips.z.cash/protocol/nu5.pdf#concretesinsemillacommit
```
daira, line 357:
```suggestion
    /// [concretesinsemillacommit]: https://zips.z.cash/protocol/protocol.pdf#concretesinsemillacommit
```

str4d (approved):
utACK a17a9301d7b58c27c4fc8ce663c1745b3d1c4000

src/circuit/gadget/sinsemilla.rs
```diff
@@ -284,37 +290,123 @@ where
...
+        assert_eq!(self.M.sinsemilla_chip, message.chip);
+        let (blind, _) = self.R.mul(layouter.namespace(|| "[r] R"), r)?;
+        self.M
+            .hash_to_point(layouter.namespace(|| "M"), message)?
+            .add_incomplete(layouter.namespace(|| "M ⸭ [r] R"), &blind)
+    }
+
+    /// $\mathsf{SinsemillaShortCommit}$ from [§ 5.4.8.4][concretesinsemillacommit].
+    ///
+    /// [concretesinsemillacommit]: https://zips.z.cash/protocol/nu5.pdf#concretesinsemillacommit
```
daira, line 373:
```suggestion
    /// [concretesinsemillacommit]: https://zips.z.cash/protocol/protocol.pdf#concretesinsemillacommit
```

daira (approved):
utACK

daira (approved):
re-utACK f44c4161af35c0c27b3ea81a00cd6e72ad467603

PR #119, opened 2021-06-15
Anchor serialization

str4d:
Necessary because I made the internals of `Anchor` opaque in #115.

### DIFF
--- a/src/tree.rs
+++ b/src/tree.rs
@@ -7,7 +7,7 @@ use crate::{
 };
 use pasta_curves::pallas;
 
-use ff::{Field, PrimeFieldBits};
+use ff::{Field, PrimeField, PrimeFieldBits};
 use rand::RngCore;
 use std::iter;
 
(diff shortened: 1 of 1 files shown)


nuttycom (approved):
utACK

src/tree.rs
```diff
@@ -21,6 +21,18 @@ impl From<pallas::Base> for Anchor {
...
 }
 
+impl Anchor {
+    /// Parses an Orchard anchor from a byte encoding.
+    pub fn from_bytes(bytes: [u8; 32]) -> Option<Anchor> {
+        pallas::Base::from_repr(bytes).map(Anchor)
+    }
+
+    /// Returns the byte encoding of this anchor.
+    pub fn to_bytes(self) -> [u8; 32] {
```
nuttycom, line 31: On second thought, I don't actually understand why this has to consume `self`?

PR #121, opened 2021-06-16
Simplify the canonicity check for variable-base scalar multiplication

daira:
Range-check the low 130 bits rather than the low 127 bits.

Signed-off-by: Daira Hopwood <daira@jacaranda.org>

### DIFF
--- a/src/circuit/gadget/ecc/chip/mul/overflow.rs
+++ b/src/circuit/gadget/ecc/chip/mul/overflow.rs
@@ -62,7 +62,7 @@ impl Config {
             let k_254 = meta.query_advice(self.advices[1], Rotation::prev());
             let alpha = meta.query_advice(self.advices[1], Rotation::cur());
 
-            // s_minus_lo_130 = s - sum_{i = 0}^{130} 2^i ⋅ s_i
+            // s_minus_lo_130 = s - sum_{i = 0}^{129} 2^i ⋅ s_i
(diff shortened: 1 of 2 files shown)


src/circuit/gadget/ecc/chip/mul/overflow.rs
```diff
@@ -57,80 +51,49 @@ impl Config {
...
-            let z_127 = meta.query_advice(self.advices[0], Rotation::cur());
+            let z_130 = meta.query_advice(self.advices[0], Rotation::cur());
             let eta = meta.query_advice(self.advices[0], Rotation::next());
 
             let k_254 = meta.query_advice(self.advices[1], Rotation::prev());
             let alpha = meta.query_advice(self.advices[1], Rotation::cur());
 
-            // s_minus_lo_127 = s - sum_{i = 0}^{126} 2^i ⋅ s_i
-            let s_minus_lo_127 = meta.query_advice(self.advices[1], Rotation::next());
+            // s_minus_lo_130 = s - sum_{i = 0}^{130} 2^i ⋅ s_i
```
therealyingtong, line 65: Bits are zero-indexed.

```suggestion
            // s_minus_lo_130 = s - sum_{i = 0}^{129} 2^i ⋅ s_i
```

therealyingtong (approved):
Tested ACK.

daira:
This will be re-reviewed as part of #121, and it simplifies that PR, so it doesn't need more than one review here.

PR #132, opened 2021-06-25
Add short range check to `lookup_range_check` util

therealyingtong:
This PR introduces a range check for field elements <= `K` bits, using two lookups in a `K`-bit table.

This also introduces a `strict` mode for the full-length lookup (for field elements greater than `K` bits) to constrain the final running sum to 0.

PR #133, opened 2021-06-25
Introduce `LookupRangeCheckConfig`s for each Sinsemilla advice column

therealyingtong:
Based on #132.

Inputs to Sinsemilla are commonly decomposed into subpieces and packed into field elements. These subpieces have to be range-checked. This PR includes a `LookupRangeCheckConfig` for each Sinsemilla advice column, to enable subpieces to be looked up in any advice column.

This PR also moves certain APIs from the instruction level to the gadget level.

### DIFF
--- a/src/circuit/gadget/sinsemilla.rs
+++ b/src/circuit/gadget/sinsemilla.rs
@@ -3,8 +3,10 @@ use crate::circuit::gadget::{
     ecc::{self, EccInstructions},
     utilities::Var,
 };
-use halo2::{arithmetic::CurveAffine, circuit::Layouter, plonk::Error};
-use std::fmt::Debug;
+use ff::PrimeField;
+use halo2::{circuit::Layouter, plonk::Error};
+use pasta_curves::arithmetic::{CurveAffine, FieldExt};
+use std::{convert::TryInto, fmt::Debug};
 
 pub mod chip;
 mod message;
@@ -27,7 +29,7 @@ pub trait SinsemillaInstructions<C: CurveAffine, const K: usize, const MAX_WORDS
     ///
     /// For example, in the case `K = 10`, `NUM_BITS = 255`, we can fit
     /// up to `N = 25` words in a single base field element.
-    type MessagePiece;
+    type MessagePiece: Clone + Debug;
 
     /// The x-coordinate of a point output of [`Self::hash_to_point`].
     type X;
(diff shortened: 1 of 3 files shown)


src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -60,13 +60,19 @@ pub struct SinsemillaConfig {
...
     /// generators of the Sinsemilla hash.
-    generator_table: GeneratorTableConfig,
+    pub(super) generator_table: GeneratorTableConfig,
     /// Fixed column shared by the whole circuit. This is used to load the
     /// x-coordinate of the domain $Q$, which is then constrained to equal the
     /// initial $x_a$.
     constants: Column<Fixed>,
-    /// Permutation over all advice columns and the `constants` fixed column.
-    perm: Permutation,
+    // Permutation over all advice columns and the `constants` fixed column.
```
str4d: Rebase mis-merge:
```suggestion
    /// Permutation over all advice columns and the `constants` fixed column.
```
EDIT: Ah, this is fixed in the (unrelated) third commit. Move that into this commit if you want.

src/circuit/gadget/sinsemilla.rs
```diff
@@ -130,24 +101,112 @@ where
...
+    fn from_pieces(
+        chip: SinsemillaChip,
+        pieces: Vec<MessagePiece<C, SinsemillaChip, K, MAX_WORDS>>,
+    ) -> Self {
         Self {
             chip,
-            inner: pieces.into(),
+            inner: pieces
+                .iter()
+                .map(|piece| piece.inner.clone())
```
str4d, line 143:
```suggestion
                .into_iter()
                .map(|piece| piece.inner)
```

src/circuit/gadget/sinsemilla.rs
```diff
@@ -337,12 +337,32 @@ mod tests {
...
+            ];
+            let constants_2 = [
+                meta.fixed_column(),
+                meta.fixed_column(),
+                meta.fixed_column(),
+                meta.fixed_column(),
+                meta.fixed_column(),
+                meta.fixed_column(),
+            ];
+            let constants_3 = meta.fixed_column();
```
str4d, line 357: Where is this used?

str4d (approved):
utACK modulo some mis-merges and minor comments.

daira (approved):
utACK

src/circuit/gadget/sinsemilla.rs
therealyingtong: This isn't used in the current PR and has been removed. However, after we merge the changes to `LookupRangeCheckConfig`, this will be used to configure the ECC chip.

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -60,13 +60,19 @@ pub struct SinsemillaConfig {
...
+    pub(super) perm: Permutation,
+}
+
+impl SinsemillaConfig {
+    pub fn advices(&self) -> [Column<Advice>; 5] {
+        [self.bits, self.lambda_1, self.lambda_2, self.x_a, self.x_p]
```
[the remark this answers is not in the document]
therealyingtong: I actually don't think the `advices()` method is needed anymore. It was only being used to construct `lookup_config`s, which have now been made native to the `SinsemillaConfig`.

PR #135, opened 2021-06-30
Partial fixes from NCC draft report

therealyingtong:
Minor fixes involving constants:
- document that `find_zs_and_us()` is not meant to be used anywhere;
- use `F::zero()` instead of `F::default()` in `constants/util.rs`;
- use personalisations from `constants` in `spec.rs`.

In `primitives/sinsemilla.rs`: document that `hash()` panics if the message is longer than `K * C`.

### DIFF
--- a/src/constants/util.rs
+++ b/src/constants/util.rs
@@ -37,7 +37,9 @@ pub fn evaluate<C: CurveAffine>(x: u8, coeffs: &[C::Base]) -> C::Base {
     coeffs
         .iter()
         .rev()
-        .fold(C::Base::zero(), |acc, coeff| acc * x + coeff)
+        .cloned()
+        .reduce(|acc, coeff| acc * x + coeff)
+        .unwrap_or_else(C::Base::zero)
 }
 
 /// Takes in an FnMut closure and returns a constant-length array with elements of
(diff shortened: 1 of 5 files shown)


src/constants/util.rs
```diff
@@ -37,7 +37,7 @@ pub fn evaluate<C: CurveAffine>(x: u8, coeffs: &[C::Base]) -> C::Base {
     coeffs
         .iter()
         .rev()
-        .fold(C::Base::default(), |acc, coeff| acc * x + coeff)
+        .fold(C::Base::zero(), |acc, coeff| acc * x + coeff)
```
str4d, line 40: We can also use `reduce` here (which initializes `acc` with the first element of the iterator) to avoid a multiplication:
```suggestion
        .reduce(|acc, coeff| acc * x + coeff)
        .unwrap_or_else(C::Base::zero)
```

str4d (approved):

str4d (approved):
re-utACK 32ea1cce27bed6c7cd483600fa276308afc10438.

nuttycom (approved):
utACK

daira (approved):
utACK. This appears to fix all of the findings in the "Implementation of the Orchard network upgrade" section of the July 12, 2021 NCC draft report, apart from "MerklePath’s hash_layer() function could panic" which was fixed elsewhere; "Undefined padding rule" which is fixed in the spec, and "Non-genericity of Poseidon implementation" which will require communication with the Poseidon designers.

PR #145, opened 2021-07-10
Refactor `mul_fixed_short` API to copy in (`magnitude`, `sign`)

therealyingtong:
Documented in #146. 

The value commitment integrity check in the Action circuit involves a fixed-base scalar mul  `[v_old - v_new] ValueCommitV`, where `v_net = v_old - v_new` is a signed 64-bit value. `v_net` will be witnessed in the circuit as two cells: a `magnitude` and `sign`. These should be copied into the `mul_fixed_short` API. This means the `magnitude` can now be decomposed using a running sum, instead of being witnessed directly as bits.

This PR also inlines all `witness_scalar_*` APIs. A scalar is now directly witnessed in the region where it is used for multiplication, whereas previously it was witnessed separately and then copied in.


src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -196,122 +304,208 @@ pub mod tests {
...
+                -pallas::Base::one(),
+            ),
+            // There is a single canonical sequence of window values for which a doubling occurs on the last step:
+            // 1333333333333333333334 in octal.
+            // [0xB6DB_6DB6_DB6D_B6DC] B
+            (
+                "mul_with_double",
+                pallas::Base::from_u64(0xB6DB_6DB6_DB6D_B6DCu64),
+                pallas::Base::one(),
+            ),
```
therealyingtong, line 338: Test this value with both signs.

therealyingtong:
For the purposes of Orchard, `mul_fixed_base_field_elem()` should only ever be used with the `NullifierK` base. We should introduce another `FixedPointBaseField` enum to separate these two APIs.

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -173,6 +162,17 @@ impl Config {
                     &self.super_config.perm,
                 )?;
 
+                // Copy last window to `u` column
```
daira, line 166: Make a note that we're reusing the `u` column for another purpose.

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -357,4 +369,144 @@ pub mod tests {
...
+                short_config.assign(layouter, magnitude_sign, &ValueCommitV::get())?;
+
+                Ok(())
+            }
+        }
+
+        // Magnitude larger than 64 bits should fail
+        {
+            let circuit = MyCircuit {
+                magnitude: Some(pallas::Base::from_u128(1 << 64)),
```
daira, line 463: Test this for `1<<66` and `1<<254` as well.

src/circuit/gadget/ecc.rs
```diff
@@ -111,34 +88,34 @@ pub trait EccInstructions<C: CurveAffine>: Chip<C::Base> {
...
     fn mul_fixed(
         &self,
         layouter: &mut impl Layouter<C::Base>,
-        scalar: &Self::ScalarFixed,
+        scalar: Option<C::Scalar>,
         base: &Self::FixedPoints,
-    ) -> Result<Self::Point, Error>;
+    ) -> Result<(Self::Point, Self::ScalarFixed), Error>;
 
     /// Performs fixed-base scalar multiplication using a short signed scalar, returning `[scalar] base`.
```
daira, line 103:
```suggestion
    /// Performs fixed-base scalar multiplication using a short signed scalar, returning
    /// `[magnitude * sign] base`.
```

src/circuit/gadget/utilities/decompose_running_sum.rs
```diff
@@ -0,0 +1,369 @@
...
+                FIXED_BASE_WINDOW_SIZE,
+                NUM_WINDOWS,
+            > = MyCircuit {
+                alpha: Some(alpha),
+                strict: true,
+            };
+            let prover = MockProver::<pallas::Base>::run(8, &circuit, vec![]).unwrap();
+            assert_eq!(prover.verify(), Ok(()));
+
+            // Strict partial decomposition should fail.
```
daira, line 312: Is this true for all possible alpha, or only true except with negligible probability?

daira (approved):
utACK with comments.

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
therealyingtong: I realise that `1 << 66` would give the same failure as `1 << 64`, because the "last window check" is on the running sum, which will be nonzero (even if the three-bit window itself is zero).

src/circuit/gadget/utilities/decompose_running_sum.rs
therealyingtong: I am now only testing the partial decomposition failure cases using `alpha` strictly larger than 64 bits.

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -13,48 +16,59 @@ use pasta_curves::{arithmetic::FieldExt, pallas};
...
+            //    z_{i+1} = (z_i - a_i) / 2^3
+            // => a_i = z_i - z_{i+1} * 2^3
+            let word = z_cur - z_next * pallas::Base::from_u64(constants::H as u64);
 
-            vec![
-                q_scalar_fixed_short.clone() * last_window_check,
-                q_scalar_fixed_short * sign_check,
-            ]
+            self.super_config
+                .coords_check(meta, q_mul_fixed_running_sum, word)
```
str4d: This feels like it's probably a duplicated gate, given that it only queries columns directly from `super_config`.

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -204,35 +205,44 @@ impl Config {
...
-                        magnitude * sign
-                    });
-            let real_mul = scalar.map(|scalar| base.generator() * scalar);
-            let result = result.point();
-
-            if let (Some(real_mul), Some(result)) = (real_mul, result) {
-                assert_eq!(real_mul.to_affine(), result);
+            if let (Some(magnitude), Some(sign)) = (scalar.magnitude.value(), scalar.sign.value()) {
+                let magnitude_is_valid =
+                    magnitude < pallas::Base::from_u64(0xFFFF_FFFF_FFFF_FFFFu64);
```
str4d, line 219: This bound is slightly too small:
```suggestion
                    magnitude <= pallas::Base::from_u64(0xFFFF_FFFF_FFFF_FFFFu64);
```
(Non-critical because this is only test code, and if someone _did_ set a maximal magnitude this test would just be skipped.)

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -264,7 +264,7 @@ impl Config {
         self.overflow_config
             .overflow_check(layouter.namespace(|| "overflow check"), alpha, &zs)?;
 
-        Ok(result)
+        Ok((result, alpha))
```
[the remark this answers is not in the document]
therealyingtong: My motivation was to make the `mul` API resemble the `mul_fixed_*` APIs.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -165,6 +180,11 @@ impl EccChip {
             lookup_table,
             perm.clone(),
         );
+        let q_mul_fixed_running_sum = meta.selector();
+        let running_sum_full_config =
+            RunningSumConfig::configure(meta, q_mul_fixed_running_sum, advices[4], perm.clone());
+        let running_sum_short_config =
+            RunningSumConfig::configure(meta, q_mul_fixed_running_sum, advices[4], perm.clone());
```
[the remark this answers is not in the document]
therealyingtong: We will also have to remove the `WORD_NUM_BITS` const generic.

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
therealyingtong: This gate was indeed duplicated across `mul_fixed::short` and `mul_fixed::base_field_elem`. It has now been moved into `mul_fixed.rs`.

str4d (approved):
utACK 91b8ea20e4a2bbd8eba4144bf3f8468b8d41ad34 after finishing the removal of duplicate coords-check gates.

src/circuit/gadget/utilities/decompose_running_sum.rs
```diff
@@ -0,0 +1,391 @@
...
+//!     $$\alpha = k_0 + (2^K) k_1 + (2^{2K}) k_2 + ... + (2^{(W-1)K}) k_{W-1}$$
+//!
+//! $z_0$ is initialized as $\alpha$. Each successive $z_{i+1}$ is computed as
+//!                $$z_{i+1} = (z_{i} - k_i) / (2^K).$$
+//! $z_W$ is constrained to be zero.
+//! The difference between each interstitial running sum output is constrained
+//! to be $K$ bits, i.e.
+//!                      `range_check`($k_i$, $2^K$),
+//! where `range_check(word, range)
+//!     = word * (1 - word) * (2 - word) * ... * ((range - 1) - word)`
```
daira, line 13: https://spec.commonmark.org/0.30/#precedence

```suggestion
//! where
//! ```
//! range_check(word, range)
//!     = word * (1 - word) * (2 - word) * ... * ((range - 1) - word)
//! ```
```

src/circuit/gadget/utilities/decompose_running_sum.rs
```diff
@@ -0,0 +1,391 @@
...
+    pub fn copy_decompose(
+        &self,
+        region: &mut Region<'_, F>,
+        offset: usize,
+        alpha: CellValue<F>,
+        strict: bool,
+        word_num_bits: usize,
+        num_windows: usize,
+    ) -> Result<(CellValue<F>, RunningSum<F>), Error> {
+        let z_0 = copy(region, || "z_0 = alpha", self.z, offset, &alpha, &self.perm)?;
```
daira, line 136:
```suggestion
        let z_0 = copy(region, || "copy z_0 = alpha", self.z, offset, &alpha, &self.perm)?;
```

src/circuit/gadget/utilities/decompose_running_sum.rs
```diff
@@ -108,78 +98,101 @@ impl<
...
+        word_num_bits: usize,
+        num_windows: usize,
     ) -> Result<(CellValue<F>, RunningSum<F>), Error> {
-        let z_0 = copy(region, || "Copy alpha", self.z, offset, &alpha, &self.perm)?;
-        self.decompose(region, offset, z_0, strict)
+        let z_0 = copy(region, || "z_0 = alpha", self.z, offset, &alpha, &self.perm)?;
+        self.decompose(region, offset, z_0, strict, word_num_bits, num_windows)
     }
 
+    /// `z_0` must be the cell at `(self.z, offset)` in `region`.
```
daira, line 140: If `z_0` must be this, why not query it rather than passing a `CellValue`?

src/circuit/gadget/utilities/decompose_running_sum.rs
```diff
@@ -0,0 +1,391 @@
...
+            } else {
+                vec![None; num_windows]
+            }
+        };
+
+        // Initialize empty vector to store running sum values [z_1, ..., z_W].
+        let mut zs: Vec<CellValue<F>> = Vec::with_capacity(num_windows);
+        let mut z = z_0;
+
+        // Assign running sum `z_i`, i = 0..=n, where z_{i+1} = (z_i - k_i) / (2^K).
```
daira, line 195:
```suggestion
        // Assign running sum `z_{i+1}`, i = 0..n, where z_{i+1} = (z_i - k_i) / (2^K).
```

src/circuit/gadget/utilities/decompose_running_sum.rs
```diff
@@ -0,0 +1,391 @@
...
+        // Outside of this helper, z_0 = alpha must have already been loaded into the
+        // `z` column at `offset`.
+        let offset = offset + 1;
+
+        let two_pow_k_inv = F::from_u64(1 << WINDOW_NUM_BITS as u64).invert().unwrap();
+        for (idx, word) in words.iter().enumerate() {
```
daira, line 201:
```suggestion
        for (i, word) in words.iter().enumerate() {
```

src/circuit/gadget/utilities/decompose_running_sum.rs
```diff
@@ -0,0 +1,391 @@
...
+            let z_next = {
+                let word = word.map(|word| F::from_u64(word as u64));
+                let z_next_val = z
+                    .value()
+                    .zip(word)
+                    .map(|(z_cur_val, word)| (z_cur_val - word) * two_pow_k_inv);
+                let cell = region.assign_advice(
+                    || format!("z_{:?}", idx + 1),
+                    self.z,
+                    offset + idx,
```
daira, line 212:
```suggestion
                    || format!("z_{:?}", i + 1),
                    self.z,
                    offset + i,
```
Also consider using `offset + i + 1` and omitting the `let offset = offset + 1;` above.

src/circuit/gadget/utilities/decompose_running_sum.rs
str4d: This is during assignment, not configuration. Here we need to know:
- it's value (if during proving).
- that it is assigned at the correct relative offset in the correct column (the precondition stated in the comment).

This method states a precondition because there are two wrapping APIs: one that witnesses the value directly, and one that copies it in from somewhere else.

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -69,14 +69,11 @@ impl Config {
                 .coords_check(meta, q_mul_fixed_running_sum, word)
         });
 
-        // Check that we get z_85 = 0 as the final output of the three-bit decomposition running sum.
-        // Also check that the base field element is canonical.
+        // Check that the base field element is canonical.
```
daira, line 57: https://zcash.github.io/orchard/design/circuit/gadgets/ecc/fixed-base-scalar-mul.html#fixed-base-scalar-multiplication-using-base-field-element says that z_85 must be constrained to 0. Do we already constrain it elsewhere? If we do, then add a comment saying where.

src/circuit/gadget/utilities/decompose_running_sum.rs
therealyingtong: We could move the logic involving `z_0` out into the callers `witness_decompose()` / `copy_decompose()`, but I chose to include it in the internal function `decompose()` to reduce code duplication.

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
therealyingtong: `z_85 = 0` is now constrained by the `self.running_sum_config.copy_decompose()` call in `base_field_elem::Config_assign()`, with `strict = true`.

I'll add a comment to this effect + update the book.

str4d: The comment should go on `Config::create_gate`, which is where it is relevant.

@daira `z_85` is constrained because we now use strict checks during scalar decomposition.

src/circuit/gadget/utilities/decompose_running_sum.rs
str4d: I think it's fine as-is, now that the inner `decompose()` method is documenting the necessary precondition.

str4d: Also, we can't _avoid_ that precondition, as the running sum intentionally uses relative offsets; even if we moved the `z_0` logic outside `decompose()`, the `z_1` logic would still require this precondition even if all we passed in was the value of `z_0`.

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
daira: I don't see any uses of `strict = false` except in tests. Did I miss any? If not, why do we need the strict check to be optional?

therealyingtong: The updated book PR is at: https://github.com/zcash/orchard/pull/146.

str4d (approved):
re-utACK c444ddebf8c6037e0fc5bd183360c28f4693dd3f

src/circuit/gadget/utilities/decompose_running_sum.rs
```diff
@@ -9,9 +9,11 @@
 //! The difference between each interstitial running sum output is constrained
 //! to be $K$ bits, i.e.
 //!                      `range_check`($k_i$, $2^K$),
-//! where `range_check(word, range)
-//!     = word * (1 - word) * (2 - word) * ... * ((range - 1) - word)`
-//! is an expression of degree range.
+//! where
+//! ```
```
str4d, line 13: @daira Rust documentation is not plain Markdown; this gets interpreted as a Rust doctest. [We need an attribute here to mark it as not code](https://doc.rust-lang.org/rustdoc/documentation-tests.html#attributes):
```
```text
```

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
therealyingtong: I chose to include the `strict` option for a more general helper API.

(Sidenote: For the `lookup_range_check` helper, we only ever use `strict = false`, so we could choose to not provide a `strict = true` option there.)

str4d (changes requested):

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
daira: I find the code more confusing to think about with the two options, and more work to review. Let's remove the ones we don't use in Orchard.

str4d (approved):
re-utACK 1b615a40ee36b7e97518cb28e5ce8aa566b47118

daira (approved):
utACK with non-blocking comments.

PR #146, opened 2021-07-12
[book] Update constraints for short signed fixed-base mul.

therealyingtong:
Corresponds to #145. 

Previously, we witnessed the magnitude of a short signed scalar
directly as three-bit windows. Now, we decompose and range-constrain
it using a running sum.


str4d (commented):
Flushing comments.

daira (approved):
ACK

book/src/design/circuit/gadgets/decomposition.md
```diff
@@ -0,0 +1,43 @@
...
+
+$$
+\begin{aligned}
+z_0 &= \alpha \\
+    &= k_0 + 2^{K} \cdot k_1 + 2^{2K} \cdot k_2 +  2^{3K} \cdot k_3 + \cdots, \\
+z_1 &= (z_0 - k_0) / 2^K \\
+    &= k_1 + 2^{K} \cdot k_2 +  2^{2K} \cdot k_3 + \cdots, \\
+z_2 &= (z_1 - k_1) / 2^K \\
+    &= k_2 +  2^{K} \cdot k_3 + \cdots, \\
+    &\vdots
```
daira, line 14:
```suggestion
    &\vdots
z_W &= (z_{W-1} - k_{W-1}) / 2^K \\
    &= 0 \text{ (because } z_{W-1} = k_{W-1} \text{)}
```

daira: Technically this is only true in "Strict mode". Remind me again what the motivation was for having that as a mode, rather than always enforced?

book/src/design/circuit/gadgets/ecc/fixed-base-scalar-mul.md
```diff
@@ -25,23 +23,97 @@ $$
 \end{array}
```
daira, line 23: Just above here there is a `\cdot 1`; what is that for?

therealyingtong: I think that was when we had an extra factor of `1` in the impl. I'm going to remove this `\cdot 1` since we no longer have that.

PR #167, opened 2021-07-28
Add Action circuit benchmarks, fix Builder bug

str4d:
This benchmarks the prover and verifier for `Bundle`s produced by the `Builder` with between 1 and 3 outputs; in particular this means the `Bundle`s will follow any defaults set by the `Builder` (currently that it has a minimum of two `Action`s).

In order to enable benchmarking the prover with minimal changes to the public API, `Builder::build` has been altered to not create the proof. The caller can now choose when to perform this, at any point after then but before signature finalization.

The PR also fixes a bug in the `Builder`'s `Circuit` initialization (from #101).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,9 @@ criterion = "0.3"
 hex = "0.4"
 proptest = "1.0.0"
 
+[target.'cfg(unix)'.dev-dependencies]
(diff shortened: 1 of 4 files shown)


daira (approved):
(ignore approval; I was just looking at one commit)

daira (approved):
utACK

PR #171, opened 2021-08-04
Update Poseidon instantiation from 58 to 56 partial rounds. fixes #166

### LINKED ISSUE #166: Ensure our Poseidon parameters are consistent with the v1.1 reference implementation
We are currently using parameters that we pulled from the December 2020 Poseidon paper for Poseidon-128 (R<sub>F</sub> = 8, R<sub>P</sub> = 57), which we then modified to use R<sub>P</sub> = 58 because we required an even number of partial rounds. However, an auditor noticed that the v1.1 reference implementation's parameter generation scripts, when fed the Pallas base field, generated the parameters R<sub>F</sub> = 8, R<sub>P</sub> = 56. It turns out that Poseidon-128 in the paper uses a prime that is significantly over 2^254, whereas the Pallas modulus is only slightly over 2^254; this causes the scripts to recommend one fewer round.

It remains to be seen whether the Poseidon authors want to standardise on the parameters listed in the paper, or the parameters generated by their reference script. However, _we_ want to match their reference script, so we should switch to R<sub>F</sub> = 8, R<sub>P</sub> = 56. This has the nice side-effect of saving a row.

daira:
Test vectors are from https://github.com/zcash-hackworks/zcash-test-vectors/pull/45

Signed-off-by: Daira Hopwood <daira@jacaranda.org>


str4d (approved):
ACK. I have checked that the test vectors are copy-pasted in correctly.

PR #178, opened 2021-08-12
Speed up batched note decryption

str4d:


daira (approved):
utACK

src/keys.rs
```diff
@@ -567,15 +567,34 @@ impl SharedSecret {
         self.0.to_bytes()
     }
 
+    /// Only for use in batched note encryption.
+    pub(crate) fn batch_to_affine(shared_secrets: &[Option<Self>]) -> Vec<pallas::Affine> {
+        let secrets: Vec<_> = shared_secrets
+            .iter()
+            .filter_map(|s| s.as_ref().map(|s| *(s.0)))
+            .collect();
+        let mut secrets_affine = vec![pallas::Affine::identity(); shared_secrets.len()];
```
str4d, line 576: This should be `secrets.len()` (as the two slices passed into `batch_normalize` need to be the same length).

src/keys.rs
```diff
@@ -567,15 +567,34 @@ impl SharedSecret {
         self.0.to_bytes()
     }
 
+    /// Only for use in batched note encryption.
+    pub(crate) fn batch_to_affine(shared_secrets: &[Option<Self>]) -> Vec<pallas::Affine> {
```
str4d, line 571: I'll see if I can make this return an `impl Iterator`, so I can move the interpolation logic into here for clarity.

daira:
Typo in commit comment:
> This makes the method interface clearer, as the same patter of shared
> secrets is returned as was provided.

s/patter/pattern/

daira (approved):
utACK

PR #184, opened 2021-08-13
poseidon::Domain: Remove Spec trait bound.

therealyingtong:
The methods in the `Domain` trait are not generic over `Spec`.

### DIFF
--- a/src/primitives/poseidon.rs
+++ b/src/primitives/poseidon.rs
@@ -242,9 +242,7 @@ impl<F: FieldExt, S: Spec<F, T, RATE>, const T: usize, const RATE: usize> Duplex
 }
 
 /// A domain in which a Poseidon hash function is being used.
-pub trait Domain<F: FieldExt, S: Spec<F, T, RATE>, const T: usize, const RATE: usize>:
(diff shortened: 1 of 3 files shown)


src/primitives/poseidon.rs
```diff
@@ -300,7 +298,7 @@ impl<F: FieldExt, S: Spec<F, T, RATE>, const T: usize, const RATE: usize, const
 pub struct Hash<
     F: FieldExt,
     S: Spec<F, T, RATE>,
-    D: Domain<F, S, T, RATE>,
+    D: Domain<F, T, RATE>,
```
str4d, line 301: I looked back at the history to see why this relationship was present. The reason is that I originally implemented this primitive before we had const generics, so `Domain` was using the old associated types `S::State` and `S::Rate`. Now that we are using const generics, the consistent parameterization of `Spec` and `Domain` is enforced here, in the type signature of `Hash`.

str4d (approved):
utACK. I looked at the Poseidon domain separators again, and I cannot envisage any of them requiring details about the specific Poseidon specification other than `F, T, RATE`.

PR #186, opened 2021-08-18
Prepare to extract gadgets into crates

### LINKED ISSUE #183: Extract gadgets into their own crates
This will allow us to add a `poseidon` dependency in `halo2`, without introducing a circular dependency on the `orchard` crate.

Similarly, we can consider extracting `primitives::sinsemilla` into a crate.

Edit:

Let's extract all gadgets/chips/primitives into their own crates. These crates can live in a workspace repo (like `zcash/halo2-gadgets`).

therealyingtong:
Part of #183.

This PR decouples the gadgets from Orchard constants.


src/circuit.rs
```diff
@@ -564,8 +570,9 @@ impl plonk::Circuit<pallas::Base> for Circuit {
             // `product` = [poseidon_hash(nk, rho_old) + psi_old] NullifierK.
             //
             let product = {
-                let nullifier_k = FixedPointBaseField::from_inner(ecc_chip.clone(), NullifierK);
```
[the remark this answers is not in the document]
therealyingtong: Passing an incorrect `FixedPoint` (i.e. too few or too many windows) would result in a constraint failure. However, asserting the correct length at runtime will make it fail faster: https://github.com/zcash/orchard/pull/186/commits/428ff45e2732d1959b6a5bdac1877bb77be328b2

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -240,37 +197,11 @@ impl<const NUM_WINDOWS: usize> Config<NUM_WINDOWS> {
...
         coords_check_toggle: Selector,
     ) -> Result<(), Error> {
-        let mut constants = None;
-
-        let build_constants = || match base {
-            OrchardFixedBases::ValueCommitV => {
-                assert_eq!(NUM_WINDOWS, constants::NUM_WINDOWS_SHORT);
-                let base = ValueCommitV::get();
-                (
-                    base.lagrange_coeffs_short.0.as_ref().to_vec(),
```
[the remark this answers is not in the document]
therealyingtong: We still assert `assert_eq!(lagrange_coeffs.len(), NUM_WINDOWS)`, where `NUM_WINDOWS` is a const generic on the `&self: Config`.

src/circuit/gadget/ecc.rs
```diff
@@ -42,11 +45,7 @@ pub trait EccInstructions<C: CurveAffine>: Chip<C::Base> + UtilitiesInstructions
     /// elliptic curve point.
     type X: Clone + Debug;
     /// Enumeration of the set of fixed bases to be used in scalar mul with a full-width scalar.
-    type FixedPoints: Clone + Debug;
-    /// Enumeration of the set of fixed bases to be used in scalar mul with a base field element.
-    type FixedPointsBaseField: Clone + Debug;
-    /// Enumeration of the set of fixed bases to be used in short signed scalar mul.
-    type FixedPointsShort: Clone + Debug;
+    type FixedPoints: FixedPoints<C>;
```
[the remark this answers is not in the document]
therealyingtong: We still pass in this enum as `FixedPoints`, where the `ValueCommitV` variant has only 22 windows (while the rest have 85 windows).

```
pub enum OrchardFixedBases {
    CommitIvkR,
    NoteCommitR,
    ValueCommitR,
    SpendAuthG,
    NullifierK,
    ValueCommitV,
}
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -23,6 +22,34 @@ pub(super) mod mul;
 pub(super) mod mul_fixed;
 pub(super) mod witness_point;
 
+/// Number of windows for a full-width scalar
+pub const NUM_WINDOWS: usize =
+    (L_ORCHARD_SCALAR + FIXED_BASE_WINDOW_SIZE - 1) / FIXED_BASE_WINDOW_SIZE;
```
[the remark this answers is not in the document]
therealyingtong: This has been replaced with `pallas::Scalar::NUM_BITS`.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -23,6 +22,34 @@ pub(super) mod mul;
...
+
+/// Number of windows for a short signed scalar
+pub const NUM_WINDOWS_SHORT: usize =
+    (L_VALUE + FIXED_BASE_WINDOW_SIZE - 1) / FIXED_BASE_WINDOW_SIZE;
```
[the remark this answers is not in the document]
therealyingtong: `L_VALUE` has been renamed to `L_SCALAR_SHORT`.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -23,6 +22,34 @@ pub(super) mod mul;
...
+/// Number of bits in an unsigned short scalar.
+pub(crate) const L_VALUE: usize = 64;
+
+/// $\ell^\mathsf{Orchard}_\mathsf{base}$
+/// Number of bits in a Pallas base field element.
+pub(crate) const L_ORCHARD_BASE: usize = 255;
+
+/// $\ell^\mathsf{Orchard}_\mathsf{scalar}$
+/// Number of bits in a Pallas scalar field element.
+pub(crate) const L_ORCHARD_SCALAR: usize = 255;
```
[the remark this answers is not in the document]
therealyingtong: We have now replaced `L_ORCHARD_BASE` -> `pallas::Base::NUM_BITS` in the gadgets; and similarly `L_ORCHARD_SCALAR` -> `pallas::Scalar::NUM_BITS`.

src/circuit/gadget/sinsemilla/merkle/chip.rs
```diff
@@ -154,13 +175,18 @@ impl MerkleChip {
...
-    pub fn construct(config: MerkleConfig) -> Self {
+    pub fn construct(config: MerkleConfig<Hash, Commit, F>) -> Self {
         MerkleChip { config }
     }
 }
 
-impl MerkleInstructions<pallas::Affine, MERKLE_DEPTH_ORCHARD, { sinsemilla::K }, { sinsemilla::C }>
-    for MerkleChip
+impl<Hash, Commit, F>
+    MerkleInstructions<pallas::Affine, MERKLE_DEPTH_ORCHARD, { sinsemilla::K }, { sinsemilla::C }>
```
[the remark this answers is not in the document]
therealyingtong: We have now renamed `MERKLE_DEPTH_ORCHARD` -> `MERKLE_DEPTH` in the gadget.

str4d (changes requested):

src/circuit/gadget/ecc.rs
```diff
@@ -126,10 +128,22 @@ pub trait EccInstructions<C: CurveAffine>: Chip<C::Base> + UtilitiesInstructions
...
+///
+/// TODO: When associated consts can be used as const generics, introduce a
+/// `const NUM_WINDOWS: usize` parameter, and return `NUM_WINDOWS`-sized
+/// arrays instead of `Vec`s.
+pub trait FixedPoints<C: CurveAffine>: Debug + Eq + Clone {
+    fn generator(&self) -> C;
+    fn u(&self) -> Vec<[[u8; 32]; constants::H]>;
+    fn z(&self) -> Vec<u64>;
+    fn lagrange_coeffs(&self) -> Vec<[C::Base; constants::H]>;
+}
```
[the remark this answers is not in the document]
therealyingtong: I tried to re-enforce the `FixedPoint` bounds on `EccChip` with something like
```
<Fixed: super::FixedPoints<
    C,
    FullScalar: FixedPoint<C>,
    BaseField: FixedPoint<C>,
    ShortScalar: FixedPoint<C>
>>
```
and ran into the error that `associated type bounds are unstable.` https://rust-lang.github.io/rfcs/2289-associated-type-bounds.html

src/circuit/gadget/sinsemilla.rs
```diff
@@ -211,7 +210,8 @@ where
         Self::from_field_elem(chip, layouter, piece_value, num_words)
     }
 
-    fn from_field_elem(
+    /// Constructs a MessagePiece from a field element.
+    pub fn from_field_elem(
```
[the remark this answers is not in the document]
therealyingtong: I've made an issue for a bitstring gadget: https://github.com/zcash/halo2_gadgets/issues/12

ebfull (approved):

src/constants/sinsemilla.rs
```diff
@@ -0,0 +1,255 @@
...
+
+// Sinsemilla Q generators
+
+/// SWU hash-to-curve personalization for Sinsemilla $Q$ generators.
+pub const Q_PERSONALIZATION: &str = "z.cash:SinsemillaQ";
+
+// Sinsemilla S generators
+
+/// SWU hash-to-curve personalization for Sinsemilla $S$ generators.
+pub const S_PERSONALIZATION: &str = "z.cash:SinsemillaS";
```
str4d, line 38: These constants are part of the definition of Sinsemilla, so these constants should end up wherever the Sinsemilla chip does (i.e. in `halo2_gadgets`). Meanwhile `MERKLE_CRH_PERSONALIZATION`, and the hard-coded generators below, are Orchard-specific and will stick around here.

src/primitives/sinsemilla.rs
```diff
@@ -1,19 +1,37 @@
...
+
+// Sinsemilla Q generators
+
+/// SWU hash-to-curve personalization for Sinsemilla $Q$ generators.
+pub const Q_PERSONALIZATION: &str = "z.cash:SinsemillaQ";
+
+// Sinsemilla S generators
+
+/// SWU hash-to-curve personalization for Sinsemilla $S$ generators.
+pub const S_PERSONALIZATION: &str = "z.cash:SinsemillaS";
```
str4d, line 34: Ah, now this file will end up in `halo2_gadgets`, so I presume this means that `src/constants/sinsemilla.rs` is intended to stay in `orchard`?

src/circuit/gadget/ecc/chip.rs
```diff
@@ -31,23 +32,15 @@ pub const H: usize = 1 << FIXED_BASE_WINDOW_SIZE;
 
 /// Number of windows for a full-width scalar
 pub const NUM_WINDOWS: usize =
-    (L_ORCHARD_SCALAR + FIXED_BASE_WINDOW_SIZE - 1) / FIXED_BASE_WINDOW_SIZE;
+    (pallas::Scalar::NUM_BITS as usize + FIXED_BASE_WINDOW_SIZE - 1) / FIXED_BASE_WINDOW_SIZE;
```
str4d, line 35: In the original code this was `pallas::Base::NUM_BITS`. Was that a bug in the original impl?

src/circuit/gadget/sinsemilla/merkle.rs
```diff
@@ -16,12 +16,8 @@ pub(in crate::circuit) mod chip;
...
 pub const MERKLE_CRH_PERSONALIZATION: &str = "z.cash:Orchard-MerkleCRH";
 
-/// $\mathsf{MerkleDepth^{Orchard}}$
-pub(crate) const MERKLE_DEPTH_ORCHARD: usize = 32;
-
-/// $\ell^\mathsf{Orchard}_\mathsf{base}$
-/// Number of bits in a Pallas base field element.
-pub(crate) const L_ORCHARD_BASE: usize = 255;
+/// Depth of Merkle tree
+pub(crate) const MERKLE_DEPTH: usize = 32;
```
str4d, line 20: Is this hard-coded in the Merkle gadget? I don't see it used anywhere other than code comments (where yes it makes sense to rename it) and in the tests, which suggests this should be moved into the testing code and not be a constant here.

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -231,64 +187,41 @@ impl Config {
...
-                    base.lagrange_coeffs.0.as_ref().to_vec(),
-                    base.z.0.as_ref().to_vec(),
-                )
-            }
-        };
+        let lagrange_coeffs = base.lagrange_coeffs();
+        assert_eq!(lagrange_coeffs.len(), NUM_WINDOWS);
+
+        let z = base.z();
+        assert_eq!(z.len(), NUM_WINDOWS);
```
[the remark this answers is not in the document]
str4d: I checked locally, and indeed this would have caused a regression:
```
proving/bundle/1        time:   [3.8017 s 3.8160 s 3.8325 s]                              
                        change: [+12.469% +13.295% +14.165%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

src/circuit/gadget/sinsemilla/merkle.rs
```diff
@@ -13,6 +13,16 @@ use std::iter;
 
 pub(in crate::circuit) mod chip;
 
+/// SWU hash-to-curve personalization for the Merkle CRH generator
+pub const MERKLE_CRH_PERSONALIZATION: &str = "z.cash:Orchard-MerkleCRH";
```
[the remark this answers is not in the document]
therealyingtong: No; and it is deleted when in `halo2_gadgets` when tests are made non-specific to Orchard: https://github.com/zcash/halo2_gadgets/pull/2/files#diff-6d36c70bcb0dafa8815e83227c5cac1f0715b3a4c80c87d541765a339fe7183dL17

src/circuit/gadget/sinsemilla/merkle.rs
therealyingtong: This has been moved into the testing code.

src/circuit/gadget/ecc/chip.rs
therealyingtong: Yes, I believe that was a bug.

src/constants/sinsemilla.rs
therealyingtong: I'll import these from `src/primitives/sinsemilla`.

src/constants/fixed_bases.rs
```diff
@@ -0,0 +1,355 @@
...
+/// $2^{`FIXED_BASE_WINDOW_SIZE`}$
+pub const H: usize = 1 << FIXED_BASE_WINDOW_SIZE;
+
+/// Number of windows for a full-width scalar
+pub const NUM_WINDOWS: usize =
+    (L_ORCHARD_SCALAR + FIXED_BASE_WINDOW_SIZE - 1) / FIXED_BASE_WINDOW_SIZE;
+
+/// Number of windows for a short signed scalar
+pub const NUM_WINDOWS_SHORT: usize =
+    (L_VALUE + FIXED_BASE_WINDOW_SIZE - 1) / FIXED_BASE_WINDOW_SIZE;
```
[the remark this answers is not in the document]
therealyingtong: I think the chip could work over other constants as well. We currently assume these constants in the chip impl because it's too annoying to provide them all as const generics without `const_evaluatable_checked`.

src/constants/fixed_bases.rs
```diff
@@ -278,78 +158,4 @@ impl FixedPoint<pallas::Affine> for ValueCommitV {
...
-
-    for ((u, z), window_points) in u.iter().zip(z.iter()).zip(window_table) {
-        for (u, point) in u.iter().zip(window_points.iter()) {
-            let y = *point.coordinates().unwrap().y();
-            let u = C::Base::from_bytes(u).unwrap();
-            assert_eq!(C::Base::from_u64(*z) + y, u * u); // allow either square root
-            assert!(bool::from((C::Base::from_u64(*z) - y).sqrt().is_none()));
-        }
-    }
 }
```
str4d, line 161: Hmm, I'm -0 on removing this helper (which is ensuring that we have the correct hard-coded `Z` and `U` values). Instead we should expose `compute_window_tables` from `gadget::ecc::chip` and use it here. Once we make the `halo2_gadgets` split, it can be behind a `test-dependencies` feature flag.

src/constants/fixed_bases/commit_ivk_r.rs
```diff
@@ -2945,16 +2943,4 @@ mod tests {
...
-    fn lagrange_coeffs() {
-        let base = super::generator();
-        test_lagrange_coeffs(base, NUM_WINDOWS);
-    }
-
-    #[test]
-    fn z() {
-        let base = super::generator();
-        test_zs_and_us(base, &Z, &U, NUM_WINDOWS);
-    }
```
str4d, line 2959: If the helper is re-introduced, re-introduce this test (and likewise for every other constants module).

str4d (approved):
utACK 64baf8683b3c4fb46126e8cc5c9c9411ddbac693, with a -0 on the removal of a test (which isn't doing anything useful _currently_, but would defend against bugs in any future refactor that alters the way hard-coded constants are represented).

src/constants/fixed_bases.rs
daira: I agree with @str4d, except that I'm -100 (i.e. absolute veto) on removing these tests.

The problem is that without them, it's possible to backdoor the constants, as I pointed out in https://github.com/zcash/orchard/pull/43#discussion_r602737833 when they were originally added. (@ebfull convinced me that tests are sufficient rather than checking them on startup.)

therealyingtong: I've moved this to `ecc::chip::constants`, similarly to how `test_lagrange_coeffs`: https://github.com/zcash/orchard/pull/186/files#diff-08c8c700930e08d8a3478c6e1dbd727b5ab751a3514dbb8bf818b052ace8f8dbR180

therealyingtong: These tests have now been reintroduced. The `test_lagrange_coeffs` and `test_zs_and_us` functionality are exported from `ecc::chip::constants`. @daira said that it was unnecessary to add a `test-dependencies` feature flag since these are effectively part of the public API (we can't test Orchard without it).

src/constants/fixed_bases/commit_ivk_r.rs
therealyingtong: The helper and test has been re-introduced for this and every other constants module.

src/constants/fixed_bases.rs
str4d: I agree we don't need that flag in `orchard`. What I meant is that, once `halo2_gadgets` exists, we have that flag _there_, and then in `orchard` we enable the flag in `dev-dependencies`, since we _only_ need these to test Orchard, and not to use it.

src/constants/fixed_bases/value_commit_v.rs
```diff
@@ -803,13 +802,13 @@ mod tests {
     }
 
     #[test]
-    fn lagrange_coeffs_short() {
+    fn lagrange_coeffs() {
```
str4d, line 805: This did not need to be moved back into `orchard` from (what will go into) `halo2_gadgets`, since it is testing code that is provided by `halo2_gadgets`, not constants hard-coded in `orchard`. That being said, if we in future wanted to hard-code the Lagrange coefficients, we _would_ need these tests here, so I don't mind `test_lagrange_coeffs` being in the public testing API of `halo2_gadgets`.

src/circuit/gadget/ecc/chip/constants.rs
```diff
@@ -0,0 +1,273 @@
...
+/// <https://github.com/zcash/pasta>
+pub(crate) const T_Q: u128 = 45560315531506369815346746415080538113;
+
+/// The Pallas base field modulus is $p = 2^{254} + \mathsf{t_p}$.
+/// <https://github.com/zcash/pasta>
+pub(crate) const T_P: u128 = 45560315531419706090280762371685220353;
+
+/// For each fixed base, we calculate its scalar multiples in three-bit windows.
+/// Each window will have $2^3 = 8$ points.
+pub fn compute_window_table<C: CurveAffine>(base: C, num_windows: usize) -> Vec<[C; H]> {
```
str4d, line 40: This does not need to be public. But we can clean this up before publishing `halo2_gadgets`.

str4d (approved):
utACK 91bc1edf8f1848b93569cc12cd382eecb9c188d3

src/circuit/gadget/ecc/chip/constants.rs
```diff
@@ -0,0 +1,273 @@
...
+/// Number of bits in an unsigned short scalar.
+pub(crate) const L_SCALAR_SHORT: usize = 64;
+
+/// The Pallas scalar field modulus is $q = 2^{254} + \mathsf{t_q}$.
+/// <https://github.com/zcash/pasta>
+pub(crate) const T_Q: u128 = 45560315531506369815346746415080538113;
+
+/// The Pallas base field modulus is $p = 2^{254} + \mathsf{t_p}$.
+/// <https://github.com/zcash/pasta>
+pub(crate) const T_P: u128 = 45560315531419706090280762371685220353;
```
daira, line 36: I checked these constants (`T_Q` and `T_P`).

src/circuit/gadget/ecc/chip/constants.rs
```diff
@@ -0,0 +1,273 @@
...
+
+/// For each fixed base, we calculate its scalar multiples in three-bit windows.
+/// Each window will have $2^3 = 8$ points.
```
daira, line 39:
```suggestion
/// Each window will have $2^3 = 8$ points. The tables are computed as described in
/// <https://zcash.github.io/orchard/design/circuit/gadgets/ecc/fixed-base-scalar-mul.html#load-fixed-base>.
```

src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -231,71 +186,53 @@ impl Config {
...
+            (lagrange_coeffs, z)
         };
 
         // Assign fixed columns for given fixed base
         for window in 0..NUM_WINDOWS {
             coords_check_toggle.enable(region, window + offset)?;
 
             // Assign x-coordinate Lagrange interpolation coefficients
-            for k in 0..(constants::H) {
+            for k in 0..(H) {
```
daira, line 235:
```suggestion
            for k in 0..H {
```

src/circuit/gadget/sinsemilla.rs
```diff
@@ -357,6 +357,7 @@ where
     pub fn new(
         sinsemilla_chip: SinsemillaChip,
         ecc_chip: EccChip,
+        // Instead of using SinsemilllaChip::CommitDomains, just use something that implements a CommitDomains trait
```
daira, line 360:
```suggestion
        // TODO: Instead of using SinsemilllaChip::CommitDomains, just use something that implements a CommitDomains trait
```

src/circuit/gadget/sinsemilla/merkle/chip.rs
```diff
@@ -207,25 +235,23 @@ impl MerkleInstructions<pallas::Affine, MERKLE_DEPTH_ORCHARD, { sinsemilla::K },
             let b_1 = {
                 let b_1 = left
                     .value()
-                    .map(|value| bitrange_subset(value, 250..L_ORCHARD_BASE));
+                    .map(|value| bitrange_subset(value, 250..(pallas::Base::NUM_BITS as usize)));
```
daira, line 238: I checked that `L_ORCHARD_BASE` = `pallas::Base::NUM_BITS` = 255.

src/circuit/gadget/utilities.rs
```diff
@@ -115,6 +115,70 @@ pub fn range_check<F: FieldExt>(word: Expression<F>, range: usize) -> Expression
...
+///
+/// # Panics
+///
+/// We are returning a `Vec<u8>` which means the window size is limited to
+/// <= 8 bits.
+pub fn decompose_word<F: PrimeFieldBits>(
+    word: &F,
+    word_num_bits: usize,
+    window_num_bits: usize,
+) -> Vec<u8> {
```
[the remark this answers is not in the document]
daira: This doesn't seem essential to me.

src/circuit/gadget/ecc/chip/constants.rs
daira: However, they could be imported from `constants.rs`, I think.

src/constants.rs
```diff
@@ -43,235 +26,14 @@ pub(crate) const L_ORCHARD_SCALAR: usize = 255;
...
-/// is larger than t_P. t_P is defined in q_P = 2^254 + t_P for the
-/// Pallas base field.
-pub(crate) const PALLAS_T_PRIME_BITS: usize = 130;
-
-/// SWU hash-to-curve personalization for the spending key base point and
-/// the nullifier base point K^Orchard
-pub const ORCHARD_PERSONALIZATION: &str = "z.cash:Orchard";
-
 /// SWU hash-to-curve personalization for the group hash for key diversification
 pub const KEY_DIVERSIFICATION_PERSONALIZATION: &str = "z.cash:Orchard-gd";
```
daira, line 30: Why is this still needed here?

src/circuit/gadget/ecc/chip/constants.rs
therealyingtong: `src/constants.rs` will be in the `orchard` crate, but `ecc::chip::constants` will be in the `halo2_gadgets` crate.

src/constants.rs
str4d: This is Orchard-specific, so it's staying in this crate.

therealyingtong: `KEY_DIVERSIFICATION_PERSONALIZATION` is used in `src/spec.rs`.

src/circuit/gadget/ecc/chip/constants.rs
daira: OK, then the former should import these constants from the latter.

src/primitives/sinsemilla.rs
```diff
@@ -1,19 +1,37 @@
...
+
+/// $\frac{1}{2^K}$
+pub const INV_TWO_POW_K: [u8; 32] = [
+    1, 0, 192, 196, 160, 229, 70, 82, 221, 165, 74, 202, 85, 7, 62, 34, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 240, 63,
+];
+
+/// The largest integer such that $2^c \leq (r_P - 1) / 2$, where $r_P$ is the order
+/// of Pallas.
+pub const C: usize = 253;
```
daira, line 24: Why are these duplicated in `constants/sinsemilla.rs` and here?

Note that `K` will need to be generic for other uses of Sinsemilla than Orchard.

daira (approved):
utACK with minor comments.

src/primitives/sinsemilla.rs
str4d: Without `const_evaluatable_checked` it's hard for `K` to be generic. For now we'll leave it as a requirement of the chip implementation, and we can figure out how to refactor it later.

str4d: (as for the duplicated constants, we can clean those up later as well).

str4d (approved):
re-utACK a2868262b335a7e357c8c58ac8573163fdb7ecac

daira (commented):
Post-hoc ACK a2868262b335a7e357c8c58ac8573163fdb7ecac

PR #189, opened 2021-08-23
utilities: Implement function to load constant values

parazyd:
This function should be helpful if we want to use `UtilitiesInstructions` to load constants.


daira (approved):
utACK

PR #192, opened 2021-09-01
Tidy Merkle path logic.

### LINKED ISSUE #151: Tidy up Merkle path logic
- `orchard::tree::MerklePath::root` was not updated to be total in #136.
- The Merkle path chip tests manually derive the roots, instead of using `orchard::tree::MerklePath::root`.

therealyingtong:
Closes #151.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -664,7 +664,8 @@ pub mod testing {
             ),
             rng_seed in prop::array::uniform32(prop::num::u8::ANY)
         ) -> ArbitraryBundleInputs<StdRng> {
-            let mut tree = BridgeTree::<MerkleHashOrchard, 4>::new(100);
+            const MERKLE_DEPTH_ORCHARD: u8 = crate::constants::MERKLE_DEPTH_ORCHARD as u8;
+            let mut tree = BridgeTree::<MerkleHashOrchard, MERKLE_DEPTH_ORCHARD>::new(100);
             let mut notes_and_auth_paths: Vec<(Note, MerklePath)> = Vec::new();
 
             for note in notes.iter() {
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -569,6 +569,8 @@ impl<V> Bundle<InProgress<Proof, PartiallyAuthorized>, V> {
 /// Generators for property testing.
 #[cfg(any(test, feature = "test-dependencies"))]
 pub mod testing {
+    use incrementalmerkletree::{bridgetree::BridgeTree, Frontier, Tree};
+
     use rand::{rngs::StdRng, CryptoRng, SeedableRng};
     use std::convert::TryFrom;
     use std::fmt::Debug;
@@ -584,7 +586,8 @@ pub mod testing {
             testing::arb_spending_key, FullViewingKey, OutgoingViewingKey, SpendAuthorizingKey,
             SpendingKey,
         },
-        tree::{testing::arb_tree, Anchor, MerklePath},
+        note::testing::arb_note,
+        tree::{Anchor, MerkleHashOrchard, MerklePath},
         value::{testing::arb_positive_note_value, NoteValue, MAX_NOTE_VALUE},
         Address, Note,
     };
@@ -647,7 +650,11 @@ pub mod testing {
             n_recipients in 1..30,
         )
         (
-            (notes_and_auth_paths, anchor) in arb_tree(n_notes),
+            // generate note values that we're certain won't exceed MAX_NOTE_VALUE in total
+            notes in vec(
+                arb_positive_note_value(MAX_NOTE_VALUE / n_notes as u64).prop_flat_map(arb_note),
+                n_notes
+            ),
             recipient_amounts in vec(
                 arb_address().prop_flat_map(move |a| {
                     arb_positive_note_value(MAX_NOTE_VALUE / n_recipients as u64)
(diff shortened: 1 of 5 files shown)


str4d (changes requested):

src/tree.rs
```diff
@@ -88,6 +88,14 @@ impl MerklePath {
         }
     }
 
+    /// Instantiates a new Merkle path given a leaf position and authentication path.
+    pub(crate) fn new(position: u32, auth_path: [pallas::Base; MERKLE_DEPTH_ORCHARD]) -> Self {
```
nuttycom, line 92: It occurs to me that it'd probably be good to change https://github.com/zcash/incrementalmerkletree/blob/master/src/lib.rs#L220 to have a const generic parameter for the depth, and then this type could be moved there and used as the result of finding the authentication path.

nuttycom (approved):
utACK

therealyingtong:
[Force-pushed](https://github.com/zcash/orchard/compare/a55c1f25dc13073f8af3e61e8002a5f691f5f40c..291400ec336e592d74e36fab45214b81cd761f51) to rebase.

src/circuit.rs
```diff
@@ -397,12 +397,20 @@ impl plonk::Circuit<pallas::Base> for Circuit {
 
         // Merkle path validity check.
         let anchor = {
+            let path = self.path.map(|typed_path| {
+                // TODO: Replace with array::map once MSRV is 1.55.0.
+                let mut inner_path = [pallas::Base::zero(); MERKLE_DEPTH_ORCHARD];
+                for (a, b) in inner_path.iter_mut().zip(typed_path.iter()) {
+                    *a = b.inner();
+                }
+                inner_path
```
therealyingtong, line 406: Use `gen_const_array` here.

src/tree.rs
```diff
@@ -76,15 +78,30 @@ impl Anchor {
...
+                MerkleHashOrchard(pallas::Base::random(&mut rng))
+            }),
+        }
+    }
+
+    /// Instantiates a new Merkle path given a leaf position and authentication path.
+    pub(crate) fn new(position: u32, auth_path: [pallas::Base; MERKLE_DEPTH_ORCHARD]) -> Self {
+        Self {
+            position,
+            auth_path: auth_path
```
therealyingtong, line 99: Can use `array::map` in future.

src/tree.rs
```diff
@@ -410,7 +390,7 @@ pub mod testing {
                 for (pos, _) in commitments.iter().enumerate() {
 
                     // Initialize the authentication path to the path for an empty tree.
-                    let mut auth_path: [pallas::Base; MERKLE_DEPTH_ORCHARD] = (0..MERKLE_DEPTH_ORCHARD).map(|idx| EMPTY_ROOTS[idx]).collect::<Vec<_>>().try_into().unwrap();
+                    let mut auth_path = [MerkleHashOrchard::empty_leaf(); MERKLE_DEPTH_ORCHARD];
```
therealyingtong, line 393: This should be initialised to the auth path for the empty tree as before.

therealyingtong: Let's replace `arb_tree` with `incrementalmerkletree` implementation.

src/circuit.rs
nuttycom: I don't understand this comment; this isn't creating an array of a single constant value?

src/tree.rs
nuttycom: Same question here; the path has a different value at each index.

src/circuit.rs
nuttycom: I guess you'd use indexed access into typed_path to obtain the values? I cordially dislike indexed access compared to operating on iterators like this.

str4d: The thing with iterators like this is you need to be careful to ensure the lengths match, or you get short-cutting. We were careful in `gen_const_array` to use constants for lengths and indexing in the closure so that we'd get panics instead of silent truncation if the source array was shorter than the destination.

str4d: In any case, once we can use `array::map`, all this goes away.

nuttycom (approved):
re-utACK d47c157

Verified the last two commits.

src/builder.rs
```diff
@@ -657,10 +664,22 @@ pub mod testing {
             ),
             rng_seed in prop::array::uniform32(prop::num::u8::ANY)
         ) -> ArbitraryBundleInputs<StdRng> {
+            let mut tree = BridgeTree::<MerkleHashOrchard, 4>::new(100);
```
str4d, line 667: Copy-paste error; this will generate a small tree with paths that are the wrong length for the circuit.
```suggestion
            let mut tree = BridgeTree::<MerkleHashOrchard, MERKLE_DEPTH_ORCHARD>::new(100);
```

src/circuit.rs
```diff
@@ -397,12 +399,18 @@ impl plonk::Circuit<pallas::Base> for Circuit {
 
         // Merkle path validity check.
         let anchor = {
+            let path = self.path.map(|typed_path| {
+                // TODO: Replace with array::map once MSRV is 1.55.0.
+                gen_const_array_with_default(MerkleHashOrchard::empty_leaf().inner(), |i| {
+                    typed_path[i].inner()
+                })
```
str4d, line 406: Here we _could_ use the default version, as `pallas::Base` implements `Default`:
```suggestion
                gen_const_array(|i| typed_path[i].inner())
```

src/tree.rs
```diff
@@ -202,66 +181,78 @@ impl MerkleCrhOrchardOutput {
...
     }
 
     fn empty_root(altitude: Altitude) -> Self {
-        MerkleCrhOrchardOutput(EMPTY_ROOTS[<usize>::from(altitude)])
+        EMPTY_ROOTS[<usize>::from(altitude)]
     }
 }
 
-impl Serialize for MerkleCrhOrchardOutput {
+impl Serialize for MerkleHashOrchard {
```
str4d, line 249: Todo: remove the `Serialize` and `Deserialize` impls before we cut the first release. Non-blocking.

str4d (changes requested):

str4d (approved):

daira (approved):
utACK

PR #194, opened 2021-09-03
Switch to `memuse` crate for measuring heap allocations

str4d:


src/bundle.rs
```diff
@@ -463,6 +458,14 @@ impl<V> Bundle<Authorized, V> {
     }
 }
 
+impl<V: DynamicUsage> DynamicUsage for Bundle<Authorized, V> {
+    fn dynamic_usage(&self) -> usize {
+        self.actions.dynamic_usage()
+            + self.value_balance.dynamic_usage()
```
str4d, line 464: In putting this PR together, I realised there was an unavoidable mistake in the original implementation: the type `V` could have contained heap-allocated memory that went untracked. We could have placed a `V: Copy` bound to prevent this, but in any case, the `V: DynamicUsage` bound now correctly forces the caller to use a type that works.

daira (approved):
utACK

PR #197, opened 2021-09-13
`fpe 0.5`

str4d:


daira (approved):
utACK. I checked that this resolves the multiple versions of `fpe` and `aes` when used in zcashd.

PR #169, opened 2021-07-29
Circuit constraint refinements to reduce proof size

str4d:
Per https://github.com/zcash/orchard/pull/169#issuecomment-889423971:
- Base proof size increases by 64 bytes.
- Marginal (per-action) proof size decreases by 160 bytes.

Part of zcash/orchard#125.
Part of zcash/orchard#168.

PR #199, opened 2021-09-14
book: Update rotations for Commit^ivk region

str4d:
Matches the change made in zcash/orchard#169.


daira (approved):
ACK. Checked for consistency with #169.

PR #204, opened 2021-09-22
Add license exceptions for ECC and ZF projects

str4d:
Also bumps the `pasta_curves` and `halo2` dependencies for compatibility.


steven-ecc (approved):
utACK

PR #207, opened 2021-09-28
Add test against hardcoded pinned verification key

ebfull:


str4d (approved):
ACK. I confirmed that the test passes for me locally.

src/circuit.rs
```diff
@@ -987,6 +987,17 @@ mod tests {
...
 
         let vk = VerifyingKey::build();
+
+        // Test that the pinned verification key (representing the circuit)
+        // is as expected.
+        {
+            // panic!("{:#?}", vk.vk.pinned());
+            assert_eq!(
+                format!("{:#?}\n", vk.vk.pinned()).as_bytes(),
+                include_bytes!("circuit_description")
```
str4d, line 997: Hmm, could we get this to compare them as strings? Then the output if there is a difference would be more readable. It would be nice if it were also an intelligent diff.

src/circuit.rs
```diff
@@ -987,6 +987,17 @@ mod tests {
...
 
         let vk = VerifyingKey::build();
+
+        // Test that the pinned verification key (representing the circuit)
+        // is as expected.
+        {
+            // panic!("{:#?}", vk.vk.pinned());
+            assert_eq!(
+                format!("{:#?}", vk.vk.pinned()).as_bytes(),
+                include_bytes!("circuit_description")
```
str4d, line 997:
```suggestion
                format!("{:#?}", vk.vk.pinned()),
                include_str!("circuit_description")
```

ebfull: Good idea, though there will still be useful diffs in PRs for when `circuit_description` itself is updated, and assert_eq!() will print out the wrong string if the test fails, hence the use of the panic just before this.

I say we don't block on this.

str4d: Yeah, I had a look around, and I couldn't find any nice crates for pretty-printing the differences between multi-line strings (`pretty_assertions` is a good starting point, but they haven't merged any of the proposed approaches for this).

I agree that simply resolving the failing test will generate git commits that provide the equivalent information.

PR #217, opened 2021-10-26
Delete gadgets and introduce halo2_gadgets dependency.

therealyingtong:
This PR cherry-picks the two commits from https://github.com/zcash/orchard/pull/273/.

It also introduces https://github.com/zcash/halo2/pull/473 as a dependency.


src/circuit.rs
```diff
@@ -24,21 +24,18 @@ use crate::{
         nullifier::Nullifier,
         ExtractedNoteCommitment,
     },
-    primitives::{
-        poseidon::{self, ConstantLength},
-        redpallas::{SpendAuth, VerificationKey},
-    },
-    spec::NonIdentityPallasPoint,
+    primitives::redpallas::{SpendAuth, VerificationKey},    spec::NonIdentityPallasPoint,
```
daira, line 27:
```suggestion
    primitives::redpallas::{SpendAuth, VerificationKey},
    spec::NonIdentityPallasPoint,
```

daira (commented):
Looks good. We need to verify that the gadgets in `halo_gadgets` are identical (or equivalent, but preferably identical) to the ones from `orchard` before merging this.

Cargo.toml
```diff
@@ -89,4 +89,5 @@ debug = true
 debug = true
 
 [patch.crates-io]
-halo2 = { git = "https://github.com/zcash/halo2.git", rev = "f9b3ff2aef09a5a3cb5489d0e7e747e9523d2e6e" }
+# halo2_proofs = { git = "https://github.com/zcash/halo2.git", branch = "halo2_gadgets-import" }
+# halo2_gadgets = { git = "https://github.com/zcash/halo2.git", branch = "halo2_gadgets-import" }
```
str4d: This cherry-picked commit is now technically broken, because it doesn't have any patches active. I don't mind _too_ much if it stays broken, but if you're going to rebase this PR, then please fix it to match what is in #273.

Cargo.toml
```diff
@@ -78,15 +75,12 @@ harness = false
...
 [profile.release]
 debug = true
 
 [profile.bench]
 debug = true
 
 [patch.crates-io]
-halo2 = { git = "https://github.com/zcash/halo2.git", rev = "f9b3ff2aef09a5a3cb5489d0e7e747e9523d2e6e" }
+halo2_proofs = { git = "https://github.com/zcash/halo2.git", branch = "patch-gadgets-import" }
+halo2_gadgets = { git = "https://github.com/zcash/halo2.git", branch = "patch-gadgets-import" }
```
str4d, line 86: We should never reference a branch, because that can move; instead we should reference a specific revision. Now that https://github.com/zcash/halo2/pull/475 has merged, this should be:
```suggestion
halo2_proofs = { git = "https://github.com/zcash/halo2.git", rev = "f3502421cd48767ff84e2ebd965385fc29c0940e" }
halo2_gadgets = { git = "https://github.com/zcash/halo2.git", rev = "f3502421cd48767ff84e2ebd965385fc29c0940e" }
```

src/primitives/poseidon.rs
```diff
@@ -6,7 +6,7 @@ use std::fmt;
 use std::iter;
 use std::marker::PhantomData;
 
-use halo2::arithmetic::FieldExt;
+use halo2_proofs::arithmetic::FieldExt;
```
str4d, line 9: `src/primitives/poseidon*` should be deleted; it is provided by `halo2_gadgets` now (since we _need_ the implementation over there).

src/primitives/sinsemilla.rs
```diff
@@ -1,7 +1,7 @@
 //! The Sinsemilla hash function and commitment scheme.
 
 use group::{Curve, Wnaf};
-use halo2::arithmetic::{CurveAffine, CurveExt};
+use halo2_proofs::arithmetic::{CurveAffine, CurveExt};
```
str4d, line 4: `src/primitives/sinsemilla*` should be deleted; it is provided by `halo2_gadgets` now (since we _need_ the implementation over there).

str4d (changes requested):
The branch dependency on `zcash/halo2` needs to be fixed, and the extra modules deleted.

Cargo.toml
```diff
@@ -30,13 +30,14 @@ blake2b_simd = "1"
 ff = "0.11"
 fpe = "0.5"
 group = "0.11"
-halo2 = "=0.1.0-beta.1"
+halo2_gadgets = { version = "0.0", features = ["test-dependencies"] }
```
str4d, line 33: Instead of always enabling this, do the same thing we do for `zcash_note_encryption`: have this dependency line in both `dependencies` and `dev-dependencies`, with the latter enabling this feature flag.

str4d (approved):
utACK

daira (approved):
utACK

ebfull (approved):

PR #223, opened 2021-11-04
Add `orchard::circuit::Instance::from_parts()`

dconnolly:
When using `orchard` to validate Halo2 proofs, the only public option right now is to use `orchard::Bundle::verify_proof()`, which requires us to construct a whole `Bundle` instance, including encrypted ciphertext, ephemeral keys, and other data we don't need to verify the proof. 

Adding a public constructor for `orchard::circuit::Instance` will allow us to call `orchard::circuit::Proof::verify()` directly with less unnecessary data shuffling.

### DIFF
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -817,6 +817,27 @@ pub struct Instance {
 }
 
 impl Instance {
+    /// Constructs an [`Instance`] from its constituent parts
+    pub fn from_parts(
+        anchor: Anchor,
+        cv_net: ValueCommitment,
+        nf_old: Nullifier,
+        rk: VerificationKey<SpendAuth>,
+        cmx: ExtractedNoteCommitment,
+        enable_spend: bool,
+        enable_output: bool,
+    ) -> Self {
+        Instance {
+            anchor,
+            cv_net,
+            nf_old,
+            rk,
+            cmx,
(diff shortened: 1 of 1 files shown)


src/circuit.rs
```diff
@@ -817,6 +817,27 @@ pub struct Instance {
 }
 
 impl Instance {
+    /// Constructs an [`Instance`] from its constituent parts
```
str4d, line 820:
```suggestion
    /// Constructs an [`Instance`] from its constituent parts.
    ///
    /// This API can be used in combination with [`Proof::verify`] to build verification
    /// pipelines for many proofs, where you don't want to pass around the full bundle.
    /// Use [`Bundle::verify_proof`] instead if you have the full bundle.
    ///
    /// [`Bundle::verify_proof`]: crate::Bundle::verify_proof
```

str4d (approved):
utACK after documentation change. I looked over the current public API, and agree that this makes sense to expose for lower-level usage.

dconnolly:
@str4d I think the `CRITERION_TOKEN` is not available to workflows on PRs from forks, so the Benchmark job is failing

str4d (approved):
re-ACK

daira (commented):
Post-hoc utACK

Issue #214, opened 2021-10-21
`circuit.rs`: Remove superfluous Poseidon assignment


therealyingtong:
We are currently assigning `nk`, `rho_old` manually just to cast them from `CellValue` into `StateWord`. There is no need to do this, since `StateWord`s are copied into the Poseidon gadget anyway.

https://github.com/zcash/orchard/blob/4f9c0be42eabcf3cb79916b1a4af61844c3a3d8b/src/circuit.rs#L488-L510

str4d:
Closed by #224, merged into #237.

Issue #222, opened 2021-11-03
Rewrite current Poseidon implementation to be built around a Sponge, not a Duplex


str4d:
When I originally wrote the Poseidon implementation, I was operating under the mistaken belief that a Sponge could be built out of a Duplex. This is not actually true, because a Sponge pads on the whole-message level and then chunks on `RATE`, while a Duplex pads on the permutation level and thus must chunk on at most `RATE  - 1`. (I had believed they were interoperable because the Duplex paper constructs a Sponge from a Duplex in its security argument by interpolating padding, but I had misread the definition of the padding function and thought it allowed for a Sponge construction that didn't interpolate padding.)

This mismatch went unnoticed during the implementation, which led to some awkwardness around how padding was handled, and I just found a bug in it (which does not affect the Orchard circuit, but could affect someone else trying to use Poseidon once we extract it into `halo2_gadgets`).

The `ConstantLength` domain defines a constant `L`, which is intended to be an arbitrary but fixed length. In particular, it is supposed to work if `L > RATE`:
https://github.com/zcash/orchard/blob/4f9c0be42eabcf3cb79916b1a4af61844c3a3d8b/src/primitives/poseidon.rs#L351-L361

However, the `Domain::padding` implementation for `ConstantLength` does not correctly pad in this case: because the function is called on a permutation level, it only returns padding of length `RATE`, but the position of the permutation within `L` is not provided to the function, so if `L > RATE` it will always provide no padding, even if `L` is not a multiple of `RATE`. This happens because we use [`Iterator::skip`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.skip), which returns an empty iterator if the skipped amount is longer than the original iterator:
https://github.com/zcash/orchard/blob/4f9c0be42eabcf3cb79916b1a4af61844c3a3d8b/src/primitives/poseidon.rs#L274-L282

This is not a bug in the Orchard circuit, because our Poseidon spec has `RATE = 2`, and our only use of it is in a constant-length hash with `L = 2`, so there is intentionally no padding.
https://github.com/zcash/orchard/blob/4f9c0be42eabcf3cb79916b1a4af61844c3a3d8b/src/circuit.rs#L244-L247
https://github.com/zcash/orchard/blob/4f9c0be42eabcf3cb79916b1a4af61844c3a3d8b/src/primitives/poseidon/p128pow5t3.rs#L6-L15
[...]

PR #224, opened 2021-11-04
Allow passing `CellValue` as input to Poseidon gadget.

therealyingtong:
This removes a superfluous assignment in the Orchard circuit. 

This PR changes the circuit layout.

Closes zcash/orchard#214.


src/circuit.rs
```diff
@@ -485,30 +482,8 @@ impl plonk::Circuit<pallas::Base> for Circuit {
...
-                            )?;
-                            region.constrain_equal(var, message[i].cell())?;
-                            Ok(Word::<_, _, poseidon::P128Pow5T3, 3, 2>::from_inner(
-                                StateWord::new(var, value),
-                            ))
-                        };
-
-                        Ok([message_word(0)?, message_word(1)?])
-                    },
-                )?;
```
str4d, line 509: SAFETY: Removal of this region is fine:

- This region was copying the message pieces into new word cells.
- `PoseidonHash::hash` then passes those word cells to `Duplex::absorb`:
  https://github.com/zcash/orchard/blob/dfcea20569541399a32196c3cbb05af26c53afb1/src/circuit/gadget/poseidon.rs#L253-L257
- `Duplex` stores the word cells in an inner array, which is then passed to `poseidon_duplex`, and through to the `pad_and_add` instruction:
  https://github.com/zcash/orchard/blob/dfcea20569541399a32196c3cbb05af26c53afb1/src/circuit/gadget/poseidon.rs#L102-L104
- `pad_and_add` copies the word cells into a new region:
  https://github.com/zcash/orchard/blob/dfcea20569541399a32196c3cbb05af26c53afb1/src/circuit/gadget/poseidon/pow5t3.rs#L356-L381

So:
- Prior to this change, we were copying the origin cells into word cells, and then into the pad-and-add region.
- After this change, we copy the origin cells into the pad-and-add region directly.

src/circuit/gadget/poseidon.rs
```diff
@@ -149,13 +151,13 @@ impl<
     pub fn absorb(
         &mut self,
         mut layouter: impl Layouter<F>,
-        value: Word<F, PoseidonChip, S, T, RATE>,
+        value: CellValue<F>,
```
str4d, line 154: For symmetry, we should squeeze out `CellValue`s as well, keeping `Word`s for state cells, and the only guarantee we provide for the squeezed cells would be that they are in equality-constrained columns.

Non-blocking, as this is just as easy for me to do when addressing #222.

src/circuit_description
```diff
@@ -28342,30 +28342,30 @@ PinnedVerificationKey {
...
+            (0x19eb8760e7d0e6ae6d28d65627d958661cdde4e58a0aeb55a6b7017bcf723002, 0x064575794bf9bfdbc50e94b8afbbd2851ae4e12ff2809777d28fe71c235727d9),
+            (0x0e5c408b5224841cb4f75aec5cdb7bc95c150bbe858dbde8dbac9f72e5392462, 0x01030c69ac5fc7dd63e4e0bb1718d26f51b79dccc81e0b172e98c26e59145414),
+            (0x12437cb05ecff24131b52b5a55f6f143d8437c28c9d7c31eb06cfa093972a64b, 0x06e1a5e39566b4ce097a6c7dace6dcb827e54dac7d64fa79d994fb1557717614),
+            (0x34636ff9de412da15f41a8a006abbe8f43a5cffc94e6c3deb02f98af4fb2b8c7, 0x0270f0c3fa8cc7338f20fbcd5ec4e62799e051e0c5938d9e8c41581de8da6165),
+            (0x218e047b1c0a3b92c59539b3f6d9c23d34ebeeb65ca0be98f5e0e9642bdf1085, 0x20c1117f40b375688a94ff5c5c0b70004e43c7c7cd492fe8055fea081ea5ca78),
+            (0x2478c8226d4ede1c203fa7455b5fe28f99d5a0cb8ccdb5be4b54d5edcce974c4, 0x1ce69b76f05daeae57cd3d452370439237da89f2ddc84f7b2e35703acbf99655),
+            (0x08383138ecc6f2fb5459043c7666ae3df7802f1f02392af44db6ba25cd7d2c56, 0x20957d7a3f00a8589f627c5f5e471f45a84fbcbdcde00dfc97b9a97f3f723202),
```
str4d, line 28368: These changes make sense because we removed some equality constraints (by removing a region that was only involved in copies within advice columns), and so we're altering the contents of equality-permuted columns.

src/circuit_description
```diff
@@ -28342,30 +28342,30 @@ PinnedVerificationKey {
...
-        (0x29085e497c6097147a72a256459a7237a9bc8cb8e66b56a60f4839468f50ab7b, 0x3a8f21bdd0be7b7d0f641745f5008eedf16f02f04df65691045ef7b21d91443c),
-        (0x1fcae59b79411349d85836f3dfcba5ab523c2aa1426f62ddf80b7f85e41dfce5, 0x3d0eeb91ef1e68f59a6aead5c4e536f7d6a78721666002d0a92b0b923def40c7),
-        (0x161c09589b8e0e1d1712b98a4b5cea260d48365ab3f2bba6faec85a2b1966fe6, 0x29e201d0935308e8b1b04ef13c0b9b6891948f69da260bc5c8b12b62aa21155d),
-        (0x0e6d6b798948be5e9577257fcac9226c659ba7d6040b3b5d0c2df7bf541f18b7, 0x169ba3030178091cce287c13af9b83b376988cc95bc0aa51ea387be0e74964dd),
+        (0x00d87a2c430f1db50a63f18f8cf8807f4f70d3acb940d4130ba6811f8ba2d479, 0x13d5742320e1b2cecda6073b7f2bf5816b9067453deeaa829f356a65ef5621b2),
+        (0x3118979ade023f3977d034f86eed6506d7e0586ead81f80bc5ca01a7660ee0c9, 0x30f6731193d5c786cf61b05523c05e2664a066c2d39a685588f02883057320ad),
+        (0x0df51bd411d5f95da66fcc57f5e4d8dbcca3d82b08ceb61e9ff1befa074e08d3, 0x11c9092b6d02c46f173b0108854499ca4922afaf58e0734e77a6088715e84b64),
+        (0x24289014ede2672df8e8e32eb4e0d71709846041319fb85b1328cdb3b8764565, 0x0833de9c0b76ae816df0e41ae33daece27c63a41f2ba9abbbc7c08724211b50c),
```
str4d, line 28353: These changes make sense because they correspond to fixed columns containing selectors. Removing a region will cause other regions to be layed out slightly differently (at a minimum on one row further up), which could affect the order in which those regions' selectors are combined.

src/circuit_description
```diff
@@ -28325,15 +28325,15 @@ PinnedVerificationKey {
...
+        (0x37e70146801841d95832259e4d8f39aeee8a364ea5007f82aa84a0e387231315, 0x20ef65833381c985722a95e65125a1123cab3357741168a5ed7e92e972dbd30d),
+        (0x12e3af8e507a2ca30e544568cf319701ecbc29dc5919f0198d541938edecc8f3, 0x3fb1bb2804faaa4c215e93694d5d1e6f87874cb8c34cb9206ff958de14e44275),
+        (0x334d99f316343f01c8eb67e327c76f743f8de7f957c893c6f30ccd87e5d1af3a, 0x1da64caf127c8eb513653791147a85ed1edcca12935c95b7d615a9377c9406d8),
+        (0x1430dca15080286939046d3cb9fb7ace4ed1da2fbe7a362e8f3575489bc6e3e1, 0x1ef5f4aff2902f3bc8acb2cc884a2d44334f910a68b2701f1c37943652b46d8f),
+        (0x02ebdeac7e91b3d15e4b7c0533e42678672ec384d51e51e22342be7edeeb8074, 0x03638952e0489de03ff03236433f5d2617cb4ddd0a72637ed31095deca85a2a6),
+        (0x31179b7f5b01ad2a55cf9a66f57b696e9b9cb4919cca426d469b270827f3018b, 0x2b231a9a71b674cc546ba9b36916a13411648903cfdcb808926fc47ee745586c),
+        (0x0673497660cec8a8c391cfb16f5ffdcb710f9e9e194a1a85935cf4bc45b01359, 0x15ce8801c51811f0256a43f762e497fe38c88866c09bb256289d93e0393dc225),
+        (0x098d2a0cfedae91fe29e21a01b5e43045b9cc8d9a334f1aee3e075f36a16e323, 0x0be47cc41643c9e6d95012428d5b46ae5a44e765392e05152a0181e30689d8b6),
```
str4d, line 28336: These changes correspond to the nine fixed columns containing the incomplete addition portions of the fixed-base muls. It's not clear to me why their commitments change.

str4d (changes requested):
utACK. The second commit (that updates the circuit's pinned verification key) needs to be squashed into the first commit.

src/circuit_description
therealyingtong: Some of these columns are re-used as the Poseidon config's `rc_a` and `rc_b`, which are now being assigned in a different place?

str4d (approved):
utACK. Commits still need to be squashed together, but given that there's only one circuit change in this PR, we can do that with a squash-merge (**as long as the person merging the PR remembers to do this**).

therealyingtong:
> utACK. Commits still need to be squashed together, but given that there's only one circuit change in this PR, we can do that with a squash-merge (**as long as the person merging the PR remembers to do this**).

Do we need to squash the `squeeze()` commit, since it does not result in circuit changes?

daira (approved):
utACK

str4d:
> > utACK. Commits still need to be squashed together, but given that there's only one circuit change in this PR, we can do that with a squash-merge (**as long as the person merging the PR remembers to do this**).
> 
> Do we need to squash the `squeeze()` commit, since it does not result in circuit changes?

Ah no, sorry, I must have had an outdated commit view when I wrote that. Once looked again and it's fine now.

ebfull (approved):

daira (commented):
[deleted, apparently I did already ACK this]

PR #225, opened 2021-11-04
Switch from crates.io patches to git revision dependencies

str4d:
This is going to cause `zcash_note_encryption` to be present twice in the dependency tree of downstream users, until we cut a versioned release of it.


str4d:
Closing as obsolete: we will shortly be depending on versioned crates.

Issue #143, opened 2021-07-07
Ensure in the circuit we consistently use `N - x`, not `x - N` (when they are equivalent)


str4d:
We use a mix of these in various equivalent places throughout the circuit, and we also don't always match what is specified in the Orchard book.

daira:
We changed our mind, partly because there's a `Polynomial::one_minus` function and we'd have to change uses of that. Also 1 - x seems to be more common for booleans (possibly because of legacy habits from libsnark).

str4d:
This does make the range-constraint case `x(1 - x)(2 - x)...` a bit weird though, since it's semantically `(0 - x)(1 - x)(2 - x)...` (we'd need to make sure the `0` case is handled specially).

str4d:
We should go through the circuit and make a list of all the places we use this syntax, so we can decide which way to go (@ebfull has opinions I'm hoping he will share here).

therealyingtong:
- `circuit::gadget::utilities`
    - `bool_check()` uses `1 - x`. (Note: `bool_check()` should be re-written as a special case of `range_check()`.)
    - `range_check()` uses `x - N`
    - `utilities::cond_swap`: `configure()`'s `bool_check` should be replaced with `utilities::bool_check()`.
- `ecc::chip`
    - `add.rs`: `create_gate()`'s `poly2, poly{7..=12}` expressions use `(1 - x)`.
    - `mul_fixed.rs`:
        - `coords_check()` `on_curve` check uses `y^2 - x^3 - b`.
        - `mul_fixed::short`: `create_gate()`'s `sign_check` uses `x^2 - 1`; its `last_window_check` should be replaced with `utilities::bool_check()`.
    - `mul.rs`:
        - `create_gate()`'s `bool_check`  should be replaced with `utilities::bool_check()`.
        - `mul::complete`: `create_gate()'s` `bool_check`  should be replaced with `utilities::bool_check()`.
        - `mul::incomplete`: `create_gate()'s` `bool_check`  should be replaced with `utilities::bool_check()`.
        - `mul::overflow`: `create_gate()` uses `x - N` for its `recovery` and `lo_zero` checks; and `1 - x` for its `canonicity` check.
    - `witness_point.rs`: `create_gate()`'s `curve_eqn` closure uses `y^2 - x^3 - b`.

therealyingtong:
I prefer `N - x` because it doesn't result in a negative value when `x < N`.

str4d:
Closed by #235, merged into #237.

PR #235, opened 2021-11-18
circuit: Use `N - x` instead of `x - N`.

therealyingtong:
Closes https://github.com/zcash/orchard/issues/143.

### DIFF
--- a/src/circuit/gadget/ecc/chip/mul/complete.rs
+++ b/src/circuit/gadget/ecc/chip/mul/complete.rs
@@ -70,8 +70,8 @@ impl Config {
                 // k_i = 0 => y_p = -base_y
                 // k_i = 1 => y_p = base_y
                 let y_switch = {
-                    let k_minus_one = k.clone() - Expression::Constant(pallas::Base::one());
-                    k_minus_one * (base_y.clone() + y_p.clone()) + k * (base_y - y_p)
+                    let one_minus_k = Expression::Constant(pallas::Base::one()) - k.clone();
+                    one_minus_k * (base_y.clone() + y_p.clone()) + k * (base_y - y_p)
                 };
 
                 std::array::IntoIter::new([("bool_check", bool_check), ("y_switch", y_switch)])
--- a/src/circuit/gadget/ecc/chip/mul/complete.rs
+++ b/src/circuit/gadget/ecc/chip/mul/complete.rs
@@ -1,5 +1,6 @@
 use super::super::{add, copy, CellValue, EccConfig, EccPoint, Var};
 use super::{COMPLETE_RANGE, X, Y, Z};
+use crate::circuit::gadget::utilities::bool_check;
 
 use halo2::{
     circuit::Region,
@@ -58,9 +59,8 @@ impl Config {
 
                 // k_{i} = z_{i} - 2⋅z_{i+1}
                 let k = z_next - Expression::Constant(pallas::Base::from_u64(2)) * z_prev;
-                let k_minus_one = k.clone() - Expression::Constant(pallas::Base::one());
-                // (k_i) ⋅ (k_i - 1) = 0
-                let bool_check = k.clone() * k_minus_one.clone();
+                // (k_i) ⋅ (1 - k_i) = 0
+                let bool_check = bool_check(k.clone());
 
                 // base_y
                 let base_y = meta.query_advice(self.z_complete, Rotation::cur());
@@ -69,7 +69,10 @@ impl Config {
 
                 // k_i = 0 => y_p = -base_y
                 // k_i = 1 => y_p = base_y
-                let y_switch = k_minus_one * (base_y.clone() + y_p.clone()) + k * (base_y - y_p);
+                let y_switch = {
+                    let k_minus_one = k.clone() - Expression::Constant(pallas::Base::one());
+                    k_minus_one * (base_y.clone() + y_p.clone()) + k * (base_y - y_p)
+                };
 
                 std::array::IntoIter::new([("bool_check", bool_check), ("y_switch", y_switch)])
                     .map(move |(name, poly)| (name, q_mul_decompose_var.clone() * poly))
--- a/src/circuit/gadget/utilities.rs
+++ b/src/circuit/gadget/utilities.rs
@@ -102,7 +102,7 @@ pub fn transpose_option_array<T: Copy + std::fmt::Debug, const LEN: usize>(
 
 /// Checks that an expresssion is either 1 or 0.
 pub fn bool_check<F: FieldExt>(value: Expression<F>) -> Expression<F> {
-    value.clone() * (Expression::Constant(F::one()) - value)
+    range_check(value, 2)
 }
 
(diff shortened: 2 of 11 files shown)


str4d (changes requested):
Circuit description needs to be updated in each commit individually, so we can identify the effect of each individual change.

src/circuit/gadget/utilities/cond_swap.rs
```diff
@@ -179,7 +179,7 @@ impl<F: FieldExt> CondSwapChip<F> {
             let one = Expression::Constant(F::one());
 
             // a_swapped - b ⋅ swap - a ⋅ (1-swap) = 0
-            // This checks that `a_swapped` is equal to `y` when `swap` is set,
+            // This checks that `a_swapped` is equal to `b` when `swap` is set,
```
str4d, line 182: Move this comment fix into the commit that alters this file.

src/circuit/gadget/ecc/chip/mul/complete.rs
```diff
@@ -70,8 +70,8 @@ impl Config {
                 // k_i = 0 => y_p = -base_y
                 // k_i = 1 => y_p = base_y
                 let y_switch = {
-                    let k_minus_one = k.clone() - Expression::Constant(pallas::Base::one());
-                    k_minus_one * (base_y.clone() + y_p.clone()) + k * (base_y - y_p)
+                    let one_minus_k = Expression::Constant(pallas::Base::one()) - k.clone();
+                    one_minus_k * (base_y.clone() + y_p.clone()) + k * (base_y - y_p)
```
str4d, line 74: It occurs to me that we should probably also be consistent about expression ordering for these "either-or" constraints: either `(1 - b) * false_case + b * true_case` or `b * expr_if_true + (1 - b) * expr_if_false` (depending on whether we want it to read like a case expression, or a ternary expression).

Non-blocking; this cleanup should go into a separate PR (as it would also be nice to consider how we could ensure that `b` is a boolean-constrained expression).

str4d (changes requested):

src/circuit/gadget/ecc/chip/mul/complete.rs
```diff
@@ -58,9 +59,8 @@ impl Config {
 
                 // k_{i} = z_{i} - 2⋅z_{i+1}
                 let k = z_next - Expression::Constant(pallas::Base::from_u64(2)) * z_prev;
-                let k_minus_one = k.clone() - Expression::Constant(pallas::Base::one());
                 // (k_i) ⋅ (k_i - 1) = 0
```
str4d, line 62:
```suggestion
                // (k_i) ⋅ (1 - k_i) = 0
```
Or replace this comment entirely with something about it being a boolean constraint.

src/circuit/gadget/ecc/chip/mul/complete.rs
```diff
@@ -58,9 +59,8 @@ impl Config {
...
-                let bool_check = k.clone() * k_minus_one.clone();
+                let bool_check = bool_check(k.clone());
```
str4d, line 63: I believe this is the source of the circuit change in this particular commit.

src/circuit/gadget/utilities/cond_swap.rs
```diff
@@ -179,18 +179,18 @@ impl<F: FieldExt> CondSwapChip<F> {
...
             // but remains as `a` when `swap` is not set.
             let a_check =
                 a_swapped - b.clone() * swap.clone() - a.clone() * (one.clone() - swap.clone());
 
             // b_swapped - a ⋅ swap - b ⋅ (1-swap) = 0
             // This checks that `b_swapped` is equal to `a` when `swap` is set,
             // but remains as `b` when `swap` is not set.
-            let b_check = b_swapped - a * swap.clone() - b * (one.clone() - swap.clone());
+            let b_check = b_swapped - a * swap.clone() - b * (one - swap.clone());
```
str4d, line 190: Aside: Here's an interesting constraint where a ternary helper that ensures its halves are either positive or zero could be used (strengthening the argument for the ternary helper being built around `1 - x` instead of `x - 1`).

str4d (approved):
utACK after comment fix. I have not checked that the individual commits pass tests, but the changes to the verification key in each commit match what I expect.

ebfull (approved):
ACK modulo @str4d's suggestions.

src/circuit/gadget/ecc/chip/mul/complete.rs
daira: This is not a correct change.

daira (approved):
utACK for code, but please check against the Orchard and halo2 books and make any necessary changes there too.

src/circuit/gadget/ecc/chip/mul/complete.rs
therealyingtong: Sidenote: I don't think the complete addition constraints for variable-base scalar mul are currently documented in our book. https://zcash.github.io/orchard/design/circuit/gadgets/ecc/var-base-scalar-mul.html#circuit-design

therealyingtong:
[Force-pushed](https://github.com/zcash/orchard/compare/7c288b63ceae202a483cc2109d6efc4993d9b30c..409bbf36a0a1a538883cfc7cd2cf2790b1278ed3) to include book changes.

src/circuit/gadget/ecc/chip/mul/complete.rs
daira: Note that this change is correct because `y_switch` is being constrained to 0 (when `q_mul_decompose_var` is selected), and so sign does not matter.

daira (approved):
utACK 409bbf36a0a1a538883cfc7cd2cf2790b1278ed3

src/circuit/gadget/utilities.rs
```diff
@@ -131,7 +131,7 @@ pub fn bitrange_subset<F: FieldExt + PrimeFieldBits>(field_elem: F, bitrange: Ra
 /// i.e. 0 ≤ word < range.
 pub fn range_check<F: FieldExt>(word: Expression<F>, range: usize) -> Expression<F> {
     (1..range).fold(word.clone(), |acc, i| {
-        acc * (word.clone() - Expression::Constant(F::from_u64(i as u64)))
+        acc * (Expression::Constant(F::from_u64(i as u64)) - word.clone())
```
str4d, line 134: I noticed that this commit didn't include a book change, and that's because this implementation was previously conflicting with the [book](https://zcash.github.io/orchard/design/circuit/gadgets/decomposition.html#short-range-decomposition) (that definition was originally added in https://github.com/zcash/orchard/commit/2dd23f47b80da03c202f7cad2d04e7494ca7e433). It now matches the book, but we should verify that all uses of `range_check` (for which the circuit changes in this commit) match all uses of `range_check` in the book.

str4d (approved):
re-ACK e0a0a0d509ea47ba27fb70ae2f53646d24beb2a7

PR #237, opened 2021-11-19
Orchard proposed mainnet circuit

### LINKED ISSUE #231: Cut proposed mainnet version of the Orchard circuit
We are making some cleanup changes to the Action circuit (in response to comments from auditors, as well as making some of the Orchard gadgets reusable). We are making these changes in the `orchard-mainnet-circuit` branch.

Prior to defining the mainnet activation height, we need to commit to the exact circuit we will be using. This will be via a PR from the `orchard-mainnet-circuit` branch to `main`. At the point we make this PR, the branch should depend on `halo2 0.1.0` and `halo2_gadgets 0.1.0` crate releases, thus enabling us to make both backwards-compatible and backwards-incompatible changes to those crates.

str4d:
This is the feature branch for the proposed mainnet version of the Orchard Action circuit.

Closes #231.


str4d:
`orchard 0.1.0-beta.1` has been released, which encodes the first testnet version of the circuit. The next beta release of this crate will encode the proposed mainnet circuit, and we've made the changes we wanted to here. The only remaining change to the circuit code will be #183, but that will be easier to perform directly on `main`.

str4d:
I confirmed locally that this branch only contains merge commits. So there's new to review in this PR, beyond what was already reviewed in the PRs themselves.

PR #240, opened 2021-11-26
Migrate to latest halo2 revision

str4d:


therealyingtong (approved):
utACK. I like that we can now specify failing values in unit tests.

PR #247, opened 2021-11-30
circuit: Introduce ternary expression helper.

### LINKED ISSUE #242: circuit: Introduce helper for ternary expression
See https://github.com/zcash/orchard/pull/235#discussion_r753218334.

therealyingtong:
Closes #242.

### DIFF
--- a/src/circuit/gadget/utilities/cond_swap.rs
+++ b/src/circuit/gadget/utilities/cond_swap.rs
@@ -1,7 +1,7 @@
-use super::{bool_check, copy, CellValue, UtilitiesInstructions, Var};
+use super::{bool_check, copy, ternary, CellValue, UtilitiesInstructions, Var};
 use halo2::{
     circuit::{Chip, Layouter},
-    plonk::{Advice, Column, ConstraintSystem, Error, Expression, Selector},
+    plonk::{Advice, Column, ConstraintSystem, Error, Selector},
     poly::Rotation,
 };
 use pasta_curves::arithmetic::FieldExt;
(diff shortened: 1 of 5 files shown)


src/circuit/gadget/utilities/cond_swap.rs
```diff
@@ -176,18 +176,17 @@ impl<F: FieldExt> CondSwapChip<F> {
...
-            // a_swapped - b ⋅ swap - a ⋅ (1-swap) = 0
             // This checks that `a_swapped` is equal to `b` when `swap` is set,
             // but remains as `a` when `swap` is not set.
-            let a_check =
-                a_swapped - b.clone() * swap.clone() - a.clone() * (one.clone() - swap.clone());
+            let a_check = ternary(
+                swap.clone(),
+                a_swapped.clone() - b.clone(),
+                a_swapped - a.clone(),
+            );
```
str4d, line 185: Hmm, this is a larger change than I'd anticipated. I had expected this to look like:
```suggestion
            let a_check = a_swapped - ternary(swap.clone(), b.clone(), a.clone());
```
which would result in a circuit change that is just rearranging some negations. This approach works because `ternary` returns an expression where the multiplicative factor of each half is either `0` or `1`, not `-1`.

Weakly blocking; I'm also fine with this approach (it results in a slightly larger constraint but that is clearly correct regardless of how `ternary` works), but I'd prefer if it were implemented across two commits, to make review easier (first commit introducing `ternary` per my suggestion above, second commit switching to the encompassing usage).

src/circuit/gadget/utilities/cond_swap.rs
```diff
@@ -176,18 +176,17 @@ impl<F: FieldExt> CondSwapChip<F> {
...
 
-            // b_swapped - a ⋅ swap - b ⋅ (1-swap) = 0
             // This checks that `b_swapped` is equal to `a` when `swap` is set,
             // but remains as `b` when `swap` is not set.
-            let b_check = b_swapped - a * swap.clone() - b * (one - swap.clone());
+            let b_check = ternary(swap.clone(), b_swapped.clone() - a, b_swapped - b);
```
str4d, line 189: Similarly, this would have been simpler as:
```suggestion
            let b_check = b_swapped - ternary(swap.clone(), a, b);
```
Similarly weakly blocking: if we take the encompassing approach, it would be nice to make the change across two commits, so the diff on `circuit_description` is easier to review.

str4d (approved):
utACK 556229d8d4b4b419988cc491136b6d2eda43c7d4. I have not closely checked the `circuit_description` diff in the last commit because as-structured it's tricky to review.

str4d (approved):
utACK. I checked the constraint changes against the `circuit_description` changes.

daira (approved):
utACK. I only checked the source, not `circuit_description` (but I'm happy that @str4d has sufficiently checked the latter).

PR #248, opened 2021-11-30
circuit: Refactor `EccConfig` away from `impl From<EccConfig>`.

### LINKED ISSUE #246: circuit: Refactor away from `impl From<EccConfig>`.

therealyingtong:
Closes #246. Based on #247. 

Review recommendation: commit-by-commit, with whitespace changes hidden.

This PR introduces a `configure()` method for each sub-config, refactoring away from the `impl From<EccConfig>` pattern. This allows each sub-config to enable equalities for the columns it needs.

This PR changes the order in which columns and gates are created; however, it DOES NOT change the number of columns.

### DIFF
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -145,9 +145,9 @@ pub struct EccConfig {
     add: add::Config,
 
     /// Variable-base scalar multiplication (hi half)
-    pub q_mul_hi: (Selector, Selector, Selector),
+    mul_hi: mul::incomplete::Config<{ mul::INCOMPLETE_HI_LEN }>,
     /// Variable-base scalar multiplication (lo half)
-    pub q_mul_lo: (Selector, Selector, Selector),
+    mul_lo: mul::incomplete::Config<{ mul::INCOMPLETE_LO_LEN }>,
     /// Selector used to enforce boolean decomposition in variable-base scalar mul
     pub q_mul_decompose_var: Selector,
     /// Selector used to enforce switching logic on LSB in variable-base scalar mul
@@ -253,14 +253,23 @@ impl EccChip {
             advices[6], advices[7], advices[8],
         );
 
+        // Components of mul::Config
+        // TODO: Move this into mul::Config.
+        let mul_hi = mul::incomplete::Config::configure(
+            meta, advices[9], advices[3], advices[0], advices[1], advices[4], advices[5],
+        );
+        let mul_lo = mul::incomplete::Config::configure(
+            meta, advices[6], advices[7], advices[0], advices[1], advices[8], advices[2],
+        );
+
         let config = EccConfig {
             advices,
             lagrange_coeffs,
             fixed_z: meta.fixed_column(),
             add_incomplete,
             add,
-            q_mul_hi: (meta.selector(), meta.selector(), meta.selector()),
-            q_mul_lo: (meta.selector(), meta.selector(), meta.selector()),
+            mul_hi,
+            mul_lo,
             q_mul_decompose_var: meta.selector(),
             q_mul_overflow: meta.selector(),
             q_mul_lsb: meta.selector(),
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -139,7 +139,7 @@ pub struct EccConfig {
     pub fixed_z: Column<Fixed>,
 
     /// Incomplete addition
-    pub q_add_incomplete: Selector,
+    add_incomplete: add_incomplete::Config,
 
     /// Complete addition
     pub q_add: Selector,
@@ -249,12 +249,15 @@ impl EccChip {
 
         // Create witness point gate
         let witness_point = witness_point::Config::configure(meta, advices[0], advices[1]);
+        // Create incomplete point addition gate
+        let add_incomplete =
+            add_incomplete::Config::configure(meta, advices[0], advices[1], advices[2], advices[3]);
 
         let config = EccConfig {
             advices,
             lagrange_coeffs,
             fixed_z: meta.fixed_column(),
-            q_add_incomplete: meta.selector(),
+            add_incomplete,
             q_add: meta.selector(),
             q_mul_hi: (meta.selector(), meta.selector(), meta.selector()),
             q_mul_lo: (meta.selector(), meta.selector(), meta.selector()),
@@ -269,12 +272,6 @@ impl EccChip {
             lookup_config: range_check,
             running_sum_config,
         };
-
-        // Create incomplete point addition gate
-        {
-            let config: add_incomplete::Config = (&config).into();
-            config.create_gate(meta);
-        }
 
         // Create complete point addition gate
         {
@@ -432,7 +429,7 @@ impl EccInstructions<pallas::Affine> for EccChip {
         a: &Self::NonIdentityPoint,
         b: &Self::NonIdentityPoint,
     ) -> Result<Self::NonIdentityPoint, Error> {
-        let config: add_incomplete::Config = self.config().into();
+        let config = self.config().add_incomplete;
         layouter.assign_region(
             || "incomplete point addition",
             |mut region| config.assign_region(a, b, 0, &mut region),
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -133,11 +133,6 @@ pub struct EccConfig {
     /// Advice columns needed by instructions in the ECC chip.
     pub advices: [Column<Advice>; 10],
 
-    /// Coefficients of interpolation polynomials for x-coordinates (used in fixed-base scalar multiplication)
-    pub lagrange_coeffs: [Column<Fixed>; constants::H],
-    /// Fixed z such that y + z = u^2 some square, and -y + z is a non-square. (Used in fixed-base scalar multiplication)
-    pub fixed_z: Column<Fixed>,
-
     /// Incomplete addition
     add_incomplete: add_incomplete::Config,
 
@@ -146,6 +141,9 @@ pub struct EccConfig {
 
     /// Variable-base scalar multiplication
     mul: mul::Config,
+
+    /// TODO: Remove this.
+    pub mul_fixed: mul_fixed::Config,
 
     /// Fixed-base full-width scalar multiplication
     pub q_mul_fixed_full: Selector,
@@ -206,10 +204,6 @@ impl EccChip {
     ) -> <Self as Chip<pallas::Base>>::Config {
         // The following columns need to be equality-enabled for their use in sub-configs:
         //
-        // mul_fixed::Config:
-        // - advices[4]: window
-        // - advices[5]: u
-        //
         // mul_fixed::base_field_element::Config:
         // - [advices[6], advices[7], advices[8]]: canon_advices
         //
@@ -238,13 +232,26 @@ impl EccChip {
         // Create variable-base scalar mul gates
         let mul = mul::Config::configure(meta, add, range_check, advices);
 
+        // Create config that is shared across short, base-field, and full-width
+        // fixed-base scalar mul.
+        let mul_fixed = mul_fixed::Config::configure(
+            meta,
+            q_mul_fixed_running_sum,
+            lagrange_coeffs,
+            advices[4],
+            advices[0],
+            advices[1],
+            advices[5],
+            add,
+            add_incomplete,
+        );
+
         let config = EccConfig {
             advices,
-            lagrange_coeffs,
-            fixed_z: meta.fixed_column(),
             add_incomplete,
             add,
             mul,
+            mul_fixed,
             q_mul_fixed_full: meta.selector(),
             q_mul_fixed_short: meta.selector(),
             q_mul_fixed_base_field: meta.selector(),
@@ -253,14 +260,6 @@ impl EccChip {
             lookup_config: range_check,
             running_sum_config,
         };
-
-        // Create gate that is used both in fixed-base mul using a short signed exponent,
-        // and fixed-base mul using a base field element.
-        {
-            // The const generic does not matter when creating gates.
-            let mul_fixed_config: mul_fixed::Config<{ constants::NUM_WINDOWS }> = (&config).into();
-            mul_fixed_config.running_sum_coords_gate(meta);
-        }
 
         // Create gate that is only used in full-width fixed-base scalar mul.
         {
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -142,7 +142,7 @@ pub struct EccConfig {
     add_incomplete: add_incomplete::Config,
 
     /// Complete addition
-    pub q_add: Selector,
+    add: add::Config,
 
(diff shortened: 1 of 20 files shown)


src/circuit/gadget/ecc/chip.rs
```diff
@@ -221,9 +223,6 @@ impl EccChip {
         // mul_fixed::base_field_element::Config:
         // - [advices[6], advices[7], advices[8]]: canon_advices
         //
-        // mul::overflow::Config:
-        // - [advices[0], advices[1], advices[2]]: advices
-        //
```
therealyingtong, line 234: I believe this (now deleted) comment should have said `[advices[6], advices[7], advices[8]]`.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -276,12 +280,6 @@ impl EccChip {
             config.create_gate(meta);
         }
 
-        // Create incomplete point addition gate
-        {
-            let config: add_incomplete::Config = (&config).into();
-            config.create_gate(meta);
-        }
```
str4d: The change to `circuit_description` is more complex than it needs to be, because `add_incomplete::Config` is refactored before `witness_point::Config`. This causes constraints to be reordererd in the first commit, that are subsequently un-reordered in the third commit when `witness_point::Config` is refactored and one more configured before `add_incomplete::Config`.

If the commits were reordered so that they refactored these gates in the order they are configured in the circuit, then the only change to `circuit_description` should be renumbering fixed columns (or replacing standalone selector columns with compressed selector columns, or vice versa), due to moving the sub-configs above the main config's selector creation. I manually confirmed this to be the case for the first three commits (in fact, as a result the `add_incomplete::Config` and `add::Config` refactors cause no circuit change).

str4d (changes requested):

src/circuit/gadget/ecc/chip.rs
therealyingtong: [Force-pushed](https://github.com/zcash/orchard/compare/8d433ff62d2bba5ab8c4a51b2aa62c5737a7cb8a..5671355ade4008a052 750cad8247998743dad266) to reorder the commits as suggested.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -131,7 +132,6 @@ impl Config {
         self.hi_config.create_gate(meta);
         self.lo_config.create_gate(meta);
         self.complete_config.create_gate(meta);
-        self.overflow_config.create_gate(meta);
```
str4d, line 134: Refactoring `mul::overflow::Config` first from `mul::Config` causes gate reorderings. I'll see if there's a way to prevent this...

src/circuit/gadget/ecc/chip.rs
```diff
@@ -256,6 +253,7 @@ impl EccChip {
         // TODO: Move this into mul::Config.
         let mul_overflow =
             mul::overflow::Config::configure(meta, range_check, advices[6..9].try_into().unwrap());
+        let mul_complete = mul::complete::Config::configure(meta, advices[9], add);
```
str4d: This was placed after `mul_overflow`, whereas prior to this PR it was configured before. At a minimum, reorder these two commits so `mul_complete` is refactored before `mul_overflow`.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -254,15 +254,21 @@ impl EccChip {
         let mul_overflow =
             mul::overflow::Config::configure(meta, range_check, advices[6..9].try_into().unwrap());
         let mul_complete = mul::complete::Config::configure(meta, advices[9], add);
+        let mul_hi = mul::incomplete::Config::configure(
+            meta, advices[9], advices[3], advices[0], advices[1], advices[4], advices[5],
+        );
+        let mul_lo = mul::incomplete::Config::configure(
+            meta, advices[6], advices[7], advices[0], advices[1], advices[8], advices[2],
+        );
```
str4d: And then the third commit refactors `mul::incomplete::Config` last. I think we can probably eliminate the constraint reordering entirely if we:
- Reorder the commits to refactor `mul::incomplete`, `mul::complete`, `mul::overflow` in that order.
- Reorder the variables above to be `mul_hi, mul_lo, mul_complete, mul_overflow`.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -219,10 +207,6 @@ impl EccChip {
             meta.enable_equality((*column).into());
         }
 
-        let q_mul_fixed_running_sum = meta.selector();
-        let running_sum_config =
-            RunningSumConfig::configure(meta, q_mul_fixed_running_sum, advices[4]);
```
str4d, line 250: Moving these into `mul_fixed::Config::configure` causes constraint reordering. If we do want to do this, we should do it in a separate commit, so that the `mul_fixed::Config` refactor does not include constraint reordering.

str4d (changes requested):

therealyingtong:
[Force-pushed](https://github.com/zcash/orchard/compare/5671355ade4008a052750cad8247998743dad266..a08853d32bdb71ff118f8561a20b32fbcc9f208e) to reorder commits as suggested by @str4d.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -128,8 +141,6 @@ impl Config {
             .map(move |(name, poly)| (name, q_mul_lsb.clone() * poly))
         });
 
-        self.hi_config.create_gate(meta);
-        self.lo_config.create_gate(meta);
```
str4d, line 132: Hmm, I didn't spot that there was a gate defined just above these; all three of these sub-config refactors still move across the above gate, causing a constraint reorder in every commit.

I'm going to try inserting a commit that temporarily splits up `mul::Config` into two pieces, so that the first can be refactored before this commit, to prevent re-ordering. Then the existing "`mul::Config` refactor" commit can re-assemble the pieces.

str4d:
I [force-pushed](https://github.com/zcash/orchard/compare/a08853d32bdb71ff118f8561a20b32fbcc9f208e..a09173a331394a77c004e17faa8bc0adafca8a4c) to add an extra commit that cleans up the circuit description changes. Now every mul sub-config refactor causes no change to the circuit, while the overall PR diff is unaltered.

src/circuit/gadget/ecc/chip/mul.rs
str4d: Done in 0ede6b230144ec994a4fcbcc24ef5858f7a78084.

ebfull (approved):

src/circuit/gadget/ecc/chip.rs
```diff
@@ -165,10 +165,8 @@ pub struct EccConfig {
     /// when the scalar is a signed short exponent or a base-field element.
     pub q_mul_fixed_running_sum: Selector,
 
-    /// Witness point (can be identity)
-    pub q_point: Selector,
-    /// Witness non-identity point
-    pub q_point_non_id: Selector,
+    /// Witness point
+    witness_point: witness_point::Config,
```
daira, line 154: Use `witness_point_config` for consistency with `lookup_config`, etc.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -139,7 +139,7 @@ pub struct EccConfig {
     pub fixed_z: Column<Fixed>,
 
     /// Incomplete addition
-    pub q_add_incomplete: Selector,
+    add_incomplete: add_incomplete::Config,
```
daira, line 138: `add_incomplete_config`

src/circuit/gadget/ecc/chip.rs
```diff
@@ -142,7 +142,7 @@ pub struct EccConfig {
     add_incomplete: add_incomplete::Config,
 
     /// Complete addition
-    pub q_add: Selector,
+    add: add::Config,
```
daira, line 141: `add_config`

src/circuit/gadget/sinsemilla.rs
```diff
@@ -487,16 +487,15 @@ mod tests {
 
             let range_check = LookupRangeCheckConfig::configure(meta, advices[9], table_idx);
```
daira, line 488: `range_check_config`

src/circuit/gadget/ecc/chip.rs
```diff
@@ -145,9 +145,9 @@ pub struct EccConfig {
     add: add::Config,
 
     /// Variable-base scalar multiplication (hi half)
-    pub q_mul_hi: (Selector, Selector, Selector),
+    mul_hi: mul::incomplete::Config<{ mul::INCOMPLETE_HI_LEN }>,
```
daira: `mul_hi_config`

src/circuit/gadget/ecc/chip.rs
```diff
@@ -145,9 +145,9 @@ pub struct EccConfig {
...
     /// Variable-base scalar multiplication (lo half)
-    pub q_mul_lo: (Selector, Selector, Selector),
+    mul_lo: mul::incomplete::Config<{ mul::INCOMPLETE_LO_LEN }>,
```
daira: `mul_lo_config`

src/circuit/gadget/ecc/chip.rs
str4d: I disagree with @daira's suggestion; it will be accessed as `config.add`, so `config.add_config` is redundant.

daira: In that case `_config` should be removed from the existing names that have it.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -62,17 +66,34 @@ pub struct Config {
...
+        add_config: add::Config,
+        lookup_config: LookupRangeCheckConfig<pallas::Base, { sinsemilla::K }>,
+        advices: [Column<Advice>; 10],
+    ) -> Self {
+        let hi_config = incomplete::Config::configure(
+            meta, advices[9], advices[3], advices[0], advices[1], advices[4], advices[5],
+        );
+        let lo_config = incomplete::Config::configure(
+            meta, advices[6], advices[7], advices[0], advices[1], advices[8], advices[2],
+        );
```
daira, line 81: I know this hasn't changed from before this PR, but what's the motivation for this ordering?

src/circuit/gadget/ecc/chip/mul_fixed/base_field_elem.rs
```diff
@@ -46,8 +42,6 @@ impl From<&EccConfig> for Config {
             );
         }
 
-        assert_eq!(config.running_sum_config.z, config.super_config.window);
```
daira, line 49: Why is this assertion removed rather than moved?

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -33,41 +39,61 @@ const INCOMPLETE_RANGE: Range<usize> = 0..INCOMPLETE_LEN;
 // (It is a coincidence that k_{130} matches the boundary of the
 // overflow check described in [the book](https://zcash.github.io/halo2/design/gadgets/ecc/var-base-scalar-mul.html#overflow-check).)
 const INCOMPLETE_HI_RANGE: Range<usize> = 0..(INCOMPLETE_LEN / 2);
+const INCOMPLETE_HI_LEN: usize = INCOMPLETE_LEN / 2;
 
 // Bits k_{254} to k_{4} inclusive are used in incomplete addition.
 // The `lo` half is k_{129} to k_{4} inclusive (length 126 bits).
 const INCOMPLETE_LO_RANGE: Range<usize> = (INCOMPLETE_LEN / 2)..INCOMPLETE_LEN;
```
daira, line 46: Why is this range of length 125, when `INCOMPLETE_LO_LEN` is 126?

daira (approved):
utACK a09173a331394a77c004e17faa8bc0adafca8a4c with comments.

str4d (approved):
utACK after comment addressed.

src/circuit/gadget/ecc/chip.rs
str4d: I think the sub-sub-configs use `_config` in places, but the sub-configs don't. I agree we should be consistent.

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -33,10 +34,12 @@ const INCOMPLETE_RANGE: Range<usize> = 0..INCOMPLETE_LEN;
 // (It is a coincidence that k_{130} matches the boundary of the
 // overflow check described in [the book](https://zcash.github.io/halo2/design/gadgets/ecc/var-base-scalar-mul.html#overflow-check).)
 const INCOMPLETE_HI_RANGE: Range<usize> = 0..(INCOMPLETE_LEN / 2);
+pub const INCOMPLETE_HI_LEN: usize = INCOMPLETE_LEN / 2;
 
 // Bits k_{254} to k_{4} inclusive are used in incomplete addition.
 // The `lo` half is k_{129} to k_{4} inclusive (length 126 bits).
 const INCOMPLETE_LO_RANGE: Range<usize> = (INCOMPLETE_LEN / 2)..INCOMPLETE_LEN;
+pub const INCOMPLETE_LO_LEN: usize = (INCOMPLETE_LEN / 2) + 1;
```
str4d:
```suggestion
const INCOMPLETE_LO_RANGE: Range<usize> = INCOMPLETE_HI_LEN..INCOMPLETE_LEN;
pub const INCOMPLETE_LO_LEN: usize = INCOMPLETE_LEN - INCOMPLETE_HI_LEN;
```

src/circuit/gadget/ecc/chip/mul.rs
```diff
@@ -33,41 +39,61 @@ const INCOMPLETE_RANGE: Range<usize> = 0..INCOMPLETE_LEN;
 // (It is a coincidence that k_{130} matches the boundary of the
 // overflow check described in [the book](https://zcash.github.io/halo2/design/gadgets/ecc/var-base-scalar-mul.html#overflow-check).)
 const INCOMPLETE_HI_RANGE: Range<usize> = 0..(INCOMPLETE_LEN / 2);
+const INCOMPLETE_HI_LEN: usize = INCOMPLETE_LEN / 2;
 
 // Bits k_{254} to k_{4} inclusive are used in incomplete addition.
 // The `lo` half is k_{129} to k_{4} inclusive (length 126 bits).
 const INCOMPLETE_LO_RANGE: Range<usize> = (INCOMPLETE_LEN / 2)..INCOMPLETE_LEN;
+const INCOMPLETE_LO_LEN: usize = (INCOMPLETE_LEN / 2) + 1;
```
str4d, line 47:
```suggestion
const INCOMPLETE_LO_RANGE: Range<usize> = INCOMPLETE_HI_LEN..INCOMPLETE_LEN;
const INCOMPLETE_LO_LEN: usize = INCOMPLETE_LEN - INCOMPLETE_HI_LEN;
```

src/circuit/gadget/ecc/chip/mul.rs
str4d: I think it evolved organically as we optimised the chip? It does look a bit worse than it is because of the argument ordering to `configure()`.

PR #249, opened 2021-11-30
Replace `PartialEq, PartialOrd` with `ConstantTimeEq` on `{Extended}SpendingKey`

### LINKED ISSUE #241: Remove `PartialEq` and `PartialOrd` impls from `SpendingKey`, and add `ConstantTimeEq`
https://github.com/zcash/orchard/blob/68b790c7dadade049f44ad4aafa0ff71a3a10e91/src/keys.rs#L35-L36

These were added in https://github.com/zcash/orchard/pull/170 using automatic derives, which use non-constant-time logic. We should at a minimum be using constant-time comparisons, but I'd prefer that we do not provide `PartialEq` or `PartialOrd` impls as part of our public API, because no one should be using spending keys in a way that requires them (e.g. using them as keys in a `HashMap` or `BTreeMap`).

str4d:
Closes #241.

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -35,6 +35,12 @@ const ZIP32_PURPOSE: u32 = 32;
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct SpendingKey([u8; 32]);
 
+impl ConstantTimeEq for SpendingKey {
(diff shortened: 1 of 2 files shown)


ebfull (approved):

ebfull (approved):

daira (commented):
Post-hoc utACK.

PR #250, opened 2021-12-02
Migrate to halo2 version with `AssignedCell`

str4d:
Best reviewed by commit.

### DIFF
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -1,7 +1,7 @@
 use super::EccInstructions;
 use crate::{
     circuit::gadget::utilities::{
-        copy, lookup_range_check::LookupRangeCheckConfig, CellValue, UtilitiesInstructions, Var,
+        copy, lookup_range_check::LookupRangeCheckConfig, CellValue, UtilitiesInstructions,
     },
     constants::{self, NullifierK, OrchardFixedBasesFull, ValueCommitV},
     primitives::sinsemilla,
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -8,6 +8,7 @@ use crate::{
 };
 use arrayvec::ArrayVec;
 
+use ff::Field;
 use group::prime::PrimeCurveAffine;
 use halo2::{
     circuit::{Chip, Layouter},
@@ -50,7 +51,7 @@ impl EccPoint {
     pub fn point(&self) -> Option<pallas::Affine> {
         match (self.x.value(), self.y.value()) {
             (Some(x), Some(y)) => {
-                if x == pallas::Base::zero() && y == pallas::Base::zero() {
+                if x.is_zero_vartime() && y.is_zero_vartime() {
                     Some(pallas::Affine::identity())
                 } else {
                     Some(pallas::Affine::from_xy(x, y).unwrap())
@@ -72,7 +73,7 @@ impl EccPoint {
 
     #[cfg(test)]
     fn is_identity(&self) -> Option<bool> {
-        self.x.value().map(|x| x == pallas::Base::zero())
+        self.x.value().map(|x| x.is_zero_vartime())
     }
 }
 
@@ -102,7 +103,7 @@ impl NonIdentityEccPoint {
     pub fn point(&self) -> Option<pallas::Affine> {
         match (self.x.value(), self.y.value()) {
             (Some(x), Some(y)) => {
-                assert!(x != pallas::Base::zero() && y != pallas::Base::zero());
+                assert!(!x.is_zero_vartime() && !y.is_zero_vartime());
                 Some(pallas::Affine::from_xy(x, y).unwrap())
             }
             _ => None,
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -26,7 +26,7 @@ pub(super) mod witness_point;
 /// A curve point represented in affine (x, y) coordinates, or the
 /// identity represented as (0, 0).
 /// Each coordinate is assigned to a cell.
-#[derive(Copy, Clone, Debug)]
+#[derive(Clone, Debug)]
 pub struct EccPoint {
     /// x-coordinate
     x: CellValue<pallas::Base>,
@@ -62,12 +62,12 @@ impl EccPoint {
     /// The cell containing the affine short-Weierstrass x-coordinate,
     /// or 0 for the zero point.
     pub fn x(&self) -> CellValue<pallas::Base> {
-        self.x
+        self.x.clone()
     }
     /// The cell containing the affine short-Weierstrass y-coordinate,
     /// or 0 for the zero point.
     pub fn y(&self) -> CellValue<pallas::Base> {
-        self.y
+        self.y.clone()
     }
 
     #[cfg(test)]
@@ -78,7 +78,7 @@ impl EccPoint {
 
 /// A non-identity point represented in affine (x, y) coordinates.
 /// Each coordinate is assigned to a cell.
-#[derive(Copy, Clone, Debug)]
+#[derive(Clone, Debug)]
 pub struct NonIdentityEccPoint {
     /// x-coordinate
     x: CellValue<pallas::Base>,
@@ -110,11 +110,11 @@ impl NonIdentityEccPoint {
     }
     /// The cell containing the affine short-Weierstrass x-coordinate.
     pub fn x(&self) -> CellValue<pallas::Base> {
-        self.x
+        self.x.clone()
     }
     /// The cell containing the affine short-Weierstrass y-coordinate.
     pub fn y(&self) -> CellValue<pallas::Base> {
-        self.y
+        self.y.clone()
     }
 }
 
@@ -298,7 +298,7 @@ struct EccBaseFieldElemFixed {
 
 impl EccBaseFieldElemFixed {
     fn base_field_elem(&self) -> CellValue<pallas::Base> {
-        self.base_field_elem
+        self.base_field_elem.clone()
     }
 }
 
@@ -396,7 +396,7 @@ impl EccInstructions<pallas::Affine> for EccChip {
         let config = self.config().mul;
         config.assign(
             layouter.namespace(|| "variable-base scalar mul"),
-            *scalar,
+            scalar.clone(),
             base,
         )
     }
--- a/src/circuit/gadget/ecc/chip.rs
+++ b/src/circuit/gadget/ecc/chip.rs
@@ -1,7 +1,7 @@
 use super::EccInstructions;
 use crate::{
     circuit::gadget::utilities::{
-        lookup_range_check::LookupRangeCheckConfig, CellValue, UtilitiesInstructions,
+        lookup_range_check::LookupRangeCheckConfig, UtilitiesInstructions,
     },
     constants::{self, NullifierK, OrchardFixedBasesFull, ValueCommitV},
     primitives::sinsemilla,
@@ -11,7 +11,7 @@ use arrayvec::ArrayVec;
 use ff::Field;
 use group::prime::PrimeCurveAffine;
 use halo2::{
-    circuit::{Chip, Layouter},
+    circuit::{AssignedCell, Chip, Layouter},
     plonk::{Advice, Column, ConstraintSystem, Error, Fixed},
 };
 use pasta_curves::{arithmetic::CurveAffine, pallas};
@@ -30,9 +30,9 @@ pub(super) mod witness_point;
 #[derive(Clone, Debug)]
 pub struct EccPoint {
     /// x-coordinate
-    x: CellValue<pallas::Base>,
+    x: AssignedCell<pallas::Base, pallas::Base>,
     /// y-coordinate
-    y: CellValue<pallas::Base>,
+    y: AssignedCell<pallas::Base, pallas::Base>,
 }
 
 impl EccPoint {
@@ -41,8 +41,8 @@ impl EccPoint {
     /// This is an internal API that we only use where we know we have a valid curve point
     /// (specifically inside Sinsemilla).
     pub(in crate::circuit::gadget) fn from_coordinates_unchecked(
-        x: CellValue<pallas::Base>,
-        y: CellValue<pallas::Base>,
+        x: AssignedCell<pallas::Base, pallas::Base>,
+        y: AssignedCell<pallas::Base, pallas::Base>,
     ) -> Self {
         EccPoint { x, y }
     }
@@ -62,12 +62,12 @@ impl EccPoint {
     }
     /// The cell containing the affine short-Weierstrass x-coordinate,
     /// or 0 for the zero point.
-    pub fn x(&self) -> CellValue<pallas::Base> {
+    pub fn x(&self) -> AssignedCell<pallas::Base, pallas::Base> {
         self.x.clone()
     }
     /// The cell containing the affine short-Weierstrass y-coordinate,
     /// or 0 for the zero point.
-    pub fn y(&self) -> CellValue<pallas::Base> {
+    pub fn y(&self) -> AssignedCell<pallas::Base, pallas::Base> {
         self.y.clone()
     }
 
@@ -82,9 +82,9 @@ impl EccPoint {
 #[derive(Clone, Debug)]
 pub struct NonIdentityEccPoint {
     /// x-coordinate
-    x: CellValue<pallas::Base>,
+    x: AssignedCell<pallas::Base, pallas::Base>,
     /// y-coordinate
-    y: CellValue<pallas::Base>,
+    y: AssignedCell<pallas::Base, pallas::Base>,
 }
 
 impl NonIdentityEccPoint {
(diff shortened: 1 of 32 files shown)


src/circuit/gadget/poseidon.rs
```diff
@@ -144,7 +145,13 @@ impl<
         chip.initial_state(&mut layouter, &domain)
             .map(|state| Duplex {
                 chip,
-                sponge: Sponge::Absorbing([None; RATE]),
+                sponge: Sponge::Absorbing(
+                    (0..RATE)
+                        .map(|_| None)
+                        .collect::<Vec<_>>()
+                        .try_into()
+                        .unwrap(),
```
daira, line 152: Why is this change needed?

src/circuit/gadget/ecc/chip/add.rs
```diff
@@ -310,11 +310,11 @@ impl Config {
...
                         if x_p.is_zero_vartime() {
                             // 0 + Q = Q
-                            (x_q, y_q)
+                            (*x_q, *y_q)
                         } else if x_q.is_zero_vartime() {
                             // P + 0 = P
-                            (x_p, y_p)
-                        } else if (x_q == x_p) && (y_q == -y_p) {
+                            (*x_p, *y_p)
+                        } else if (x_q == x_p) && (*y_q == -y_p) {
```
daira, line 317: Why does only `y_q` change here?

src/circuit/gadget/ecc/chip/add_incomplete.rs
```diff
@@ -136,8 +136,8 @@ impl Config {
         )?;
 
         let result = NonIdentityEccPoint {
-            x: CellValue::<pallas::Base>::new(x_r_var, x_r),
-            y: CellValue::<pallas::Base>::new(y_r_var, y_r),
```
daira, line 139: Consider renaming these to `x_r_cell` and `y_r_cell` for consistency.

src/circuit/gadget/ecc/chip/add_incomplete.rs
```diff
@@ -136,8 +136,8 @@ impl Config {
...
+            x: x_r_var,
+            y: y_r_var,
```
daira, line 140: Consider renaming these to `x_r_cell` and `y_r_cell` for consistency.

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -183,7 +183,7 @@ impl Config {
 
                 Ok(EccPoint {
                     x: magnitude_mul.x,
-                    y: CellValue::new(y_var, y_val),
+                    y: y_var,
```
daira, line 183: Consider renaming this to `y_cell` for consistency.

src/circuit/gadget/ecc/chip/witness_point.rs
```diff
@@ -81,10 +81,7 @@ impl Config {
         let y_var =
             region.assign_advice(|| "y", self.y, offset, || y_val.ok_or(Error::Synthesis))?;
 
-        Ok((
-            CellValue::<pallas::Base>::new(x_var, x_val),
-            CellValue::<pallas::Base>::new(y_var, y_val),
-        ))
+        Ok((x_var, y_var))
```
daira, line 90: Consider renaming these to `x_cell` and `y_cell`.

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -46,11 +46,8 @@ impl SinsemillaChip {
...
-            let x_a = {
-                let cell =
-                    region.assign_advice_from_constant(|| "fixed x_q", config.x_a, offset, x_q)?;
-                CellValue::new(cell, Some(x_q))
-            };
+            let x_a =
+                region.assign_advice_from_constant(|| "fixed x_q", config.x_a, offset, x_q)?;
 
             x_a.into()
         };
```
daira, line 60:
```suggestion
        let mut x_a: X<pallas::Base> =
            region.assign_advice_from_constant(|| "fixed x_q", config.x_a, offset, x_q)?.into();
```

src/circuit/gadget/sinsemilla/chip/hash_to_point.rs
```diff
@@ -46,16 +53,13 @@ impl SinsemillaChip {
...
```
daira, line 60:
```
        let mut x_a: X<pallas::Base> =
            region.assign_advice_from_constant(|| "fixed x_q", config.x_a, offset, x_q)?.into();
```

src/circuit/gadget/poseidon.rs
str4d: Because of the removal of the `Copy` bound; `let x = [init_val; N];` requires that `init_val` can be copied.

src/circuit/gadget/ecc/chip/add_incomplete.rs
str4d: I was trying to keep the diff as small as possible 😅 but yes, I can go through and clean up variable names.

src/circuit/gadget/ecc/chip/mul/overflow.rs
```diff
@@ -135,16 +135,10 @@ impl Config {
...
-                copy(
-                    &mut region,
-                    || "copy z_130",
-                    self.advices[0],
-                    offset + 1,
-                    &*zs[130],
-                )?;
+                zs[130].copy_advice(|| "copy z_130", &mut region, self.advices[0], offset + 1)?;
 
                 // Witness η = inv0(z_130), where inv0(x) = 0 if x = 0, 1/x otherwise
```
daira, line 144: (For future reference) It's a common pattern to use inv0 to implement nonzero constraints. I'd really like to have an easy way to defer these inverses so that they can be *automatically* batched. To do that, the chip code shouldn't compute the inverse itself, it should call some method on the region that assigns an inverse. Then any use of an inverted value triggers the batch inversion.

src/circuit/gadget/sinsemilla/commit_ivk.rs
```diff
@@ -474,28 +474,24 @@ impl CommitIvkConfig {
                 {
                     let offset = 0;
                     // Copy in `ak`
```
daira, line 488: Consider deleting the "Copy in ..." and "Witness ..." comments in this method, which don't say anything that isn't obvious from the code.

src/circuit/gadget/ecc/chip/mul/overflow.rs
str4d: We already have support for this during witness assignment: that's what `Assigned<F>` represents.

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -73,7 +73,10 @@ impl Config {
         &self,
         region: &mut Region<'_, pallas::Base>,
         offset: usize,
-        magnitude_sign: (CellValue<pallas::Base>, CellValue<pallas::Base>),
+        magnitude_sign: (
+            AssignedCell<pallas::Base, pallas::Base>,
+            AssignedCell<pallas::Base, pallas::Base>,
```
daira, line 78: These could have a narrower value type.

src/circuit/gadget/ecc/chip.rs
```diff
@@ -349,9 +350,11 @@ pub struct EccScalarFixed {
 /// k_21 must be a single bit, i.e. 0 or 1.
 #[derive(Clone, Debug)]
 pub struct EccScalarFixedShort {
-    magnitude: CellValue<pallas::Base>,
-    sign: CellValue<pallas::Base>,
-    running_sum: ArrayVec<CellValue<pallas::Base>, { constants::NUM_WINDOWS_SHORT + 1 }>,
+    magnitude: AssignedCell<pallas::Base, pallas::Base>,
```
daira, line 353:
```suggestion
    // TODO: Make V a `u64`.
    magnitude: AssignedCell<pallas::Base, pallas::Base>,
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -365,23 +368,24 @@ pub struct EccScalarFixedShort {
...
+    fn base_field_elem(&self) -> AssignedCell<pallas::Base, pallas::Base> {
+        self.base_field_elem.clone()
     }
 }
 
 impl EccInstructions<pallas::Affine> for EccChip {
     type ScalarFixed = EccScalarFixed;
     type ScalarFixedShort = EccScalarFixedShort;
-    type ScalarVar = CellValue<pallas::Base>;
+    // TODO: V = pallas::Scalar?
```
daira, line 384: No; ScalarVar is used for variable-base scalar multiplication where the scalar is represented as a base element in F<sub>p</sub>, and the implementation assumes that p < q. So `V = pallas::Base` is correct.

```suggestion
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -491,7 +495,11 @@ impl EccInstructions<pallas::Affine> for EccChip {
     fn mul_fixed_short(
         &self,
         layouter: &mut impl Layouter<pallas::Base>,
-        magnitude_sign: (CellValue<pallas::Base>, CellValue<pallas::Base>),
+        // TODO: Return AssignedCell<Sign, pallas::Base>
```
daira, line 498:
```suggestion
        // TODO: Return (AssignedCell<u64, pallas::Base>, AssignedCell<Sign, pallas::Base>)
```

src/circuit/gadget/ecc/chip.rs
```diff
@@ -491,7 +495,11 @@ impl EccInstructions<pallas::Affine> for EccChip {
...
+        magnitude_sign: (
+            AssignedCell<pallas::Base, pallas::Base>,
+            AssignedCell<pallas::Base, pallas::Base>,
+        ),
```
daira, line 502: I suggest making `MagnitudeSign` a typedef for
`(AssignedCell<pallas::Base, pallas::Base>, AssignedCell<pallas::Base, pallas::Base>)` now, then changing it to `(AssignedCell<u64, pallas::Base>, AssignedCell<Sign, pallas::Base>)` later.

```suggestion
        magnitude_sign: MagnitudeSign,
```

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -75,15 +73,18 @@ impl Config {
...
+        ),
```
daira, line 79:
```suggestion
        magnitude_sign: MagnitudeSign,
```

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -99,7 +100,10 @@ impl Config {
     pub fn assign(
         &self,
         mut layouter: impl Layouter<pallas::Base>,
-        magnitude_sign: (CellValue<pallas::Base>, CellValue<pallas::Base>),
+        magnitude_sign: (
+            AssignedCell<pallas::Base, pallas::Base>,
+            AssignedCell<pallas::Base, pallas::Base>,
+        ),
```
daira, line 106:
```suggestion
        magnitude_sign: MagnitudeSign,
```

src/circuit/gadget/ecc/chip/mul_fixed/short.rs
```diff
@@ -267,7 +264,13 @@ pub mod tests {
...
             magnitude: pallas::Base,
             sign: pallas::Base,
-        ) -> Result<(CellValue<pallas::Base>, CellValue<pallas::Base>), Error> {
+        ) -> Result<
+            (
+                AssignedCell<pallas::Base, pallas::Base>,
+                AssignedCell<pallas::Base, pallas::Base>,
+            ),
+            Error,
+        > {
```
daira, line 273:
```suggestion
        ) -> Result<MagnitudeSign, Error> {
```

src/circuit/gadget/ecc/chip/witness_point.rs
```diff
@@ -70,7 +70,13 @@ impl Config {
...
         offset: usize,
         region: &mut Region<'_, pallas::Base>,
-    ) -> Result<(CellValue<pallas::Base>, CellValue<pallas::Base>), Error> {
+    ) -> Result<
+        (
+            AssignedCell<pallas::Base, pallas::Base>,
+            AssignedCell<pallas::Base, pallas::Base>,
+        ),
+        Error,
+    > {
```
daira, line 79:
```suggestion
    ) -> Result<MagnitudeSign, Error> {
```

src/circuit/gadget/sinsemilla/chip.rs
```diff
@@ -239,14 +238,14 @@ impl SinsemillaChip {
 impl SinsemillaInstructions<pallas::Affine, { sinsemilla::K }, { sinsemilla::C }>
     for SinsemillaChip
 {
-    type CellValue = CellValue<pallas::Base>;
+    type CellValue = AssignedCell<pallas::Base, pallas::Base>;
```
daira, line 241: Consider renaming this.

daira (approved):
utACK with comments/suggestions.

src/circuit/gadget/ecc/chip/mul/overflow.rs
therealyingtong: So instead of (for example) `z_130.invert().unwrap()` should we do `(pallas::Base::one(), z_130).into()`?

src/circuit/gadget/ecc/chip/add.rs
therealyingtong: Could be because `-y_p` is dereferenced?

src/circuit/gadget/ecc/chip/mul/overflow.rs
str4d: `Assigned::from(z_130).invert()` would be clearer.

str4d:
[Rebased on #237](https://github.com/zcash/orchard/compare/cd1c8f1c45ab6880986cdf64924225dec22394a5..d12fc94839c031665afa614f56ea4c7fdabccc6a) to fix merge conflicts (which were all in `use` statements). I'll address @daira's comments next.

src/circuit/gadget/ecc/chip/witness_point.rs
str4d: This one is a point, not a magnitude-sign. Which makes your point :P

str4d:
[Force-pushed](https://github.com/zcash/orchard/compare/d12fc94839c031665afa614f56ea4c7fdabccc6a..bacdf67428f8535e0cfcdfde7529fa440dad6496) to remove an incorrect TODO comment.

src/circuit/gadget/ecc/chip/mul/overflow.rs
str4d: I tried this locally, and discovered a bug in the `AssignedCell` `halo2` PR! Fixed here: https://github.com/zcash/halo2/pull/417 (which I'll immediately start depending on in this PR).

str4d: Hmm, actually this gets more complex to migrate to in all but a single case, so I'll do this in a separate PR.

src/circuit/gadget/ecc/chip/add_incomplete.rs
str4d: Decided to not do this in this PR, it's already large enough. This can be done post-migration to `halo2_gadgets`.

src/circuit/gadget/ecc/chip/mul_fixed/full_width.rs
```diff
@@ -294,7 +295,9 @@ pub mod tests {
         {
             let scalar_fixed = pallas::Scalar::zero();
             let (result, _) = base.mul(layouter.namespace(|| "mul by zero"), Some(scalar_fixed))?;
-            assert!(result.inner().is_identity().unwrap());
+            if let Some(is_identity) = result.inner().is_identity() {
+                assert!(is_identity);
+            }
```
ebfull, line 300: We should panic on `else` here right?

ebfull (approved):

src/circuit/gadget/ecc/chip/mul_fixed/full_width.rs
str4d: No; that is what the prior behaviour was, and it prevented the test circuit from being used in non-witness contexts like keygen or dev tooling. We can only enforce this assertion if we have a witness to enforce it on.

src/circuit/gadget/ecc/chip/add.rs
daira: Yes, with some wrinkles. The LHS and RHS of `==` are [place expression contexts](https://doc.rust-lang.org/reference/expressions.html#place-expressions-and-value-expressions), which means they are implicitly borrowed. But because *both* sides are implicitly borrowed, we end up with `&y_q == &-y_p`, in which the reference levels are still mismatched (although the error message says `no implementation for `&pasta_curves::Fp == pasta_curves::Fp` and doesn't mention the implicit borrowing, which I may report as a bug).

PR #251, opened 2021-12-03
Migrate to `halo2::plonk::Constraints` helper

str4d:
Best reviewed with whitespace hidden.

### DIFF
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -6,7 +6,8 @@ use group::{Curve, GroupEncoding};
 use halo2_proofs::{
     circuit::{floor_planner, AssignedCell, Layouter},
     plonk::{
-        self, Advice, Column, Expression, Instance as InstanceColumn, Selector, SingleVerifier,
+        self, Advice, Column, Constraints, Expression, Instance as InstanceColumn, Selector,
+        SingleVerifier,
     },
     poly::Rotation,
     transcript::{Blake2bRead, Blake2bWrite},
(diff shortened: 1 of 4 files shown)


src/circuit.rs
```diff
@@ -162,22 +162,24 @@ impl plonk::Circuit<pallas::Base> for Circuit {
...
-                ("v_old = 0 or enable_spends = 1", v_old * not_enable_spends),
-                (
-                    "v_new = 0 or enable_outputs = 1",
-                    v_new * not_enable_outputs,
-                ),
-            ])
-            .map(move |(name, poly)| (name, q_orchard.clone() * poly))
+            Constraints::with_selector(
+                q_orchard,
+                std::array::IntoIter::new([
```
daira, line 167: I'm confused as to why we need the explicit `std::array::IntoIter::new`, given that `with_selector` takes `IntoIterator<Item = C>` which is implemented by references to arrays ([and by arrays](https://doc.rust-lang.org/edition-guide/rust-2021/IntoIterator-for-arrays.html)). On the other hand, if I try just passing a reference to array then it doesn't work:
```
error[E0277]: the trait bound `halo2::plonk::Constraint<pasta_curves::Fp>: From<&(&str, halo2::plonk::Expression<pasta_curves::Fp>)>` is not satisfied
   --> src/circuit.rs:165:13
    |
165 |             Constraints::with_selector(
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<&(&str, halo2::plonk::Expression<pasta_curves::Fp>)>` is not implemented for `halo2::plonk::Constraint<pasta_curves::Fp>`
    |
    = help: the following implementations were found:
              <halo2::plonk::Constraint<F> as From<(&'static str, halo2::plonk::Expression<F>)>>
              <halo2::plonk::Constraint<F> as From<halo2::plonk::Expression<F>>>
    = note: required because of the requirements on the impl of `Into<halo2::plonk::Constraint<pasta_curves::Fp>>` for `&(&str, halo2::plonk::Expression<pasta_curves::Fp>)`
    = note: required by `Constraints::<F, C, Iter>::with_selector`
```
(and similarly for an array).

str4d: I just gave it a go, and every "simpler" array variant I can come up with fails as an input to `Constraints`, because they all implicitly take the array as a reference, which fails because the reference (returned inside `Constraints`) would need to outlive the array (dropped at the end of the method).

However, I also confirmed that we don't actually need to declare Rust 2021 edition; as soon as we reach MSRV of Rust 1.53, we can simply drop the `array::IntoIter::new()` and return the array directly. That also means that other users of `halo2` can similarly use arrays directly right now, if their MSRV is high enough.

str4d: Opened https://github.com/zcash/orchard/issues/253.

str4d: This is addressed by #315.

src/circuit.rs
```diff
@@ -163,22 +164,24 @@ impl plonk::Circuit<pallas::Base> for Circuit {
...
```
daira, line 169: We should be able to remove these `std::array::IntoIter::new`s now.

src/circuit.rs
daira: I'm confused, #315 is merged but this PR is still using `std::array::IntoIter::new()`.

daira (approved):
utACK modulo comment.

src/circuit.rs
str4d: That's because I needed to rebase this after it merged, which I've now done.

Issue #261, opened 2021-12-13
Reject the identity in `SpendValidatingKey::from_bytes`


str4d:
`ak_P` is not allowed to be the identity in the Orchard protocol. In the `orchard` crate's type system, we enforce this by construction: `SpendingKey` can't be constructed unless it produces a non-zero `ask`, and since the Pallas group has prime order, this means it will produce a non-identity `ak_P`.

However, when parsing an Orchard FVK, we call `SpendValidatingKey::from_bytes`, which enforces the `ak`/`ak_P` equivalence, but does _not_ currently reject the identity:

https://github.com/zcash/orchard/blob/99b767a3a134028d77684607f29caee11ca22654/src/keys.rs#L181-L193

This can't cause many problems, because if the user were to construct an FVK in this way, it could not be used to generate proofs. But for consistency we should ensure we reject the identity here.

(Note that a RedPallas verification key _is_ allowed to be the identity; we reflect this distinction in the type system by having `SpendValidatingKey::randomize` return a `redpallas::VerificationKey`, _not_ a `SpendValidatingKey`.)

PR #262, opened 2021-12-15
Reject the identity in `SpendValidatingKey::from_bytes`

str4d:
`ak_P` is not allowed to be the identity in the Orchard protocol. We
were enforcing this by construction in most places, except for the
parsing of an Orchard full viewing key.

Closes zcash/orchard#261.


therealyingtong (approved):
utACK.

daira (approved):
utACK

ebfull (approved):

PR #269, opened 2022-01-05
Pin `pprof = 0.6.1`.

therealyingtong:


Cargo.toml
```diff
@@ -31,9 +31,11 @@ ff = "0.11"
 fpe = "0.5"
 group = "0.11"
 halo2 = "=0.1.0-beta.1"
+inferno = "=0.10.7"
 lazy_static = "1"
 memuse = { version = "0.2", features = ["nonempty"] }
 pasta_curves = "0.2.1"
+pprof = "=0.6.1"
```
str4d, line 38: `pprof` is a Unix-only dev-dependency, and we don't need to pin `inferno`. Undo these changes, and instead edit the `pprof` dependency below to pin `0.6.1`.

str4d (changes requested):

str4d (approved):

PR #270, opened 2022-01-06
Derive internal keys.

therealyingtong:
Closes zcash/orchard#272.

Uses test vectors from https://github.com/zcash-hackworks/zcash-test-vectors/pull/75/.


src/keys.rs
```diff
@@ -256,6 +269,12 @@ impl From<&SpendingKey> for CommitIvkRandomness {
     }
 }
 
+impl From<&InternalSpendingKey> for CommitIvkRandomness {
+    fn from(sk: &InternalSpendingKey) -> Self {
+        CommitIvkRandomness(to_scalar(PrfExpand::OrchardRivkInternal.expand(&sk.0)))
```
daira, line 274: This isn't correct. rivk<sub>internal</sub> is derived from ak, nk, and rivk of the external full viewing key, not directly from the spending key.

src/keys.rs
```diff
@@ -399,6 +418,22 @@ impl FullViewingKey {
...
     }
+
+    /// Derives an internal full viewing key from a full viewing key, as specified in [ZIP32][orchardinternalfullviewingkey]
+    ///
+    /// [orchardinternalfullviewingkey]: https://zips.z.cash/zip-0032#orchard-internal-key-derivation
+    pub fn derive_internal(&self) -> Option<Self> {
+        let k = self.rivk().to_bytes();
+        let rivk_internal = PrfExpand::OrchardRivkInternal
+            .with_ad_slices(&k, &[&self.ak.clone().to_bytes(), &self.nk().to_bytes()]);
+        let rivk_internal = CommitIvkRandomness::from_bytes(&rivk_internal)?;
```
daira, line 429: This is incorrect too; it's missing a ToScalar<sup>Orchard</sup> call.

daira (changes requested):
Please delete `InternalSpendingKey` and fix the rivk<sub>internal</sub> derivation to use ToScalar<sup>Orchard</sup>.

str4d (approved):
utACK. I verified that the test vectors match https://github.com/zcash-hackworks/zcash-test-vectors/blob/6c2bac7b73e345d311cffb044522fc453f40cd73/test-vectors/rust/orchard_key_components.rs

nuttycom (approved):
utACK

daira (approved):
utACK. I haven't checked that the test vectors match the generated ones.

PR #273, opened 2022-01-20
Migrate to pasta_curves 0.3

str4d:
The MSRV is now 1.54.0, because reddsa 0.2.0 included a fix to its nightly CI that inadvertently bumped its MSRV.

This also brings in the 20% prover performance improvement from zcash/halo2#447.


src/circuit/gadget/ecc/chip/mul_fixed.rs
```diff
@@ -557,6 +557,8 @@ impl ScalarFixed {
             .iter()
             .map(|window| {
                 if let Some(window) = window {
+                    // TODO: Remove this trait dependency
+                    use pasta_curves::arithmetic::SqrtRatio;
                     let window = window.get_lower_32() as usize;
```
str4d, line 562: This dependency should not exist, but appears to have crept in because `FieldExt::get_lower_32` existed and was easy to use. We should find some other way to extract the window.

str4d:
Rebased to include a version of `halo2` (now `halo2_proofs`) that included a fix to the `poly::Evaluator` logic (https://github.com/zcash/halo2/pull/471) that was causing macOS CI to fail. I'll probably rebase this PR again once #186 merges and `halo2_gadgets` is extracted.

str4d:
Cherry-picked into #217.

Issue #203, opened 2021-09-21
Decide whether we will randomise or sort the order of spends and outputs in Actions, and implement if so


str4d:
https://github.com/zcash/orchard/blob/f8280c98a3d0e41b8ff5b7f615802bd197f781e1/src/builder.rs#L290-L292

We did this for Sapling spends and outputs (and could do so independently easily), and there is a SHOULD in [section 4.7.3](https://zips.z.cash/protocol/protocol.pdf#orchardsend) for Orchard Actions. But adding this does complicate the builder API (the caller either needs to be given the randomisation map to find their outputs, or needs to trial-decrypt them all after making the transaction).

str4d:
Another idea that @daira and I had independently, was that we could just sort the spends by the byte encoding of their nullifiers, and the outputs by the byte encoding of their commitments. This is equivalent to a random shuffle from the perspective of a transaction recipient, as note commitments are themselves random (via `rcm`).

We can also do this for Sapling; in that case, we should only do this for v5 transactions, so that the "wallet creator" information leaked by Sapling spends and outputs being visibly sorted vs random, is coincident with "transaction version" information.

str4d:
On the topic of returning a mapping, I think we should _not_ do this, and just require the caller to trial-decrypt the resulting transaction. This is for two reasons:
- Wallets will end up doing this anyway when they see the transaction again on-chain, so it's at most doubling the work they do, but only relative to the number of transactions they create, and that could likely be optimised out.
- For Sapling, the mapping was from the "order of calls to `AddSaplingRecipient`" to "Sapling outputs". With Unified Addresses, this mapping becomes more complex, because an `AddUnifiedRecipient` API call may or may not result in a Sapling or Orchard recipient addition.

str4d:
> Another idea that @daira and I had independently, was that we could just sort the spends by the byte encoding of their nullifiers, and the outputs by the byte encoding of their commitments. This is equivalent to a random shuffle from the perspective of a transaction recipient, as note commitments are themselves random (via `rcm`).

I realised that we can't do this for outputs, because in Orchard an output commitment depends on the nullifier of the corresponding spent note. We therefore cannot use commitment ordering as a means of ordering outputs. We could instead use the ciphertexts, but a) those involve random data so they wouldn't be deterministic orderings anyway, and b) it would require completely destroying the type safety we currently have with the `Note` and `OrchardNoteEncryption` types.

[...]

PR #278, opened 2022-02-07
Shuffle spends and recipients before pairing them into Actions

str4d:
Callers cannot assume that any specific output corresponds to a specific
Orchard recipient, and must trial-decrypt all outputs to find the ones
belonging to them. This is consistent with higher-layer semantics like
having Unified Addresses as recipients (where the mapping from recipient
to a specific output would become much more complex).

Closes zcash/orchard#203.

Co-authored-by: Daira Hopwood <daira@jacaranda.org>

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -7,7 +7,7 @@ use ff::Field;
 use group::GroupEncoding;
 use nonempty::NonEmpty;
 use pasta_curves::pallas;
-use rand::{CryptoRng, RngCore};
+use rand::{prelude::SliceRandom, CryptoRng, RngCore};
 
 use crate::{
     address::Address,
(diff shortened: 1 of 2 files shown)


src/builder.rs
```diff
@@ -307,6 +303,12 @@ impl Builder {
                     .take(num_actions - num_recipients),
             );
 
+            // Shuffle the spends and recipients, so that learning the position of a
+            // specific spent note or output note doesn't reveal anything on its own about
+            // the meaning of that note in the transaction context.
+            self.spends.shuffle(&mut rng);
+            self.recipients.shuffle(&mut rng);
```
daira, line 310: I verified that the [algorithm used in rand 0.8.4](https://github.com/rust-random/rand/blob/0.8.4/src/seq/mod.rs#L586-L592) is a correctly implemented Durstenfeld shuffle, using `rng.gen_range` which is specified to be uniform. (The [implementation of that](https://github.com/rust-random/rand/blob/0.8.4/src/distributions/uniform.rs) is complicated but ends up using rejection sampling.)

daira (approved):
utACK. Needs at least one more review because I paired with @str4d on it.

daira (approved):
utACK c1447d6af2458a6f6c3c2e08d05c2e9f5e548be8

nuttycom (approved):
utACK

PR #282, opened 2022-02-12
Add `Clone` impls to various structs

str4d:
This enables `InProgress<Unproven, Unauthorized>: Clone`, which allows
the bundle returned by `Builder::build` to be cloned. In pure-Rust
wallets this should not be necessary, but it is required for `zcashd`
due to FFI-crossing.


nuttycom (approved):
utACK

PR #285, opened 2022-02-15
Functional test for building and verifying bundles

str4d:
This also makes some changes to the public API that are necessary to enable an `orchard` user to perform this workflow themselves.

### DIFF
--- a/src/bundle.rs
+++ b/src/bundle.rs
@@ -2,6 +2,7 @@
 
 pub mod commitments;
 
+use std::convert::TryInto;
 use std::io;
 
 use blake2b_simd::Hash as Blake2bHash;
@@ -298,15 +299,6 @@ impl<T: Authorization, V> Bundle<T, V> {
         &self.authorization
     }
 
-    /// Computes a commitment to the effects of this bundle, suitable for inclusion within
-    /// a transaction ID.
-    pub fn commitment<'a>(&'a self) -> BundleCommitment
-    where
-        i64: From<&'a V>,
-    {
-        BundleCommitment(hash_bundle_txid_data(self))
-    }
-
     /// Construct a new bundle by applying a transformation that might fail
     /// to the value balance.
     pub fn try_map_value_balance<V0, E, F: FnOnce(V) -> Result<V0, E>>(
@@ -405,7 +397,13 @@ impl<T: Authorization, V> Bundle<T, V> {
     }
 }
 
-impl<T: Authorization, V: Copy + Into<ValueSum>> Bundle<T, V> {
+impl<T: Authorization, V: Copy + Into<i64>> Bundle<T, V> {
+    /// Computes a commitment to the effects of this bundle, suitable for inclusion within
+    /// a transaction ID.
+    pub fn commitment(&self) -> BundleCommitment {
+        BundleCommitment(hash_bundle_txid_data(self))
+    }
+
     /// Returns the transaction binding validating key for this bundle.
     ///
     /// This can be used to validate the [`Authorized::binding_signature`] returned from
@@ -416,7 +414,10 @@ impl<T: Authorization, V: Copy + Into<ValueSum>> Bundle<T, V> {
             .iter()
             .map(|a| a.cv_net())
             .sum::<ValueCommitment>()
-            - ValueCommitment::derive(self.value_balance.into(), ValueCommitTrapdoor::zero()))
+            - ValueCommitment::derive(
+                ValueSum::from_raw(self.value_balance.into()),
+                ValueCommitTrapdoor::zero(),
+            ))
         .into_bvk()
     }
 }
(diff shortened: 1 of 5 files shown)


src/bundle.rs
```diff
@@ -405,7 +397,13 @@ impl<T: Authorization, V> Bundle<T, V> {
     }
 }
 
-impl<T: Authorization, V: Copy + Into<ValueSum>> Bundle<T, V> {
+impl<T: Authorization, V: Copy + Into<i64>> Bundle<T, V> {
+    /// Computes a commitment to the effects of this bundle, suitable for inclusion within
+    /// a transaction ID.
+    pub fn commitment(&self) -> BundleCommitment {
```
str4d, line 403:
```
### Changed
- `Bundle::commitment` now requires the bound `V: Copy + Into<i64>` instead of
  `i64: From<&'a V>`.

File: src/bundle.rs [hunk 2]
```diff
@@ -405,7 +397,13 @@ impl<T: Authorization, V> Bundle<T, V> {
     }
 }
 
-impl<T: Authorization, V: Copy + Into<ValueSum>> Bundle<T, V> {
+impl<T: Authorization, V: Copy + Into<i64>> Bundle<T, V> {
```
str4d, line 400:
```markdown
### Changed
- `Bundle::binding_validating_key` now requires the bound `V: Into<i64>` instead
  of `V: Into<ValueSum>`.
```

src/bundle.rs
```diff
@@ -500,6 +501,13 @@ impl<V: DynamicUsage> DynamicUsage for Bundle<Authorized, V> {
 #[derive(Debug)]
 pub struct BundleCommitment(pub Blake2bHash);
 
+impl From<BundleCommitment> for [u8; 32] {
```
str4d, line 504:
```markdown
### Added
- `impl From<BundleCommitment> for [u8; 32]`
```

src/bundle/commitments.rs
```diff
@@ -28,10 +28,9 @@ fn hasher(personal: &[u8; 16]) -> State {
 /// personalized with ZCASH_ORCHARD_ACTIONS_HASH_PERSONALIZATION
 ///
 /// [zip244]: https://zips.z.cash/zip-0244
-pub fn hash_bundle_txid_data<'a, A: Authorization, V>(bundle: &'a Bundle<A, V>) -> Blake2bHash
-where
-    i64: From<&'a V>,
-{
+pub fn hash_bundle_txid_data<A: Authorization, V: Copy + Into<i64>>(
+    bundle: &Bundle<A, V>,
+) -> Blake2bHash {
```
str4d, line 33: This change should go into the changelog, but I also don't see any usage of this outside of this crate (we use `Bundle::commitment` instead), so I'd argue we should actually remove it from the crate API. cc @nuttycom.

src/primitives/redpallas.rs
```diff
@@ -147,6 +147,13 @@ impl VerificationKey<Binding> {
     }
 }
 
+impl<T: SigType> VerificationKey<T> {
+    /// Verifies a purported `signature` over `msg` made by this verification key.
+    pub fn verify(&self, msg: &[u8], signature: &Signature<T>) -> Result<(), reddsa::Error> {
```
str4d, line 152:
```markdown
### Added
- `orchard::primitives::redpallas::VerificationKey::verify`
```

src/tree.rs
```diff
@@ -107,11 +107,19 @@ impl MerklePath {
...
             position,
-            auth_path: gen_const_array_with_default(MerkleHashOrchard::empty_leaf(), |i| {
+            gen_const_array_with_default(MerkleHashOrchard::empty_leaf(), |i| {
                 MerkleHashOrchard(auth_path[i])
             }),
+        )
+    }
+
+    /// Instantiates a new Merkle path given a leaf position and authentication path.
+    pub fn from_parts(position: u32, auth_path: [MerkleHashOrchard; MERKLE_DEPTH_ORCHARD]) -> Self {
```
str4d, line 119:
```markdown
### Added
- `MerklePath::from_parts`
```
This is a slightly awkward API (like all "from parts" APIs) in that the caller could construct this with an invalid `(position, auth_path)` tuple. But the effects of that is relatively limited: if the resulting path is used in `Builder` (which is the only in-crate consumer of `MerklePath`), it will result in a bundle with an invalid proof.

str4d:
[Force-pushed](https://github.com/zcash/orchard/compare/5dcff9e1b9285ed7fa2cacf03556b6d2b9e0b38c..f21b993db5ec7f8bd1503e95019e10e252424a78) a no-op to re-run CI (I accidentally targeted `main` instead of the non-consensus branch, so CI was running against the wrong branch).

src/bundle.rs
nuttycom: What distinguishes how we choose between providing `as_bytes`, `to_repr`, or other such methods and providing a `From` impl? After all the recent discussions we've had on this issue, I'd like us to have a standard style guide to refer to.

src/bundle/commitments.rs
nuttycom: Agreed, this seems like it could be crate-private.

src/tree.rs
nuttycom: Yeah. Constructing from parts is the lowest-common-denominator and relies only on builtin types; any additional structure we might lay over this would more or less just obfuscate the fundamental operation you're having to implement anyway.

nuttycom (approved):
utACK

src/bundle.rs
str4d: https://rust-lang.github.io/api-guidelines/checklist.html is what I generally refer to. Specifically:
- [Ad-hoc conversions follow `as_`, `to_`, `into_` conventions](https://rust-lang.github.io/api-guidelines/naming.html#c-conv)
- [Conversions use the standard traits `From`, `AsRef`, `AsMut`](https://rust-lang.github.io/api-guidelines/interoperability.html#c-conv-traits)

PR #289, opened 2022-02-21
Add `FullViewingKey::derive_internal`

str4d:
This is identical to the changes introduced in zcash/orchard#270, except that the output is non-optional (since the derivation is non-fallible). This PR is needed because that PR was not based on non-circuit changes, but we need it before bringing the circuit changes into `zcashd`.


nuttycom (approved):
utACK. I have verified that this is the same set of changes as have been merged to `main`.

daira (approved):
utACK

PR #292, opened 2022-02-28
Add missing `Debug` trait bounds for `Builder` components

str4d:
All relevant types have `Debug` impls, but some of the trait and method impls were lacking `Debug` bounds on their generic types. This prevented `Debug` impls being used on the overall partially-constructed `Bundle` types.


nuttycom (approved):
utACK

daira (commented):
Post-hoc ACK

PR #296, opened 2022-03-15
spec.rs: Check that commit_ivk returns a nonzero base.

### LINKED ISSUE #294: Fix `spec::commit_ivk` to handle partial-completeness of SinsemillaShortCommit
https://github.com/zcash/orchard/blob/3ddf6c49f7484ed1295bd5351317bbfe49e14472/src/spec.rs#L187-L193

The above argument relied on the fact that Sinsemilla was uniformly defined in terms of incomplete addition, and so SinsemillaCommit would never produce the identity (meaning that SinsemillaShortCommit would never produce zero). However, shortly after the QEDIT audit, ECC discovered two circuit bugs that led to us making two changes:

- We redefined g<sub>d</sub> and pk<sub>d</sub> to be non-identity points. g<sub>d</sub> needs to be non-identity in order to ensure that only a single `ivk` can be used with a committed note (to produce a single nullifier), which is necessary for balance. pk<sub>d</sub> on the other hand was set to be a non-identity point because `ivk` was defined as a non-identity scalar (per the above rationale), and so the identity could never be produced from `[ivk]` g<sub>d</sub>.
- We redefined SinsemillaCommit to use complete addition for the final `+ [r] R` step, because the case `r = 0` can be forced by an attacker. This in turn makes it possible for SinsemillaCommit to return the identity (as well as returning ⊥ if an edge case of incomplete addition is ever reached in the inner SinsemillaHash call).

The interaction of these two changes meant that `ivk` could now be zero (because Extract<sub>P</sub><sup>⊥</sup> maps the identity to zero), but in a pairing, @daira and I confirmed that the circuit now constrains `ivk != 0` (indirectly, because `pk_d_old` and `g_d_old` are both constrained to be non-identity). Previously neither of these were the case. This interaction is somewhat unfortunate (had we identified it at the time, we might have decided to allow pk<sub>d</sub> to remain a possibly-identity point), but it is too late to change the circuit now (as among other reasons, allowing pk<sub>d</sub> to be the identity would require additional analysis).

So what we need to do here is replace `.map(NonZeroPallasBase::guaranteed)` with `.and_then(NonZeroPallasBase::from_base)` (i.e. mapping the identity to an error). Note that we also return an error from this outside-circuit function if `sinsemilla::CommitDomain::short_commit` returns ⊥, which in the circuit is mapped to "circuit is satisfied"; we collapse both cases into the same error case, which is fine because we don't want users of the `orchard` crate to encounter either case.

therealyingtong:
Closes #294

### DIFF
--- a/src/spec.rs
+++ b/src/spec.rs
@@ -184,12 +184,20 @@ pub(crate) fn commit_ivk(
                 .chain(nk.to_le_bits().iter().by_val().take(L_ORCHARD_BASE)),
             rivk,
         )
-        // Commit^ivk.Output is specified as [1..q_P] ∪ {⊥}. We get this from
-        // sinsemilla::CommitDomain::short_commit by construction:
+        // sinsemilla::CommitDomain::short_commit returns a value in range [0..q_P] ∪ {⊥}:
+        // - sinsemilla::HashDomain::hash_to_point uses incomplete addition and returns a
+        //   point in P* ∪ {⊥}.
+        // - sinsemilla::CommitDomain::commit applies a final complete addition step and
+        //   returns a point in P ∪ {⊥}.
         // - 0 is not a valid x-coordinate for any Pallas point.
         // - sinsemilla::CommitDomain::short_commit calls extract_p_bottom, which replaces
-        //   the identity (which has no affine coordinates) with 0. but Sinsemilla is
-        //   defined using incomplete addition, and thus will never produce the identity.
+        //   the identity (which has no affine coordinates) with 0.
+        //
+        // Commit^ivk.Output is specified as [1..q_P] ∪ {⊥}, so we explicitly check for 0
+        // and map it to None. Note that we are collapsing this case (which is rejected by
+        // the circuit) with ⊥ (which the circuit explicitly allows for efficiency); this
+        // is fine because we don't want users of the `orchard` crate to encounter either
+        // case.
         .and_then(NonZeroPallasBase::from_base)
 }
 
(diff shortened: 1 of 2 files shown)


src/spec.rs
```diff
@@ -190,7 +190,7 @@ pub(crate) fn commit_ivk(
         // - sinsemilla::CommitDomain::short_commit calls extract_p_bottom, which replaces
         //   the identity (which has no affine coordinates) with 0. but Sinsemilla is
         //   defined using incomplete addition, and thus will never produce the identity.
-        .map(NonZeroPallasBase::guaranteed)
+        .and_then(NonZeroPallasBase::from_base)
```
str4d, line 193: The comment also needs updating. Something like:
```rust
    // sinsemilla::CommitDomain::short_commit returns a value in range [0..q_P] ∪ {⊥}:
    // - sinsemilla::HashDomain::hash_to_point uses incomplete addition and returns a
    //   point in P* ∪ {⊥}.
    // - sinsemilla::CommitDomain::commit applies a final complete addition step and
    //   returns a point in P ∪ {⊥}.
    // - 0 is not a valid x-coordinate for any Pallas point.
    // - sinsemilla::CommitDomain::short_commit calls extract_p_bottom, which replaces
    //   the identity (which has no affine coordinates) with 0.
    //
    // Commit^ivk.Output is specified as [1..q_P] ∪ {⊥}, so we explicitly check for 0 and
    // map it to None. Note that we are collapsing this case (which is rejected by the
    // circuit) with ⊥ (which the circuit explicitly allows for efficiency); this is fine
    // because we don't want users of the `orchard` crate to encounter either case.
```

str4d (changes requested):

str4d (approved):
utACK 5356804bfe3add134477f053ebdc25c6930ea1a6

ebfull (approved):

daira (approved):
utACK, this is consistent with the spec change.

PR #299, opened 2022-03-16
Support externally computed signatures

### LINKED ISSUE #298: Support external signatures
In order to support hardware wallets, there should be an API for signing a bundle by appending externally computed signatures.

Signature index can be inferred via `alpha` or `SpendValidatingKey`. I suggest the second option since `alpha`s should stay internal.

jarys:
closes #298

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -36,6 +36,11 @@ pub enum Error {
     /// An overflow error occurred while attempting to construct the value
     /// for a bundle.
     ValueSum(value::OverflowError),
+    /// External signature is not valid.
+    InvalidExternalSignature,
+    /// A signature is valid for more than one input. This should never happen if `alpha`
+    /// is sampled correctly, and indicates a critical failure in randomness generation.
+    DuplicateSignature,
 }
 
 impl From<halo2_proofs::plonk::Error> for Error {
@@ -566,6 +571,47 @@ impl<P: fmt::Debug, V> Bundle<InProgress<P, PartiallyAuthorized>, V> {
             },
             |_, partial| partial,
         )
+    }
+    /// Appends externally computed [`Signature`]s.
+    ///
+    /// Each signature will be applied to the one input for which it is valid. An error
+    /// will be returned if the signature is not valid for any inputs, or if it is valid
+    /// for more than one input.
+    ///
+    /// [`Signature`]: redpallas::Signature
+    pub fn append_signatures(
+        self,
+        signatures: &[redpallas::Signature<SpendAuth>],
+    ) -> Result<Self, Error> {
+        signatures
+            .into_iter()
+            .try_fold(self, Self::append_signature)
+    }
+
+    fn append_signature(self, signature: &redpallas::Signature<SpendAuth>) -> Result<Self, Error> {
+        let mut signature_valid_for = 0usize;
+        let bundle = self.map_authorization(
+            &mut signature_valid_for,
+            |valid_for, partial, maybe| match maybe {
+                MaybeSigned::SigningMetadata(parts) => {
+                    let rk = parts.ak.randomize(&parts.alpha);
+                    if rk.verify(&partial.sigs.sighash[..], signature).is_ok() {
+                        *valid_for += 1;
+                        MaybeSigned::Signature(signature.clone())
+                    } else {
+                        // Signature isn't for this input.
+                        MaybeSigned::SigningMetadata(parts)
+                    }
+                }
+                s => s,
+            },
+            |_, partial| partial,
+        );
+        match signature_valid_for {
+            0 => Err(Error::InvalidExternalSignature),
+            1 => Ok(bundle),
+            _ => Err(Error::DuplicateSignature),
+        }
     }
 }
 


src/builder.rs
```diff
@@ -556,6 +565,31 @@ impl<P: fmt::Debug, V> Bundle<InProgress<P, PartiallyAuthorized>, V> {
...
+                    let rk = parts.ak.randomize(&parts.alpha);
+                    rk.verify(&partial.sigs.sighash[..], signature)
+                        .map_err(Error::from)?;
+                    Ok(MaybeSigned::Signature(signature.clone()))
+                }
+                s => Ok(s),
+            },
+            |_, partial| Ok(partial),
+        )
+    }
```
str4d, line 590: The `alpha` argument is unnecessary, because we can solely determine whether the signature applies to any output by trial-verifying in each output. If all signatures in the transaction are applied externally, this require `2n` trial signature verifications for `n` inputs (since we necessarily skip inputs that already have signatures). Given that we need `n` signature verifications anyway (since we don't want to append an invalid signature), I think this is an acceptable amount of work to simplify the API.

The `rng` argument is also unnecessary (as we aren't generating signatures); we can replace that with a context integer to determine error conditions.

```suggestion
    /// Appends an externally computed [`Signature`].
    ///
    /// The signature will be applied to the one input for which it is valid. An error
    /// will be returned if the signature is not valid for any inputs, or if it is valid
    /// for more than one input.
    pub fn append_signature(
        self,
        signature: &redpallas::Signature<SpendAuth>,
    ) -> Result<Self, Error> {
        let mut signature_valid_for = 0usize;
        self.try_authorize(
            &mut signature_valid_for,
            |valid_for, partial, maybe| match maybe {
                MaybeSigned::SigningMetadata(parts) => {
                    let rk = parts.ak.randomize(&parts.alpha);
                    if rk.verify(&partial.sigs.sighash[..], signature).is_ok() {
                        *valid_for += 1;
                        Ok(MaybeSigned::Signature(signature.clone()))
                    } else {
                        // Signature isn't for this input.
                        Ok(s)
                    }
                }
                s => Ok(s),
            },
            |_, partial| Ok(partial),
        ).expect("Closures cannot fail");
        match signature_valid_for {
            0 => Err(Error::InvalidExternalSignature),
            1 => Ok(()),
            _ => Err(Error::DuplicateSignature),
        }
    }
```

src/builder.rs
```diff
@@ -36,6 +37,8 @@ pub enum Error {
     /// An overflow error occurred while attempting to construct the value
     /// for a bundle.
     ValueSum(value::OverflowError),
+    /// External signature is not valid.
+    InvalidExternalSignature(reddsa::Error),
```
str4d, line 41:
```suggestion
    /// External signature is not valid.
    InvalidExternalSignature,
    /// A signature is valid for more than one input. This should never happen if `alpha`
    /// is sampled correctly, and indicates a critical failure in randomness generation.
    DuplicateSignature,
```

src/builder.rs
```diff
@@ -50,6 +53,12 @@ impl From<value::OverflowError> for Error {
     }
 }
 
+impl From<reddsa::Error> for Error {
+    fn from(e: reddsa::Error) -> Self {
+        Error::InvalidExternalSignature(e)
+    }
+}
+
```
str4d, line 61: No longer necessary:
```suggestion
```

str4d (changes requested):
API direction LGTM. We will separately need to figure out how best to expose the information that signers need.

src/builder.rs
jarys: You are right. That's a nice simplification.

jarys:
I realized it will be more practical to append all signatures at once.

src/builder.rs
```diff
@@ -556,6 +561,50 @@ impl<P: fmt::Debug, V> Bundle<InProgress<P, PartiallyAuthorized>, V> {
     }
 }
 
+impl<P: fmt::Debug, V> Bundle<InProgress<P, PartiallyAuthorized>, V> {
```
str4d, line 564: These are two adjacent `impl` blocks with the same bounds, so they can be merged:
```suggestion

```

str4d (changes requested):

jarys:
suggested changes applied

should I squash commits before merging ?

str4d:
Yeah, there's no need to preserve the fixup commits in git history.

jarys:
force-pushing to squash fixup commits

str4d (approved):
utACK

src/builder.rs
```diff
@@ -554,6 +559,47 @@ impl<P: fmt::Debug, V> Bundle<InProgress<P, PartiallyAuthorized>, V> {
...
+        signatures: &[redpallas::Signature<SpendAuth>],
+    ) -> Result<Self, Error> {
+        signatures
+            .into_iter()
+            .try_fold(self, Self::append_signature)
+    }
+
+    fn append_signature(self, signature: &redpallas::Signature<SpendAuth>) -> Result<Self, Error> {
+        let mut signature_valid_for = 0usize;
+        let bundle = self.authorize(
```
str4d, line 580: Ah, this PR has a non-code merge conflict with #317. Rebase this on `main` and change this to:
```suggestion
        let bundle = self.map_authorization(
```

str4d (changes requested):

src/builder.rs
jarys: done

str4d (approved):
re-utACK

src/builder.rs
```diff
@@ -567,6 +572,47 @@ impl<P: fmt::Debug, V> Bundle<InProgress<P, PartiallyAuthorized>, V> {
             |_, partial| partial,
         )
     }
+    /// Appends externally computed [`Signature`]s.
+    ///
+    /// Each signature will be applied to the one input for which it is valid. An error
+    /// will be returned if the signature is not valid for any inputs, or if it is valid
+    /// for more than one input.
```
daira, line 579: This is O(n<sup>2</sup>) in the number of signatures, but I guess that's okay, and I see why it was done this way.

daira (commented):
Post-hoc utACK

src/builder.rs
str4d: It's only O(n^2) for untrusted inputs, in the case where every single signature is invalid. If you've generated the signatures yourself, it's O(n).

PR #300, opened 2022-03-16
Add serialized proof test case

str4d:
Co-authored-by: Sean Bowe <ewillbefull@gmail.com>


src/circuit.rs
```diff
@@ -952,65 +952,68 @@ mod tests {
...
+        let nf_old = spent_note.nullifier(&fvk);
+        let ak: SpendValidatingKey = fvk.into();
+        let alpha = pallas::Scalar::random(&mut rng);
+        let rk = ak.randomize(&alpha);
+
+        let (_, _, output_note) = Note::dummy(&mut rng, Some(nf_old));
+        let cmx = output_note.commitment().into();
+
+        let value = spent_note.value() - output_note.value();
+        let rcv = ValueCommitTrapdoor::random(&mut rng);
```
nuttycom, line 970: The change from `ValueCommitTrapdoor::zero` to `ValueCommitTrapdoor::random` here was only visible when comparing with whitespace changes disregarded; it seems fine to me but I want to check that it is intended.

src/circuit.rs
```diff
@@ -952,65 +952,68 @@ mod tests {
...
+                alpha: Some(alpha),
+                ak: Some(ak),
+                nk: Some(nk),
+                rivk: Some(rivk),
+                g_d_new_star: Some((*output_note.recipient().g_d()).to_bytes()),
+                pk_d_new_star: Some(output_note.recipient().pk_d().to_bytes()),
+                v_new: Some(output_note.value()),
+                psi_new: Some(output_note.rseed().psi(&output_note.rho())),
+                rcm_new: Some(output_note.rseed().rcm(&output_note.rho())),
+                rcv: Some(rcv),
```
nuttycom, line 996: Same here.

nuttycom (approved):
utACK. The round trip encoding of the example instance looks correct to me, and regeneration using the RNG should result in a different note commitment trapdoor value being selected.

src/circuit.rs
str4d: Yes it is intended; this snuck in because the test was implemented back when we had little crate structure, and that part was never updated. I noticed it while inspecting the serialized proof, because there was a run of all-zeroes due to having two dummy (zero-valued) notes and a zero `rcv` giving a `cv_net` of the identity.

daira (commented):
Post-hoc ACK

PR #305, opened 2022-03-30
Add explicit scoping for viewing keys and addresses

### LINKED ISSUE #303: Check IVK derivations during `FullViewingKey::from_bytes`
We need to ensure that neither ⊥ nor 0 is encountered (which should be negligible, but for consistency we want to reject these at parse time along with more likely-to-occur errors, so that deriving an IVK from an FVK can be purely infallible).

### LINKED ISSUE #304: Check that the internal IVK can be derived from a spending key
Section 4.2.3 of the protocol spec says to check that the regular IVK can be derived from a generated spending key. But we also need to check that the internal IVK is also valid.

therealyingtong:
Closes #303. Closes #304.  

Co-authored-by: Jack Grigg <jack@electriccoin.co>
Co-authored-by: Daira Hopwood <daira@jacaranda.org>

### DIFF
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -906,7 +906,7 @@ mod tests {
 
         let sender_address = spent_note.recipient();
         let nk = *fvk.nk();
-        let rivk = *fvk.rivk();
+        let rivk = fvk.rivk(fvk.scope_for_address(&spent_note.recipient()).unwrap());
         let nf_old = spent_note.nullifier(&fvk);
         let ak: SpendValidatingKey = fvk.into();
         let alpha = pallas::Scalar::random(&mut rng);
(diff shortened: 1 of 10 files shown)


src/circuit.rs
```diff
@@ -906,7 +906,7 @@ mod tests {
 
         let sender_address = spent_note.recipient();
         let nk = *fvk.nk();
-        let rivk = *fvk.rivk();
+        let rivk = fvk.rivk(fvk.scope_for_address(&spent_note.recipient()).unwrap());
```
nuttycom, line 909:
```suggestion
        let rivk = fvk.rivk(fvk.scope_for_address(&sender_address).unwrap());
```
nonblocking

nuttycom (approved):
utACK

PR #309, opened 2022-04-06
Add support for `no-std` builds

### LINKED ISSUE #211: Add `no-std` support
# Background

I'm currently working on implementation of Orchard shielded transactions for Trezor T hardrware wallet (HWW). In our model, keys are generated in the device and spending key never leaves the device. Transaction details must be verified by user on the device screen. Next

- note commitments,
- notes encryption,
- transaction signature digest and
- transaction signatures

must be computed inside of the device to check the integrity. Finally, tx is authorized by a proof computed externally by a host.

# Motivation

The goal of this issue is to start cooperation on making certain components of the Orchard crate `no_std` compatible. This effort will result in code, which can be reused by other HWWs, such as Ledger. Ideally, we would like to use the orchard crate as dependency (not as a fork) to simplify future maintenance. 

# Current state
### Modules
| module | `no_std` compatible |
|-       |-                    |
| address | ✔️ yes |
| constants | ✔️ yes - excluding halo2 constants |
| keys   | ✔️ yes  - excluding `FullViewingKey::{read, write}` |
| note | ✔️ yes |
| note::commitment | ✔️ yes |
| note::nullifier | ✔️ yes |
| note_encryption | ✔️ yes |
| spec | ✔️ yes |
| value | ✔️ yes |
| zip32 | ✔️ yes |

modules that are not necessary for HWW implementation: `builder`, `circuit`, `tree`, `bundle`.

### Dependencies
| dependency | `no_std` compatible |
|-       |-      |
| aes | :heavy_check_mark:  yes |
| bitvec | :heavy_check_mark:  yes |
| blake2b_simd | :heavy_check_mark:  yes |
| ff | :heavy_check_mark:  yes |
| fpe| :heavy_check_mark:  yes |
| group | :heavy_check_mark:  yes |
| halo2_gadgets | :x: no |
| halo2_gadgets::poseidon | :x: almost, `Spec::constants` should enable to return `&'static[[F; T]]` |
| halo2_gadgets::sinsemilla | :heavy_check_mark: yes |
| pasta_curves | :heavy_check_mark:  yes |
| rand | :heavy_check_mark:  yes |
| reddsa | :heavy_check_mark: yes |
| subtle | :heavy_check_mark:  yes |
| zcash_note_encryption | :heavy_check_mark:  yes |

Other dependencies can be disabled on no-std builds.

# Notes

#### hash_to_curve
It would be nice to avoid the `Box` design used for `hash_to_curve` from the `pasta_curves` crate. Why not just

```rust
fn hash_to_curve(dst: &str, message: &[u8]) -> Self;
```
or
```rust
struct HashToCurveDomain {
  dst: &str,
}

impl HashToCurveDomain {
  fn new(dst: &str) -> Self {...} 
  fn hash(self, message: &[u8]) -> Point {...}
}
```
?

#### halo2::arithmetics
In order to don't load some std-dependent components from `halo2` crate, it is necessary to load `arithmetic` directly from `pasta_curves`,

#### Sinsemilla tables
Sinsemilla uses 64kB table, which is a noticeable amount of memory for HWW. It should be possible to use Sinsemilla without this precomputed table, for example through a `sinsemilla-compact` feature. This decreases efficiency dramatically, but still has throughput around 100bit/s on Trezor, which is sufficient for our purpose. [proof-of-concept](https://github.com/jarys/sinsemilla_trezor)

#### alloc

We could start with `no_std` + `alloc` version to overcome all `Box` related design issues. We are can get `alloc` in work on Trezor (using micropython allocator), but unfortunately [`no_std` + `alloc` is  unstable](https://github.com/rust-lang/rust/issues/66741) so far.

Would you like to cooperate on this and merge it?

jarys:
Closes #211.

Curve arithmetic is now imported directly from `pasta_curves` instead of `halo2_proofs`.

`no-std` support for `halo2_gadgets::{sinsemilla, poseidon}` can be added later.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,29 +23,29 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs", "--html-in-header", "katex-header.html"]
 
 [dependencies]
-aes = "0.7"
-bitvec = "1"
-blake2b_simd = "1"
-ff = "0.12"
-fpe = "0.5"
-group = "0.12"
+aes = { version = "0.7", default-features = false }
+bitvec = { version = "1", default-features = false }
+blake2b_simd = { version = "1", default-features = false }
+ff = { version = "0.12", default-features = false, features = ["bits"] }
+fpe = { version = "0.5", default-features = false, features = ["alloc"] }
+group = { version = "0.12", default-features = false }
 halo2_gadgets = "0.2"
-halo2_proofs = "0.2"
-hex = "0.4"
-lazy_static = "1"
-memuse = { version = "0.2", features = ["nonempty"] }
-pasta_curves = "0.4"
+halo2_proofs = { version = "0.2", optional = true }
+hex = { version = "0.4", default-features = false }
+lazy_static = { version = "1", optional = true }
+memuse = { version = "0.2", optional = true, features = ["nonempty"] }
+pasta_curves = { version = "0.4", default-features = false, features = ["alloc", "bits"] }
 proptest = { version = "1.0.0", optional = true }
-rand = "0.8"
-reddsa = "0.3"
-nonempty = "0.7"
-serde = { version = "1.0", features = ["derive"] }
-subtle = "2.3"
-zcash_note_encryption = "0.1"
-incrementalmerkletree = "0.3"
+rand = { version = "0.8", default-features = false }
+reddsa = { version = "0.3", default-features = false }
+nonempty = { version = "0.7", optional = true }
+serde = { version = "1.0", optional = true, features = ["derive"] }
+subtle = { version = "2.3", default-features = false }
+zcash_note_encryption = { version = "0.1", default-features = false }
+incrementalmerkletree = { version = "0.3", optional = true }
 
 # Logging
-tracing = "0.1"
+tracing = { version = "0.1", optional = true }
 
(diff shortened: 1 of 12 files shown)


str4d:
Thanks for this. It's hard to tell from this PR precisely what functionality you need, so what I'm probably going to do is open some smaller PRs that address smaller components of this migration, so we can filter this PR down to its essentials and make sure it addresses the HW needs satisfactorily.

str4d:
Aha, I see you edited the main issue. I'll go look at that.

jarys:
rebase needed after #316

jarys:
pr rebased

jarys:
@str4d , any comments on this peace of work ?

jarys:
Almost all items in `fixed_bases.rs` are `std` conditioned. It think it would worth to `std` conditioning the whole module and moving the remaining constants somewhere else. These constants are
```
VALUE_COMMITMENT_PERSONALIZATION
VALUE_COMMITMENT_V_BYTES
VALUE_COMMITMENT_R_BYTES
NOTE_COMMITMENT_PERSONALIZATION
COMMIT_IVK_PERSONALIZATION
```

jarys:
https://github.com/zcash/orchard/issues/211#issuecomment-1218134130

PR #315, opened 2022-04-28
Cleanups for MSRV 1.56.1

### LINKED ISSUE #253: Remove `array::IntoIter::new` usage once MSRV is 1.53+
The `halo2` gate creation function, for flexibility, allows you to return any type that implements `IntoIterator<Item = C>` where `C: Into<Constraint<F>>`. Prior to Rust 1.53, arrays themselves didn't implement `IntoIterator`, only references to arrays did, so you couldn't return the array on its own (either it didn't satisfy the bound, or the reference to it didn't live long enough), and needed to use the `array::IntoIter::new` helper.

Starting from Rust 1.53, the helper is unnecessary, and we can return arrays directly. Once we bump our MSRV past that, we should remove `array::IntoIter::new` usage, and we can also probably switch several places away from using `std::iter::empty().chain(...)` semantics as well (which we used in many places because the array syntax was messy).

### LINKED ISSUE #275: Raise MSRV to 1.56.1
Rationale:
- 2021 edition support:
  - v2 resolver.
  - `rust-version` in `Cargo.toml`
- `halo2_proofs` will be raising its MSRV to 1.56.1.

str4d:
Closes #253.
Closes #275.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ authors = [
     "Kris Nuttycombe <kris@electriccoin.co>",
 ]
 edition = "2018"
+rust-version = "1.56.1"
 description = "[BETA] The Orchard shielded transaction protocol"
 license-file = "LICENSE-BOSL"
 repository = "https://github.com/zcash/orchard"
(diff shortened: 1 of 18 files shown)


nuttycom (approved):
utACK

PR #316, opened 2022-04-28
Preparations for `no-std` support

str4d:
Part of #211.

### DIFF
--- a/src/bundle.rs
+++ b/src/bundle.rs
@@ -10,139 +10,31 @@ use nonempty::NonEmpty;
 use zcash_note_encryption::{try_note_decryption, try_output_recovery_with_ovk};
 
 use crate::{
+    action::Action,
     address::Address,
     bundle::commitments::{hash_bundle_auth_data, hash_bundle_txid_data},
     circuit::{Instance, Proof, VerifyingKey},
     keys::{IncomingViewingKey, OutgoingViewingKey},
-    note::{ExtractedNoteCommitment, Note, Nullifier, TransmittedNoteCiphertext},
+    note::Note,
     note_encryption::OrchardDomain,
     primitives::redpallas::{self, Binding, SpendAuth},
     tree::Anchor,
     value::{ValueCommitTrapdoor, ValueCommitment, ValueSum},
 };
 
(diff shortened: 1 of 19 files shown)


src/constants/sinsemilla.rs
```diff
@@ -161,27 +161,18 @@ mod tests {
     #[test]
     fn i2lebsp_k_round_trip() {
         {
-            let bitstring = (0..K).map(|_| rand::random()).collect::<Vec<_>>();
-            assert_eq!(
-                i2lebsp_k(lebs2ip_k(&bitstring) as usize).to_vec(),
-                bitstring
-            );
+            let bitstring = [0; K].map(|_| rand::random());
```
nuttycom, line 164: You can map arrays now??? :exploding_head:

src/bundle.rs
```diff
@@ -10,142 +10,34 @@ use nonempty::NonEmpty;
...
-    cmx: ExtractedNoteCommitment,
-    /// The transmitted note ciphertext.
-    encrypted_note: TransmittedNoteCiphertext,
-    /// A commitment to the net value created or consumed by this action.
-    cv_net: ValueCommitment,
-    /// The authorization for this action.
-    authorization: A,
-}
-
 impl<T> Action<T> {
```
nuttycom, line 25: Why is this still allowed to be here without violating the Orphan rule?

nuttycom (approved):
utACK

src/bundle.rs
str4d: The orphan rule applies to crates, not modules. We're still inside the `orchard` crate; it doesn't matter where in it we impl these. See also how we implemented `zcash_address`.

src/constants/sinsemilla.rs
str4d: Yep, since 1.55.

str4d:
[Force-pushed](https://github.com/zcash/orchard/compare/34ad90bc4636e523c9b10f76b80eaabeef6a00cb..5c5c9994393df1b982957b756f72f31f0abea4df) to fix a broken intra-doc link.

Issue #83, opened 2021-05-22
Internal security/correctness review of the Orchard Action circuit implementation and docs


daira:
The PRs implementing the circuit have been / are being reviewed individually, but this PR-based review process leaves potential gaps, because code changed in a PR often has interdependencies with code that hasn't changed.

We need to have another pass over the whole implementation, checking consistency with the Orchard book, the spec, and the intended design in our heads.

daira:
* [ ] Fixed-base full width scalar multiplication
* [ ] Fixed-base signed 64-bit scalar multiplication
* [ ] Variable-base scalar multiplication
* [ ] Complete point addition
* [ ] Sinsemilla hash
* [ ] Sinsemilla NoteCommit
* [ ] Sinsemilla Commit<sup>ivk</sup>
* [ ] Pedersen value commitments
* [ ] Poseidon hash/PRF
* [ ] Nullifier derivation
* [ ] Merkle path validation
* [ ] enableSpends/enableOutputs checks
* [ ] High-level statement

str4d:
Stealing @daira's list above for my own review (#318):

* ECC:
  * [ ] Fixed-base full width scalar multiplication
  * [ ] Fixed-base signed 64-bit scalar multiplication
  * [ ] Variable-base scalar multiplication
  * [ ] Complete point addition
* [ ] Range checks
* Sinsemilla:
  * [ ] Sinsemilla hash
  * [ ] Sinsemilla NoteCommit
  * [ ] Sinsemilla Commit<sup>ivk</sup>
* [ ] Pedersen value commitments
* [ ] Poseidon hash/PRF
* [x] Nullifier derivation
* [ ] Merkle path validation
* [x] enableSpends/enableOutputs checks
* [ ] Inputization
* [x] High-level statement

str4d:
I have reviewed everything I wanted to. (I implemented Poseidon, so I did not review it; others have.)

PR #318, opened 2022-04-29
Changes from circuit review

str4d:
Part of #83.

### DIFF
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -108,8 +108,8 @@ pub struct Circuit {
     pub(crate) ak: Option<SpendValidatingKey>,
     pub(crate) nk: Option<NullifierDerivingKey>,
     pub(crate) rivk: Option<CommitIvkRandomness>,
-    pub(crate) g_d_new_star: Option<[u8; 32]>,
-    pub(crate) pk_d_new_star: Option<[u8; 32]>,
+    pub(crate) g_d_new: Option<NonIdentityPallasPoint>,
+    pub(crate) pk_d_new: Option<DiversifiedTransmissionKey>,
     pub(crate) v_new: Option<NoteValue>,
     pub(crate) psi_new: Option<pallas::Base>,
     pub(crate) rcm_new: Option<NoteCommitTrapdoor>,
@@ -619,11 +619,9 @@ impl plonk::Circuit<pallas::Base> for Circuit {
         {
             let new_note_commit_config = config.new_note_commit_config.clone();
 
-            // Witness g_d_new_star
+            // Witness g_d_new
             let g_d_new = {
-                let g_d_new = self
-                    .g_d_new_star
-                    .map(|bytes| pallas::Affine::from_bytes(&bytes).unwrap());
+                let g_d_new = self.g_d_new.map(|g_d_new| g_d_new.to_affine());
                 NonIdentityPoint::new(
                     ecc_chip.clone(),
                     layouter.namespace(|| "witness g_d_new_star"),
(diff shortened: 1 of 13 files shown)


src/value.rs
```diff
@@ -147,6 +148,24 @@ impl ValueSum {
     pub(crate) fn from_raw(value: i64) -> Self {
         ValueSum(value as i128)
     }
+
+    /// Splits this value sum into its magnitude and sign.
+    ///
+    /// Returns:
+    /// - `(magnitude, false)` if `self` is zero or positive.
+    /// - `(magnitude, true)` if `self` is negative.
+    pub(crate) fn magnitude_sign(&self) -> (u64, bool) {
```
nuttycom, line 157: Consider using an enum instead of this bool.

src/circuit/gadget.rs
```diff
@@ -42,3 +54,71 @@ impl super::Config {
...
+pub(in crate::circuit) fn derive_nullifier<
+    PoseidonChip: PoseidonSpongeInstructions<pallas::Base, poseidon::P128Pow5T3, ConstantLength<2>, 3, 2>,
+    AddChip: AddInstruction<pallas::Base>,
+    EccChip: EccInstructions<
+        pallas::Affine,
+        FixedPoints = OrchardFixedBases,
+        Var = AssignedCell<pallas::Base, pallas::Base>,
+    >,
+>(
+    mut layouter: impl Layouter<pallas::Base>,
```
nuttycom, line 81: Needs a Clippy mitigation for too-many-args

str4d:
I'm done with the Orchard side of my review & cleanup. PR should _definitely_ be reviewed by-commit.

src/circuit.rs
```diff
@@ -175,10 +176,13 @@ impl plonk::Circuit<pallas::Base> for Circuit {
                         "Either v_old = 0, or anchor equals public input",
                         v_old.clone() * (anchor - pub_input_anchor),
```
daira, line 176: `anchor` here is elsewhere called `root`, because it is the calculated root.

src/circuit/commit_ivk.rs
```diff
@@ -258,86 +258,64 @@ pub(in crate::circuit) mod gadgets {
...
-                sinsemilla_chip.clone(),
-                layouter.namespace(|| "a"),
-                a,
-                25,
-            )?
-        };
+        let a = MessagePiece::from_subpieces(
+            sinsemilla_chip.clone(),
+            layouter.namespace(|| "a"),
+            [RangeConstrained::subset_of(ak.value(), 0..250)],
```
daira, line 265: I would have called it `RangeConstrained::bitrange_of`.

src/circuit/note_commit.rs
```diff
@@ -540,7 +689,149 @@ impl NoteCommitConfig {
...
+        let two_pow_6 = two_pow_5 * two;
+        let two_pow_8 = two_pow_4.square();
+        let two_pow_9 = two_pow_8 * two;
+        let two_pow_10 = two_pow_9 * two;
+        let two_pow_58 = pallas::Base::from(1 << 58);
+        let two_pow_130 = Expression::Constant(pallas::Base::from_u128(1 << 65).square());
+        let two_pow_140 = Expression::Constant(pallas::Base::from_u128(1 << 70).square());
+        let two_pow_249 = pallas::Base::from_u128(1 << 124).square() * two;
+        let two_pow_250 = two_pow_249 * two;
+        let two_pow_254 = pallas::Base::from_u128(1 << 127).square();
```
nuttycom, line 1465: It seems a bit unfortunate that there's a mix of `pallas::Base` and `Expression` here. Is there overhead to `Expression::Constant`? Why not uniformly use `Expression`?

nuttycom (approved):
utACK. I've verified that these commits are refactor-only;  f7ed3025473346ae201d41a3945cb5ecb371b9ec is complex to review but everything seems to line up as best I can tell.

src/circuit/note_commit.rs
str4d: Because that alters the circuit 🙂

More specifically, `constant * x` results in two different ASTs depending on which you use. With `Expression` you end up with `AstNode::Mul(Expression, Expression)`, whereas with a bare `pallas::Base` you get `AstNode::Scale(Expression, pallas::Base)`.

The places we've used `Expression` here are where we actually use them in addition, and we only have `impl Add<Expression> for AstNode` defined. We could in theory defined `impl Add<Field> for AstNode` and have it wrap the field element in an `Expression::Constant`, which would move this in the opposite direction to what you suggested, but still be more uniform.

ebfull (approved):

daira (commented):
Post-hoc utACK

PR #323, opened 2022-05-05
Label Consensus Rules

defuse:


src/bundle.rs
```diff
@@ -108,6 +108,7 @@ impl Flags {
     ///
     /// [txencoding]: https://zips.z.cash/protocol/protocol.pdf#txnencoding
     pub fn from_byte(value: u8) -> Option<Self> {
+        // TCR:bad-txns-v5-reserved-bits-nonzero
```
str4d, line 111:
```suggestion
        // https://p.z.cash/TCR:bad-txns-v5-reserved-bits-nonzero
```
(see #321 for other examples)

str4d (approved):
utACK. The `TCR:*` anchors haven't been added to the protocol spec yet, but we should just go ahead and merge this, and separately update the protocol spec with everything needed.

Obinna26:
God bless you Sir🙏🙏 @defuse  will you mind if I plead to share some challenge I have been having on some points of the project 🙏💯 Much blessings ❤️❤️

PR #324, opened 2022-05-06
Fix patch revision for `reddsa` crate

str4d:
https://github.com/ZcashFoundation/reddsa/pull/22 was merged with a squash-commit, which caused the revision in the branch we were patching with to no longer exist.


nuttycom (approved):
utACK, verified this matches the HEAD of current reddsa main.

therealyingtong (approved):
I've checked that this commit corresponds to https://github.com/ZcashFoundation/reddsa/commit/0e912de3000fe165daf58ad98d1a22f1a66e7f18.

PR #331, opened 2022-05-19
Add exception text

steven-ecc:
Update COPYING file with BOSL exception text.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -6,14 +6,19 @@ the file ./LICENSE-BOSL for the terms of the Bootstrap Open Source Licence,
 version 1.0.
 
 Only if this Original Work is included as part of the distribution of one of the
-following projects ("the Project"):
+following (each, the "Project"):
 
-- The Zcash projects published by the Electric Coin Company,
-- The Zebra project published by the Zcash Foundation,
+- The Zcash projects published by the Electric Coin Company;
+- The Zebra project published by the Zcash Foundation;
+- A project that is designed to integrate with Zcash and provides additional
+  functionality or utility to the Zcash network and holders of the ZEC coin; or
+- A blockchain that descends from the Zcash blockchain and that is forked
+  within 100 blocks of the current block height of the Zcash blockchain at the
+  time of the code fork;
 
-then License is granted to use this package under the BOSL as modified by the
-following clarification and special exception. This exception applies only to
-the Original Work when linked or combined with the Project and not to the
+then License is granted to use the Original Work under the BOSL as modified by
+the following clarification and special exception. This exception applies only
+to the Original Work when linked or combined with the Project and not to the
 Original Work when linked, combined, or included in or with any other software
 or project or on a standalone basis.
 


ebfull (approved):
This PR just changes legal stuff I have no authority over.

COPYING
```diff
@@ -6,14 +6,19 @@ the file ./LICENSE-BOSL for the terms of the Bootstrap Open Source Licence,
 version 1.0.
 
 Only if this Original Work is included as part of the distribution of one of the
-following projects ("the Project"):
+following (each, the "Project"):
 
-- The Zcash projects published by the Electric Coin Company,
-- The Zebra project published by the Zcash Foundation,
+- The Zcash projects published by the Electric Coin Company;
+- The Zebra project published by the Zcash Foundation;
```
daira, line 12: I think this doesn't completely address the issue with code forks. If I modify zcashd or Zebra locally, in a way that is still compatible with the Zcash network consensus, can I run the result? It's not obvious, because it will depend on the orchard crate and not be published by ECC or ZF.

(By the way we do this all the time during development, for debugging and such! Being ECC developers is not enough to say that our local forks are published by ECC. So we could be technically violating our own license.)

Similarly, suppose I want to experiment with extensions to Zcash on a testnet (which we clearly intend to be possible).

It may be possible to argue that such uses are covered by the case on line 13, but it isn't clear: if that were true, then why would we need the cases in lines 11 or 12 at all? I think it would be preferable to generalise the line 13 case in such a way that it clearly covers and obviates the line 11 and 12 cases.

daira: @steven-ecc wrote:
> 11 and 12 are covered by the exception on 13 IMO but for avoidance of doubt we didn't remove the two exceptions that preexisted.

I think this is very likely to cause confusion and enable FUD. It's non-obvious to me that "designed to integrate with Zcash" includes full node implementations.

COPYING
```diff
@@ -6,14 +6,19 @@ the file ./LICENSE-BOSL for the terms of the Bootstrap Open Source Licence,
...
+- A blockchain that descends from the Zcash blockchain and that is forked
+  within 100 blocks of the current block height of the Zcash blockchain at the
+  time of the code fork;
 
-then License is granted to use this package under the BOSL as modified by the
-following clarification and special exception. This exception applies only to
-the Original Work when linked or combined with the Project and not to the
+then License is granted to use the Original Work under the BOSL as modified by
+the following clarification and special exception. This exception applies only
+to the Original Work when linked or combined with the Project and not to the
```
daira, line 21: "Combined" isn't defined. Does it include derived works generally, as long as they meet one of the exceptions?

daira (changes requested):
Responses to review comments needed.

COPYING
steven-ecc: Disclaimer: I am not an attorney but there's use and then there's distribution. Forks to test with and deployment on testnet don't fall under distribution AFAIK and it's certainly not the intent of the exception that these be restricted.

Regarding the exceptions for zcashd and zebrad that seem to be covered in some way under the first general exception, I'd rather have duplicity that appear to be _removing_ a previously granted exception.

COPYING
steven-ecc: That is my understanding that combined means the combination or Original and Derived.

COPYING
daira: I don't think that considering code that makes testnet consensus changes as by definition not "distributed" would stand up to a legal challenge. That code certainly has utility, and certainly might need to be conveyed outside of the organisation that wrote it.

In #334 I explicitly added a testnet exception. I'm not wedded to the particular wording there, which may need refinement.

PR #334, opened 2022-05-19
Clarify and widen license exceptions

### LINKED ISSUE #332: #331 was merged without addressing review comments

daira:
The intent is to make clear that it is possible to extend the functionality of Zcash implementations (not necessarily only published by the ECC or ZF), and to experiment on testnets, while still relying on the BOSL exception. The leeway given to chain forks is increased to 3456 blocks (roughly 3 days) to make reasonable allowance for release processes.

closes #332

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -11,9 +11,9 @@ In this document,
   the Electric Coin Company and the Zcash Foundation according to section 6.2 (b)
   of the Zcash Trademark Agreement.
 - A "Zcash Chain Fork" means a blockchain that descends from the Zcash
-  blockchain and that activates its first change to the consensus rules relative to Zcash at a block 
-  height that is or was within 3456 blocks of the current height of the Zcash
-  chain at the time of activation.
+  blockchain and that activates its first change to the consensus rules 
+  relative to Zcash at a block height that is or was within 3456 blocks 
+  of the current height of the Zcash chain at the time of activation.
 - A "Zcash Technology Testnet" means a blockchain designed for the purpose of
   testing technology potentially of use to the Zcash blockchain and not for
   the purpose of implementing any token of economic value, where this purpose
--- a/COPYING
+++ b/COPYING
@@ -11,7 +11,7 @@ In this document,
   the Electric Coin Company and the Zcash Foundation according to section 6.2 (b)
   of the Zcash Trademark Agreement.
 - A "Zcash Chain Fork" means a blockchain that descends from the Zcash
-  blockchain and that activates a change to the consensus rules at a block 
+  blockchain and that activates its first change to the consensus rules relative to Zcash at a block 
   height that is or was within 3456 blocks of the current height of the Zcash
   chain at the time of activation.
 - A "Zcash Technology Testnet" means a blockchain designed for the purpose of
--- a/COPYING
+++ b/COPYING
@@ -5,22 +5,40 @@ Source License, version 1.0, or at your option, any later version ("BOSL"). See
 the file ./LICENSE-BOSL for the terms of the Bootstrap Open Source Licence,
 version 1.0.
 
+In this document,
+
+- "Zcash" means the blockchain defined by the most recent agreement between
+  the Electric Coin Company and the Zcash Foundation according to section 6.2 (b)
+  of the Zcash Trademark Agreement.
+- A "Zcash Chain Fork" means a blockchain that descends from the Zcash
+  blockchain and that is or was forked within 3456 blocks of the current block
+  height of the Zcash blockchain at the time of the fork;
+- A "Zcash Technology Testnet" means a blockchain designed for the purpose of
+  testing technology potentially of use to the Zcash blockchain and not for
+  the purpose of implementing any token of economic value, where this purpose
+  and lack of such economic value are clearly communicated to its users and to
+  the general public.
+
 Only if this Original Work is included as part of the distribution of one of the
 following (each, the "Project"):
 
-- The Zcash projects published by the Electric Coin Company;
-- The Zebra project published by the Zcash Foundation;
-- A project that is designed to integrate with Zcash and provides additional
-  functionality or utility to the Zcash network and holders of the ZEC coin; or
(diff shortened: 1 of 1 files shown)


nuttycom (commented):
nonbinding: I approve of these clarifications.

COPYING
```diff
@@ -5,23 +5,41 @@ Source License, version 1.0, or at your option, any later version ("BOSL"). See
...
 version 1.0.
 
+In this document,
+
+- "Zcash" means the blockchain defined by the most recent agreement between
+  the Electric Coin Company and the Zcash Foundation according to section 6.2 (b)
+  of the Zcash Trademark Agreement.
+- A "Zcash Chain Fork" means a blockchain that descends from the Zcash
+  blockchain and that is or was forked within 3456 blocks of the current block
+  height of the Zcash blockchain at the time of the fork;
```
nuttycom, line 15: I think that "at the time of the fork" could be further clarified to reference the divergence of the consensus rules; after all, if the consensus rules don't diverge, it's not actually a fork. This would eliminate any confusion that might be related to the time at which the *source code* was forked.

daira: Can you suggest a wording change? I have COVID and I'm flagging a bit (don't worry I'll be fine).

steven-ecc:
I'll collect all of these suggestions and comments and also see if we get other feedback from partners or the community and then get them reviewed. 

Thanks for all the feedback!

COPYING
nuttycom, line 15:
```suggestion
blockchain and that activates a change to the consensus rules at a block 
height that is within 3456 blocks of the current tip of the main Zcash chain 
at the time of activation.
```

COPYING
```diff
@@ -5,23 +5,41 @@ Source License, version 1.0, or at your option, any later version ("BOSL"). See
...
-- The Zebra project published by the Zcash Foundation;
-- A project that is designed to integrate with Zcash and provides additional
-  functionality or utility to the Zcash network and holders of the ZEC coin; or
-- A blockchain that descends from the Zcash blockchain and that is forked
-  within 100 blocks of the current block height of the Zcash blockchain at the
-  time of the code fork;
+- A project that implements Zcash;
+- A project that implements a Zcash Chain Fork;
+- A project that implements a Zcash Technology Testnet;
+- A project that is designed to integrate with Zcash (whether or not it can
```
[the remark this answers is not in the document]
daira: My intent was that if you add features for a chain fork then you should make a reasonable effort to provide them for the Zcash chain. Sometimes this might not be possible, if the features rely on particular consensus or peer-to-peer protocol differences. I'm not sure that the intent *can* be fully captured in legal wording; at some point you have to assume good faith.

COPYING
daira:
```suggestion
  blockchain and that activates a change to the consensus rules at a block 
  height that is or was within 3456 blocks of the current height of the Zcash
  chain at the time of activation.
```

COPYING
```diff
@@ -5,23 +5,42 @@ Source License, version 1.0, or at your option, any later version ("BOSL"). See
 the file ./LICENSE-BOSL for the terms of the Bootstrap Open Source Licence,
 version 1.0.
 
+In this document,
+
+- "Zcash" means the blockchain defined by the most recent agreement between
+  the Electric Coin Company and the Zcash Foundation according to section 6.2 (b)
+  of the Zcash Trademark Agreement.
+- A "Zcash Chain Fork" means a blockchain that descends from the Zcash
+  blockchain and that activates a change to the consensus rules at a block
```
daira, line 14:
```suggestion
  blockchain and that activates its first change to the consensus rules at a block 
```

COPYING
```diff
@@ -5,23 +5,42 @@ Source License, version 1.0, or at your option, any later version ("BOSL"). See
...
 
+In this document,
+
+- "Zcash" means the blockchain defined by the most recent agreement between
+  the Electric Coin Company and the Zcash Foundation according to section 6.2 (b)
+  of the Zcash Trademark Agreement.
+- A "Zcash Chain Fork" means a blockchain that descends from the Zcash
+  blockchain and that activates its first change to the consensus rules relative to Zcash at a block 
+  height that is or was within 3456 blocks of the current height of the Zcash
+  chain at the time of activation.
```
nuttycom, line 16:
```suggestion
- A "Zcash Chain Fork" means a blockchain that descends from the Zcash
  blockchain and that activates its first change to the consensus rules 
  relative to Zcash at a block height that is or was within 3456 blocks 
  of the current height of the Zcash chain at the time of activation.
```

nuttycom: Is `3456` a placeholder value, or is there a justification for this specific value?

COPYING
```diff
@@ -5,23 +5,42 @@ Source License, version 1.0, or at your option, any later version ("BOSL"). See
...
 
+In this document,
+
+- "Zcash" means the blockchain defined by the most recent agreement between
+  the Electric Coin Company and the Zcash Foundation according to section 6.2 (b)
+  of the Zcash Trademark Agreement.
+- A "Zcash Chain Fork" means a blockchain that descends from the Zcash
+  blockchain and that activates its first change to the consensus rules 
+  relative to Zcash at a block height that is or was within 3456 blocks 
+  of the current height of the Zcash chain at the time of activation.
```
daira, line 16: The idea here is that the fork's implementation might have to hard-code information (e.g. a checkpoint or a snapshot of some of the chain state) that depends on the Zcash chain, and then have nodes running that code when the fork happens. This change allows 3 days between those two points, which is enough time to prepare the code, do a release, and have a reasonable number of nodes running.

Not all forks will need this; it depends how much they are changing. But it could plausibly be needed.

COPYING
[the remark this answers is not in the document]
daira: Note that this is only a clarification; it does not change the legal meaning. It's less important than the changes on lines 27 and 28.

daira:
Closing in favour of #405.

PR #338, opened 2022-06-21
Add `orchard::bundle::BatchValidator`

str4d:


src/bundle/batch.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    /// Batch-validates the accumulated bundles.
+    ///
+    /// Returns `true` if every proof and signature in every bundle added to the batch
+    /// validator is valid, or `false` if one or more are invalid. No attempt is made to
+    /// figure out which of the accumulated bundles might be invalid; if that information
+    /// is desired, construct separate [`BatchValidator`]s for sub-batches of the bundles.
+    pub fn validate<R: RngCore + CryptoRng>(self, vk: &VerifyingKey, rng: R) -> bool {
+        if self.signatures.is_empty() {
+            // An empty batch is always valid, but is not free to run; skip it.
```
daira, line 71:
```suggestion
            // An empty batch is always valid, but is not free to run; skip it.
            // Note that a transaction has at least a binding signature, so if
            // there are no signatures, there are also no proofs.
```

therealyingtong (approved):
utACK

ebfull (approved):

src/bundle/batch.rs
```diff
@@ -69,6 +69,8 @@ impl BatchValidator {
     pub fn validate<R: RngCore + CryptoRng>(self, vk: &VerifyingKey, rng: R) -> bool {
         if self.signatures.is_empty() {
             // An empty batch is always valid, but is not free to run; skip it.
+            // Note that a transaction has at least a binding signature, so if
+            // there are no signatures, there are also no proofs.
```
daira, line 73: I guess this should have more precisely said "A transaction has at least an Orchard binding signature of it has any Orchard components, so if there are no Orchard signatures, there are also no Orchard proofs."

PR #340, opened 2022-06-22
Add OrchardDomain::for_nullifier and CompactAction::from_parts

### LINKED ISSUE #339: It's impossible to create a CompactAction from data
https://github.com/zcash/orchard/blob/8aba4c59fdc02417be6af061116e1debef32dfef/src/note_encryption.rs#L270-L275

This struct feels like an intended use-case is to be able to receive a compact action from a lightclient, and deserialize it to this type. Right now there's no way to actually create one from its fields, so instead of using the `CompactAction` type I need to basically copy-paste these lines onto my own data type I'm capable of creating an instance of: https://github.com/zcash/orchard/blob/8aba4c59fdc02417be6af061116e1debef32dfef/src/note_encryption.rs#L296-L308

AloeareV:
Fixes #339.

### DIFF
--- a/src/note_encryption.rs
+++ b/src/note_encryption.rs
@@ -327,6 +327,11 @@ impl CompactAction {
             enc_ciphertext,
         }
     }
+
+    ///Returns the nullifier of the note being spent.
+    pub fn nullifier(&self) -> Nullifier {
+        self.nullifier
+    }
 }
 
 #[cfg(test)]
--- a/src/note_encryption.rs
+++ b/src/note_encryption.rs
@@ -94,7 +94,7 @@ impl OrchardDomain {
     }
 
     /// Constructs a domain from a nullifier.
-    pub fn for_nullifier<T>(nullifier: Nullifier) -> Self {
+    pub fn for_nullifier(nullifier: Nullifier) -> Self {
         OrchardDomain { rho: nullifier }
     }
 }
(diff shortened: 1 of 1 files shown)


steven-ecc:
Thanks for the PR @AloeareV! We appreciate it! 

As you may know, Orchard is licensed under [BOSL](https://github.com/zcash/orchard/blob/main/LICENSE-BOSL) and we need a contributor agreement signed before we're able to merge changes into this repo similar to [how Apache does contributions](https://www.apache.org/licenses/contributor-agreements.html). If you can email me at `steven _at_ electriccoin.co` from your preferred email address, I'll have a Docusign sent over for you to review.

daira (approved):
utACK modulo CLA.

AloeareV:
Need to either add a Nullifier getter to `CompactAction`, or add a `OrchardDomain::for_compact_action`, so that I can actually create the domain from the CompactAction. `for_compact_action` would arguably remove the need for `for_nullifier`. Being able to create the domain from just the Nullifier is more extensible, though, and I think that's more useful here.

src/note_encryption.rs
```diff
@@ -92,6 +92,11 @@ impl OrchardDomain {
             rho: *act.nullifier(),
         }
     }
+
+    /// Constructs a domain from a nullifier.
+    pub fn for_nullifier(nullifier: Nullifier) -> Self {
```
str4d, line 97: For future reference, these additions should be added to the changelog. In this case, I can do this in the release PR since we'll be cutting it imminently and won't forget.

str4d (approved):

str4d:
CLA has been received.

PR #343, opened 2022-07-04
Expose `Proof::add_to_batch` in public API

str4d:
This supports downstream users that want more control over how proof batches are processed, instead of just batch validating Orchard bundles with `orchard::bundle::BatchValidator`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `orchard::Proof::add_to_batch`
 
 ## [0.2.0] - 2022-06-24
 ### Added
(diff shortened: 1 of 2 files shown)


daira (approved):
utACK

str4d:
@dconnolly is going to test that this works for `zebra`, and if it is sufficient then I will merge this and make an `0.2.1` release PR.

dconnolly:
This seems to be working!

One more request, we cannot get the params or raw plonk verifying key out of `VerifyingKey`, can we add `fn params(self) -> Params` and `fn vk(self) -> plonk::VerifyingKey` to `orchard::circuit::VerifyingKey`? Right now I'm testing by regenerating them as in `build()`

nuttycom (approved):
utACK

str4d:
> One more request, we cannot get the params or raw plonk verifying key out of `VerifyingKey`, can we add `fn params(self) -> Params` and `fn vk(self) -> plonk::VerifyingKey` to `orchard::circuit::VerifyingKey`?

@dconnolly I don't want to expose these APIs in the `orchard` crate. I'd rather add an API to `VerifyingKey` that can accept a `halo2_proofs::plonk::BatchVerifier`. But I suspect that may not suit your `Service` usage in `zebrad`? In which case, you should likely just continue to do what you are currently doing, namely duplicating the three or four lines from inside `VerifyingKey::build()`.

PR #344, opened 2022-07-05
Publicize necessary functionality for creating diversifiers and notes from data

AloeareV:
One use-case is to enable wallets to store notes and diversifiers on disk.

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -552,7 +552,7 @@ impl DiversifierKey {
 pub struct Diversifier([u8; 11]);
 
 impl Diversifier {
-    ///Read a diversifier from a byte array.
+    /// Reads a diversifier from a byte array.
     pub fn from_bytes(d: [u8; 11]) -> Self {
         Diversifier(d)
     }
@@ -1050,7 +1050,8 @@ mod tests {
                 NoteValue::from_raw(tv.note_v),
                 rho,
                 RandomSeed::from_bytes(tv.note_rseed, &rho).unwrap(),
-            );
+            )
+            .unwrap();
 
             let cmx: ExtractedNoteCommitment = note.commitment().into();
             assert_eq!(cmx.to_bytes(), tv.note_cmx);
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -552,7 +552,8 @@ impl DiversifierKey {
 pub struct Diversifier([u8; 11]);
 
 impl Diversifier {
-    pub(crate) fn from_bytes(d: [u8; 11]) -> Self {
+    ///Read a diversifier from a byte array.
+    pub fn from_bytes(d: [u8; 11]) -> Self {
         Diversifier(d)
     }
 
--- a/src/note.rs
+++ b/src/note.rs
@@ -21,7 +21,7 @@ pub use self::nullifier::Nullifier;
 
 /// The ZIP 212 seed randomness for a note.
 #[derive(Copy, Clone, Debug)]
-pub(crate) struct RandomSeed([u8; 32]);
+pub struct RandomSeed([u8; 32]);
 
 impl RandomSeed {
     pub(crate) fn random(rng: &mut impl RngCore, rho: &Nullifier) -> Self {
@@ -35,13 +35,16 @@ impl RandomSeed {
         }
     }
 
-    pub(crate) fn from_bytes(rseed: [u8; 32], rho: &Nullifier) -> CtOption<Self> {
+    ///Read a note's random seed from bytes, given the note's nullifier.
+    ///Returns None if the nullifier is not for the same note as the seed.
+    pub fn from_bytes(rseed: [u8; 32], rho: &Nullifier) -> CtOption<Self> {
         let rseed = RandomSeed(rseed);
         let esk = rseed.esk_inner(rho);
         CtOption::new(rseed, esk.is_some())
     }
 
-    pub(crate) fn as_bytes(&self) -> &[u8; 32] {
+    /// Returns the byte array corresponding to this seed.
+    pub fn as_bytes(&self) -> &[u8; 32] {
         &self.0
     }
 
@@ -108,7 +111,8 @@ impl PartialEq for Note {
 impl Eq for Note {}
 
 impl Note {
-    pub(crate) fn from_parts(
+    ///Create a Note from its component parts.
+    pub fn from_parts(
         recipient: Address,
         value: NoteValue,
         rho: Nullifier,
--- a/src/note.rs
+++ b/src/note.rs
@@ -35,8 +35,9 @@ impl RandomSeed {
         }
     }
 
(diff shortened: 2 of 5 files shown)


therealyingtong (approved):
utACK

zancas:
Pinging on this PR...   looks like it's almost ready?

zancas:
Hiya @therealyingtong it looks like merging this PR will also unblock this development effort:

https://github.com/zcash/orchard/pull/346

Is it the case that it simply lacks a second review?

AloeareV:
Made some changes, in line with daira's feedback on #346

src/note.rs
```diff
@@ -35,13 +35,16 @@ impl RandomSeed {
         }
     }
 
-    pub(crate) fn from_bytes(rseed: [u8; 32], rho: &Nullifier) -> CtOption<Self> {
+    ///Read a note's random seed from bytes, given the note's nullifier.
+    ///Returns None if the nullifier is not for the same note as the seed.
```
daira, line 39:
```suggestion
    /// Read a note's random seed from bytes, given the note's nullifier.
    /// Returns `None` if the nullifier is not for the same note as the seed.
```

src/note.rs
```diff
@@ -108,18 +111,26 @@ impl PartialEq for Note {
 impl Eq for Note {}
 
 impl Note {
-    pub(crate) fn from_parts(
+    ///Create a Note from its component parts. Returns None if a valid
+    ///[`NoteCommitment`] cannot be derived from the note.
```
daira, line 115:
```suggestion
    /// Create a Note from its component parts. Returns `None` if a valid
    /// [`NoteCommitment`] cannot be derived from the note.
```

src/keys.rs
```diff
@@ -552,7 +552,8 @@ impl DiversifierKey {
 pub struct Diversifier([u8; 11]);
 
 impl Diversifier {
-    pub(crate) fn from_bytes(d: [u8; 11]) -> Self {
+    ///Read a diversifier from a byte array.
```
daira, line 555:
```suggestion
    /// Read a diversifier from a byte array.
```

daira (approved):
utACK modulo @str4d's suggestions for the changelog.

str4d (changes requested):
We discussed these API changes, and decided that this is consistent with how we want the public API to be expanded.

src/note.rs
```diff
@@ -108,18 +111,26 @@ impl PartialEq for Note {
...
+    pub fn from_parts(
```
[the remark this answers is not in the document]
AloeareV: Thanks! Comments and changelog updated!

str4d:
CI is failing because the new commit that was pushed triggered a test of this merged into current `main`, which is currently broken. Once #358 is merged, we will re-run CI.

nuttycom:
> CI is failing because the new commit that was pushed triggered a test of this merged into current `main`, which is currently broken. Once #358 is merged, we will re-run CI.

Since what is tested is the branch state, not the merged state, this PR will need to be rebased again, right?

str4d:
> This PR will need to be rebased again, right?

I think it just needs to be pushed again; the PR contents doesn't have to change. I have squashed the fixup commits into the original 3 locally to clean up the history in a way that doesn't alter this PR's contents, but because Zingo Labs have disabled the ability for us to push to their branches (usually this is a setting that is default-enabled when creating a PR, but maybe there's some Zingo Labs org setting that is inverting the permission), I can't update this PR.

AloeareV:
> but because Zingo Labs have disabled the ability for us to push to their branches (usually this is a setting that is default-enabled when creating a PR, but maybe there's some Zingo Labs org setting that is inverting the permission), I can't update this PR.

https://github.com/community/community/discussions/5634

It seems this feature doesn't exist for organization-owned repos, only user-owned ones. I'll do a similar squash.

str4d:
> It seems this feature doesn't exist for organization-owned repos

Aha, that explains it! I had assumed it was because org-owned repos had extra config options to lock down permissions, not that they had fewer config options.

src/note.rs
```diff
@@ -108,18 +112,33 @@ impl PartialEq for Note {
...
+    /// [Section 4.19] of the Zcash Protocol Specification. Users of this constructor
+    /// should only call it with note components that have been fully validated by
+    /// decrypting a received note according to [Section 4.19].
+    ///
+    /// [Section 4.19]: https://zips.z.cash/protocol/protocol.pdf#saplingandorchardinband
+    pub fn from_parts(
         recipient: Address,
         value: NoteValue,
         rho: Nullifier,
         rseed: RandomSeed,
```
str4d, line 133: It occurs to me now that a potentially simpler API for downstream users would be for the public `from_parts` to take `rseed` as a `[u8; 32]` and not expose the `RandomSeed` type externally at all (currently the only thing you can do with it publicly is wrap it around a byte array, and I don't see this changing in future). The `RandomSeed::from_bytes` check would then be done internally to `Note::from_parts` as:
```rust
impl Note {
    pub fn from_parts(
        recipient: Address,
        value: NoteValue,
        rho: Nullifier,
        rseed: [u8; 32],
    ) -> CtOption<Self> {
        RandomSeed::from_bytes(rseed, &rho)
            .and_then(|rseed| Self::from_parts_internal(recipient, value, rho, rseed))
    }

    fn from_parts_internal(
        recipient: Address,
        value: NoteValue,
        rho: Nullifier,
        rseed: RandomSeed,
    ) -> CtOption<Self> {
        ...
    }

    pub fn rseed(&self) -> &[u8; 32] {
        self.rseed.as_bytes()
    }
}
```

On the other hand, having the `RandomSeed` type exist may make it easier for crate users to keep it typed in the space between where they interact with their database and where they interact with `Note`. Then again, there is nothing stopping them only using `RandomSeed` immediately before `Note::from_parts`, which makes the type safety completely pointless in their usage:
```rust
let note = RandomSeed::from_bytes(rseed, &rho)
    .and_then(|rseed| Note::from_parts(recipient, value, rho, rseed))
````

str4d (approved):
utACK 0800d23fe729ebcf1ea129688bfe67f9d2d7f29d

src/note.rs
zancas: Can we call the local variable name on the line that starts with `.and_then` something other than `rseed`?

nuttycom: I think that this is okay as it is currently; it makes the role of `rho` clear in constructing the `RandomSeed` value.

nuttycom (approved):
utACK https://github.com/zcash/orchard/commit/0800d23fe729ebcf1ea129688bfe67f9d2d7f29d

src/note.rs
AloeareV: The `RandomSeed` being constructed beforehand makes it more obvious to the consumer whether the note construction went wrong due to a mismatched rho/rseed, or something else. I don't know how to weigh that against the ergonomics of it.

PR #346, opened 2022-07-29
Allow creating notes from serialized parts

adityapk00:
Allow creating an Orchard `Note` from serialized parts. 

When lightclients parse Orchard actions, they need to serialize the notes into the wallet to be able to spend them at a later time. Right now, the Note's constructors are all `pub(crate)`. This PR allows constructing a `Note` from it's components. 

It is possible I'm misunderstanding how Notes are managed, so if there's another way a lightwallet is supposed to store/load/serialize Orchard notes, please let me know.

### DIFF
--- a/src/note.rs
+++ b/src/note.rs
@@ -21,7 +21,7 @@ pub use self::nullifier::Nullifier;
 
 /// The ZIP 212 seed randomness for a note.
 #[derive(Copy, Clone, Debug)]
-pub(crate) struct RandomSeed([u8; 32]);
+pub struct RandomSeed([u8; 32]);
 
 impl RandomSeed {
     pub(crate) fn random(rng: &mut impl RngCore, rho: &Nullifier) -> Self {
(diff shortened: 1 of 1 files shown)


AloeareV:
I'm not aware of another way, hence #344.

src/note.rs
```diff
@@ -108,7 +110,8 @@ impl PartialEq for Note {
 impl Eq for Note {}
 
 impl Note {
-    pub(crate) fn from_parts(
+    /// Construct note from its component parts
+    pub fn from_parts(
```
daira, line 114: This doesn't check the invariant `note.commitment_inner().is_some()`. It can be made public provided it checks that invariant (i.e. it would return `CtOption<Self>`).

There are two existing uses of `Note::from_parts` that need to be changed in that case:

https://github.com/zcash/orchard/blob/3faab98e9e82618a0f2d887054e9e28b0f7947dd/src/note_encryption.rs#L78
I think the correct change here is for the whole `orchard_parse_note_plaintext_without_memo` function to return `None` if `Note::from_parts` does. Any change to this code is consensus-critical and requires three reviews (at least two from ECC engineers).

----
https://github.com/zcash/orchard/blob/3faab98e9e82618a0f2d887054e9e28b0f7947dd/src/keys.rs#L1047-L1052
This is in a test so can just unwrap.

daira (changes requested):
Changes needed as per comment.

str4d:
This is a strict subset of #344; we'll focus our API review there.

PR #348, opened 2022-08-11
Add `Circuit::from_action_context` constructor

jarys:


daira (approved):
utACK

src/circuit.rs
```diff
@@ -120,6 +121,49 @@ pub struct Circuit {
...
+        let scope = fvk.scope_for_address(&input_note.recipient())?;
+        Some(Circuit {
+            path: Value::known(merkle_path.auth_path()),
+            pos: Value::known(merkle_path.position()),
+            g_d_old: Value::known(input_note.recipient().g_d()),
+            pk_d_old: Value::known(*input_note.recipient().pk_d()),
+            v_old: Value::known(input_note.value()),
+            rho_old: Value::known(input_note.rho()),
+            psi_old: Value::known(input_note.rseed().psi(&input_note.rho())),
+            rcm_old: Value::known(input_note.rseed().rcm(&input_note.rho())),
```
daira, line 151:
```suggestion
        let rho = input_note.rho();
        Some(Circuit {
            path: Value::known(merkle_path.auth_path()),
            pos: Value::known(merkle_path.position()),
            g_d_old: Value::known(input_note.recipient().g_d()),
            pk_d_old: Value::known(*input_note.recipient().pk_d()),
            v_old: Value::known(input_note.value()),
            rho_old: Value::known(rho),
            psi_old: Value::known(input_note.rseed().psi(&rho)),
            rcm_old: Value::known(input_note.rseed().rcm(&rho)),
```

src/circuit.rs
```diff
@@ -120,6 +121,49 @@ pub struct Circuit {
...
+    /// - `output_note`: a note created in scope of the action
+    /// - `fvk`: full viewing key owning the `input_note`
+    /// - `alpha`: a scalar used for randomization of the action spend validating key
+    /// - `rcv`: trapdoor for the action value commitment
+    ///
+    /// Returns `None` if the `input_note` is not owned by the `fvk`.
+    pub fn from_action_context(
+        merkle_path: tree::MerklePath,
+        input_note: Note,
+        output_note: Note,
```
str4d, line 137: Passing in the notes like this means that we don't have a guarantee that the output note's `rho` is using the input note's nullifier. Add a check that these are the same (which will require deriving the nullifier for the input note).

src/circuit.rs
```diff
@@ -120,6 +121,49 @@ pub struct Circuit {
...
+            rivk: Value::known(fvk.rivk(scope)),
+            g_d_new: Value::known(output_note.recipient().g_d()),
+            pk_d_new: Value::known(*output_note.recipient().pk_d()),
+            v_new: Value::known(output_note.value()),
+            psi_new: Value::known(output_note.rseed().psi(&output_note.rho())),
+            rcm_new: Value::known(output_note.rseed().rcm(&output_note.rho())),
+            rcv: Value::known(rcv),
+        })
+    }
+}
```
str4d:
```suggestion
    pub fn from_action_context(
        spend: SpendInfo,
        output_note: Note,
        alpha: pallas::Scalar,
        rcv: ValueCommitTrapdoor,
    ) -> Option<Circuit> {
        (spend.note.nullifier(&spend.fvk) == output_note.rho()).then(|| {
            Self::from_action_context_unchecked(
                spend_info,
                output_note,
                alpha,
                rcv,
            )
        })
    }

    pub(crate) fn from_action_context_unchecked(
        spend: SpendInfo,
        output_note: Note,
        alpha: pallas::Scalar,
        rcv: ValueCommitTrapdoor,
    ) -> Circuit {
        let scope = fvk.scope_for_address(&input_note.recipient())?;

        let sender_address = spend.note.recipient();
        let rho_old = spend.note.rho();
        let psi_old = spend.note.rseed().psi(&rho_old);
        let rcm_old = spend.note.rseed().rcm(&rho_old);

        let rho_new = output_note.rho();
        let psi_new = output_note.rseed().psi(&rho_new);
        let rcm_new = output_note.rseed().rcm(&rho_new);

        Circuit {
            path: Value::known(spend.merkle_path.auth_path()),
            pos: Value::known(spend.merkle_path.position()),
            g_d_old: Value::known(sender_address.g_d()),
            pk_d_old: Value::known(*sender_address.pk_d()),
            v_old: Value::known(input_note.value()),
            rho_old: Value::known(rho_old),
            psi_old: Value::known(psi_old),
            rcm_old: Value::known(rcm_old),
            cm_old: Value::known(spend.note.commitment()),
            alpha: Value::known(alpha),
            ak: Value::known(fvk.clone().into()),
            nk: Value::known(*fvk.nk()),
            rivk: Value::known(fvk.rivk(scope)),
            g_d_new: Value::known(output_note.recipient().g_d()),
            pk_d_new: Value::known(*output_note.recipient().pk_d()),
            v_new: Value::known(output_note.value()),
            psi_new: Value::known(psi_new),
            rcm_new: Value::known(rcm_new),
            rcv: Value::known(rcv),
        })
    }
}
```

src/circuit.rs
```diff
@@ -120,6 +121,49 @@ pub struct Circuit {
...
+    /// Returns `None` if the `input_note` is not owned by the `fvk`.
+    pub fn from_action_context(
+        merkle_path: tree::MerklePath,
+        input_note: Note,
+        output_note: Note,
+        fvk: FullViewingKey,
+        alpha: pallas::Scalar,
+        rcv: ValueCommitTrapdoor,
+    ) -> Option<Circuit> {
+        let scope = fvk.scope_for_address(&input_note.recipient())?;
```
str4d, line 147: I think that an API built around the existing builder components may be simpler to reason about.

The `SpendInfo` type is already correctly constructed from the components `input_note, merkle_path, fvk` in `Builder::add_spend`, so we could encapsulate this with a public `SpendInfo::new(..) -> Option<Self>` constructor.

The `RecipientInfo` type is definitely the wrong type to use here, because we already have a constructed note and presumably don't want to run encryption again.

The `ActionInfo` type could potentially be repurposed / refactored to share codepaths with this API. One option is that this API be changed to take a `SpendInfo`, and then the builder and this external API would share most of the circuit witness generation logic.

EDIT: After updating the suggestion to do this, I much prefer it. `ActionInfo::build` would then be altered to use `Circuit::from_action_context_unchecked`.

str4d (changes requested):

src/circuit.rs
jarys: done in 603a5cf798fe82d174c514d16f79ef70cc4a4eb5

src/circuit.rs
jarys: Mind you that `fvk.scope_for_address(&input_note.recipient())?;` makes `from_action_context_unchecked` to failable.

src/circuit.rs
jarys: So the conclusion is:
- publicize `SpendInfo`
- add `SpendInfo::new`
- change `from_action_context` type to take input context packed in `SpendInfo`
- add `Circuit::from_action_context_unchecked`
- use this constructor in `Action::build`

jarys:
How about moving these new functionalities behind some "custom_builder" flag?

src/circuit.rs
daira: It wasn't redundant because `scope` is used in `fvk.rivk(scope)`. You are correct that `from_action_context_unchecked` needs to be fallible, as long as it still calls `scope_for_address`.

I think I fixed it. However I don't really like that the two possible errors are collapsed. @str4d, can you think of a better approach, or should we change it to return a `Result`?

daira:
> How about moving these new functionalities behind some "custom_builder" flag?

I don't think that's necessary.

src/circuit.rs
daira: I fixed the suggestion just to convince myself, but you'd already come up with the same thing.

daira:
Force-pushed to resolve the CHANGELOG conflict and fix a clippy lint.

src/builder.rs
```diff
@@ -58,15 +57,35 @@ impl From<value::OverflowError> for Error {
...
+    /// This constructor is public to enable creation of custom builders.
+    /// If you are not creating a custom builder, use [`Builder::add_spend`] instead.
+    ///
+    /// Creates a `SpendInfo` from note, full viewing key owning the note,
+    /// and merkle path witness of the note.
+    ///
+    /// Returns `None` if the `fvk` does not own the `note`.
+    ///
+    /// [`Builder::add_spend`]: Builder::add_spend
+    pub fn new(fvk: FullViewingKey, note: Note, merkle_path: MerklePath) -> Option<Self> {
```
str4d, line 78: Non-blocking: `Builder::add_spend` should be updated to use this constructor.
```rust
        self.spends.push(
            SpendInfo::new(fvk, note, merkle_path)
                .ok_or("FullViewingKey does not correspond to the given note")?
        );
```

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ and this project adheres to Rust's notion of
 - `orchard::note`:
   - `RandomSeed`
   - `Note::{from_parts, rseed}`
+- `orchard::builder::SpendInfo::new`
```
str4d, line 16: `SpendInfo` itself is what is being newly exposed:
```suggestion
- `orchard::builder::SpendInfo`
```

str4d (approved):
utACK 5fbbded49e3162a31fd3bb0de3c344f3cc4dfa60

PR #349, opened 2022-08-11
Add `ValueCommitTrapdoor::from_bytes` constructor

jarys:


src/value.rs
```diff
@@ -216,6 +216,11 @@ impl ValueCommitTrapdoor {
     pub(crate) fn inner(&self) -> pallas::Scalar {
         self.0
     }
+
+    /// Constructs `ValueCommitTrapdoor` from the byte reprsentation of a scalar
```
daira, line 220:
```suggestion
    /// Constructs `ValueCommitTrapdoor` from the byte representation of a scalar.
```

daira (approved):
utACK

src/value.rs
```diff
@@ -216,6 +216,11 @@ impl ValueCommitTrapdoor {
     pub(crate) fn inner(&self) -> pallas::Scalar {
         self.0
     }
+
+    /// Constructs `ValueCommitTrapdoor` from the byte representation of a scalar.
+    pub fn from_bytes(bytes: [u8; 32]) -> CtOption<Self> {
```
str4d, line 221: I'm not even sure why `ValueCommitTrapdoor` was ever a public type. AFAICT it is not used anywhere in our public API; there was no way to construct one publicly, and nothing consumes it.

This API will enable crate users to construct `ValueCommitTrapdoor` from bytes, after which the (inherently-exposed) `Add` and `Sum` impls will become usable. But what is the purpose of being able to publicly construct a `ValueCommitTrapdoor`?

str4d: Ah, I'm guessing this should be reviewed in combination with #351 which exposes an API that consumes `ValueCommitTrapdoor`. That PR provides the context that this PR lacks: that the caller is deriving a `ValueCommitment` themselves.

This API needs significantly more documentation, because it is a relatively low-level API that has the potential to result in a complete loss of privacy (because it pushes the responsibility for correct generation of the trapdoor onto the caller). The documentation should describe the contexts in which this API should be used, and how to correctly and safely generate the bytes to be used as input.

str4d (changes requested):
API needs documentation per my comment above.

src/value.rs
nuttycom: Why not remove those `Add` and `Sum` instances?

src/value.rs
```diff
@@ -216,6 +216,11 @@ impl ValueCommitTrapdoor {
     pub(crate) fn inner(&self) -> pallas::Scalar {
         self.0
     }
+
+    /// Constructs `ValueCommitTrapdoor` from the byte representation of a scalar.
```
daira, line 220:
```suggestion
    /// Constructs `ValueCommitTrapdoor` from the byte representation of a scalar.
    ///
    /// This is a low-level API, requiring a detailed understanding of the use of
    /// value commitment trapdoors in the Zcash protocol [orchardbalance] to
    /// use correctly. It is intended to be used in combination with
    /// [`ValueCommitment::derive`].
    ///
    /// [orchardbalance] https://zips.z.cash/protocol/protocol.pdf#orchardbalance
```

PR #350, opened 2022-08-11
Add `ValueSum::commit`

jarys:
This PR adds API for creating `ValueCommitment`s.

It enables us serialize less data when passing proof witness from HWW to a Prover.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -174,6 +174,15 @@ impl ValueSum {
             sign,
         )
     }
+
+    /// Commits this value with the corresponding trapdoor `rcv` by $ValueCommit^Orchard$
+    ///
+    /// Defined in [Zcash Protocol Spec § 5.4.8.3: Homomorphic Pedersen commitments (Sapling and Orchard)][concretehomomorphiccommit].
+    ///
+    /// [concretehomomorphiccommit]: https://zips.z.cash/protocol/nu5.pdf#concretehomomorphiccommit
+    pub fn commit(self, rcv: ValueCommitTrapdoor) -> ValueCommitment {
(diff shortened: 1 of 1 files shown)


src/value.rs
```diff
@@ -174,6 +174,15 @@ impl ValueSum {
             sign,
         )
     }
+
+    /// Commits this value with the corresponding trapdoor `rcv` by $ValueCommit^Orchard$
```
daira, line 178:
```suggestion
    /// Commits this value with the corresponding trapdoor `rcv` by ValueCommit^Orchard.
```
(LaTeX would not render correctly here.)

src/value.rs
```diff
@@ -174,6 +174,15 @@ impl ValueSum {
...
+    ///
+    /// Defined in [Zcash Protocol Spec § 5.4.8.3: Homomorphic Pedersen commitments (Sapling and Orchard)][concretehomomorphiccommit].
+    ///
+    /// [concretehomomorphiccommit]: https://zips.z.cash/protocol/nu5.pdf#concretehomomorphiccommit
+    pub fn commit(self, rcv: ValueCommitTrapdoor) -> ValueCommitment {
+        ValueCommitment::derive(self, rcv)
```
daira, line 184: Why not make `ValueCommitment::derive` public instead?

daira (commented):
I'm not sure it's necessary to add a new method on `ValueSum`, as opposed to making the existing `ValueCommitment::derive` method public.

jarys:
Yeah, we can just publicize `ValueCommitment::derive`.

Closing this PR in favor of #351.

PR #351, opened 2022-08-13
Publicize `ValueCommitment::derive` constructor

jarys:
This PR adds API for creating `ValueCommitment`s.

It enables us serialize less data when passing proof witness from HWW to a Prover.


daira (approved):
utACK

daira (approved):
utACK (reviewed with @sellout in conjunction with #349).

PR #352, opened 2022-08-23
add value_balance to builder

AloeareV:


src/builder.rs
```diff
@@ -221,6 +221,7 @@ pub struct Builder {
     recipients: Vec<RecipientInfo>,
     flags: Flags,
     anchor: Anchor,
+    value_balance: i64,
```
daira, line 224: @nuttycom and I: Why cache this and update it in `add_spend`/`add_recipient`, rather than recomputing it in the `value_balance` method?

src/builder.rs
```diff
@@ -302,6 +302,20 @@ impl Builder {
         Ok(())
     }
 
+    /// The net value of the bundle. The value of all spends, minus the value
+    /// of all outputs.
+    pub fn value_balance(&self) -> i64 {
```
str4d, line 307: What is the use case for this method (calculating the value balance from the in-progress builder state)? It would be useful to describe intended use cases in the documentation.

src/builder.rs
```diff
@@ -302,6 +302,23 @@ impl Builder {
...
 
+    /// The net value of the bundle to be built. The value of all spends,
+    /// minus the value of all outputs.
+    ///
+    /// Useful for balancing a transaction, as the value balance of an individual bundle
+    /// can be non-zero, but a transaction may not have a positive total value balance.  
+    pub fn value_balance(&self) -> i64 {
+        self.spends
+            .iter()
+            .map(|spend| spend.note.value().inner() as i64)
```
nuttycom, line 313: I would prefer that this return a `ValueSum` value, which wraps an i128 internally.

nuttycom (changes requested):
i64 is not sufficiently type-safe, and we already have `ValueSum` to represent the values of these kinds of balances.

str4d:
> we already have `ValueSum` to represent the values of these kinds of balances.

`ValueSum` is not for representing value balances; that type is only ever represented abstractly in this crate. But you do raise a good point that `i64` is the wrong type to use in this API.

AloeareV:
Switching to ValueSum necessitated adding a `Sub` impl... If the balance type is only ever represented abstractly, then maybe we add some safety checks and then still return an i64? Having a new type that is only ever used in the return of this one specific function is probably not the right answer either, if you can't use it anywhere without first doing a type conversion.

src/builder.rs
```diff
@@ -302,6 +302,25 @@ impl Builder {
...
+            .iter()
+            .map(|spend| ValueSum::from_raw(spend.note.value().inner() as i64))
+            .sum::<Result<ValueSum, OverflowError>>()?
+            - self
+                .recipients
+                .iter()
+                .map(|recipient| ValueSum::from_raw(recipient.value.inner() as i64))
+                .sum::<Result<ValueSum, OverflowError>>()?)
+        .ok_or(OverflowError)
+    }
```
str4d, line 322: I was in the middle of writing this comment when the PR was updated:

> This sum must not be computed with `i64` values, because a value sum within a bundle is a signed 64-bit integer, which is not the same as an `i64` (which is a signed _63-bit_ integer). The `as i64` casts can cause data loss resulting in financial errors, and overflows are not handled. This should use the existing types like the builder itself does, and its return value should be user-specified like it is for `Bundle`:

My suggestion still applies:
```suggestion
    pub fn value_balance<V: TryFrom<i64>>(&self) -> Result<V, value::OverflowError> {
        let value_balance = self
            .spends
            .iter()
            .map(|spend| spend.note.value() - NoteValue::zero())
            .chain(
                self.recipients
                    .iter()
                    .map(|recipient| NoteValue::zero() - recipient.value),
            )
            .fold(Some(ValueSum::zero()), |acc, note_value| acc? + note_value)
            .ok_or(OverflowError)?;

        i64::try_from(value_balance).and_then(|i| V::try_from(i).map_err(|_| value::OverflowError))
    }
```

src/value.rs
```diff
@@ -188,6 +188,18 @@ impl Add for ValueSum {
...
+
+    #[allow(clippy::suspicious_arithmetic_impl)]
+    fn sub(self, rhs: Self) -> Self::Output {
+        self.0
+            .checked_sub(rhs.0)
+            .filter(|v| VALUE_SUM_RANGE.contains(v))
+            .map(ValueSum)
+    }
+}
+
```
str4d, line 202: Undo this addition.

CHANGELOG.md
```diff
@@ -9,7 +9,9 @@ and this project adheres to Rust's notion of
 
 ## [0.2.0] - 2022-06-24
 ### Added
+- `orchard::value::ValueSum` now implements `core::ops::Sub`
```
str4d, line 12: Undo this addition (and squash the commits to clean up the branch).

str4d (changes requested):

AloeareV:
`zcash_primitives::transaction::components::amount::Amount` is probably the right type in the abstract, but `zcash_primitives` depends on orchard, so there's a circular dependency issue there

CHANGELOG.md
AloeareV: Yes, pushed the PR addressing @nuttycom's comments, and then saw yours.

str4d:
@AloeareV the type of a value balance is intentionally abstract in this crate, to allow different downstream consumers to use their own existing types (`zcash_primitives::transaction::components::amount::Amount` for our crates, Zebra's own type in their case). See [the documentation for the `orchard::value` module](https://docs.rs/orchard/0.2.0/orchard/value/index.html), where we document this extensively.

str4d:
Actually, now that I think about it, this PR should add "(and [`Builder::value_balance`] for in-progress bundles)" to the end of the third bullet-point in that module documentation for `orchard::value`, now that it is adding another way that value balances are produced.

str4d (approved):
utACK 8011e0d57a5d245e0a0e678590dd17fd7d992664, but I am now the author of almost all of this PR's code, so it needs another review (because I do not trust myself to not get my lefts and rights mixed up).

nuttycom (approved):
utACK, modulo it would be good to have a test for `value_balance`

src/builder.rs
```diff
@@ -302,6 +302,26 @@ impl Builder {
...
+            .spends
+            .iter()
+            .map(|spend| spend.note.value() - NoteValue::zero())
+            .chain(
+                self.recipients
+                    .iter()
+                    .map(|recipient| NoteValue::zero() - recipient.value),
+            )
+            .fold(Some(ValueSum::zero()), |acc, note_value| acc? + note_value)
+            .ok_or(OverflowError)?;
```
sellout, line 321: Would it help to minimize the risk of overflow if rather than `chain`, `itertools::interleave` were used?

sellout (commented):
Just a question, not a requested change.

daira (changes requested):
Blocking on my review tomorrow.

src/builder.rs
```diff
@@ -302,6 +302,26 @@ impl Builder {
         Ok(())
     }
 
+    /// The net value of the bundle to be built. The value of all spends,
+    /// minus the value of all outputs.
+    ///
+    /// Useful for balancing a transaction, as the value balance of an individual bundle
+    /// can be non-zero, but a transaction may not have a positive total value balance.
```
daira, line 309:
```suggestion
    /// can be non-zero. Each bundle's value balance is added to the transparent
    /// transaction value pool, which [must not have a negative value]. (If it were
    /// negative, the transaction would output more value than it spends.)
    ///
    /// [must not have a negative value]: https://zips.z.cash/protocol/protocol.pdf#transactions
```

src/builder.rs
daira: No, the sum of all spent values cannot exceed MAX_MONEY, and similarly for all output values. `ValueSum` enforces this.

src/builder.rs
```diff
@@ -302,6 +302,26 @@ impl Builder {
...
+            .iter()
+            .map(|spend| spend.note.value() - NoteValue::zero())
+            .chain(
+                self.recipients
+                    .iter()
+                    .map(|recipient| NoteValue::zero() - recipient.value),
+            )
+            .fold(Some(ValueSum::zero()), |acc, note_value| acc? + note_value)
+            .ok_or(OverflowError)?;
+        i64::try_from(value_balance).and_then(|i| V::try_from(i).map_err(|_| value::OverflowError))
```
daira, line 322: This is correct because `i64` can represent all of $[-$ MAX_MONEY, MAX\_MONEY $]$.

daira (approved):
utACK; also checked `src/value.rs` for overflow potential.

src/builder.rs
daira: I would also be fine with this returning `Result<ValueSum, OverflowError>` (which we already have at line 321).

src/builder.rs
daira: I forgot to commit this before merging. See #359.

PR #357, opened 2022-09-12
Dummy implementation of Domain API change for prepared ivk and epk

daira:
This adapts to API changes in https://github.com/zcash/librustzcash/pull/633 without changing the behaviour for Orchard.

PR #358, opened 2022-09-15
Point patch directive at `zcash/librustzcash`

str4d:
zcash/orchard#357 pointed the patch at the PR branch's repository, which broke as soon as the PR branch was deleted from that repository.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,5 +86,5 @@ debug = true
 debug = true
 
 [patch.crates-io]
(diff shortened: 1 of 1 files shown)


nuttycom (approved):
utACK

nuttycom (commented):
re-utACK https://github.com/zcash/orchard/pull/358/commits/526cbed2358578f56df29d3ed03d5b52248abf34

PR #362, opened 2022-10-15
Batch scanner improvements

str4d:


benches/note_decryption.rs
```diff
@@ -39,7 +40,7 @@ fn bench_note_decryption(c: &mut Criterion) {
             let mut sk = [0; 32];
```
daira, line 40: I know this is not related to the scanner improvement, but it might result in inaccurate benchmarking for the improvement: 1000 is not a large enough sample of (action, ivk) pairs given that 1/256 of instances have a significantly higher cost. It should be at least 10000.

src/bundle.rs
```diff
@@ -290,7 +290,8 @@ impl<T: Authorization, V> Bundle<T, V> {
             .filter_map(|(idx, action)| {
                 let domain = OrchardDomain::for_action(action);
                 keys.iter().find_map(move |ivk| {
-                    try_note_decryption(&domain, ivk, action)
+                    let prepared_ivk = PreparedIncomingViewingKey::new(ivk);
```
daira, line 293: Loop over the keys to prepare them first, so that there is only one preparation per key, not (actions × keys) preparations. (I know this is not an ideal API, but might as well avoid unnecessary inefficiency within that constraint, since it's easy to do.)

src/keys.rs
```diff
@@ -627,7 +628,8 @@ impl KeyAgreementPrivateKey {
 
     /// Returns the payment address for this key corresponding to the given diversifier.
     fn address(&self, d: Diversifier) -> Address {
-        let pk_d = DiversifiedTransmissionKey::derive_inner(self, &d);
+        let prepared_ivk = PreparedIncomingViewingKey::new_inner(&self);
```
daira, line 631: Fix the clippy lint.

src/keys.rs
```diff
@@ -841,6 +865,20 @@ impl EphemeralPublicKey {
     }
 }
 
+/// A Sapling ephemeral public key that has been precomputed for trial decryption.
+#[derive(Clone, Debug)]
+pub struct PreparedEphemeralPublicKey(PreparedNonIdentityBase);
```
daira, line 870: This looks odd: in Sapling non-identity points have no special status. Non-small order points do, and prime-order points pdo.

daira: I see the issue: $\mathsf{GroupHash}^{\mathbb{J}^{(r)\ast}}$ returns a prime-order point but is fallible, and unlike in Orchard, a $\bot$ return is not replaced with another point.

daira: Sorry I must not be quite awake yet. This is the orchard crate; the comment should say Orchard.

src/keys.rs
```diff
@@ -841,6 +865,20 @@ impl EphemeralPublicKey {
     }
 }
 
+/// A Sapling ephemeral public key that has been precomputed for trial decryption.
```
daira, line 868:
```suggestion
/// An Orchard ephemeral public key that has been precomputed for trial decryption.
```

daira, line 868:
```suggestion
/// An Orchard ephemeral public key that has been precomputed for trial decryption.
```

daira (approved):
utACK modulo non-blocking comments.

daira (approved):
utACK with @sellout and @nuttycom.

str4d:
The changelog changes in this PR had a non-merge conflict with #363, because this PR was implemented as if it would be part of 0.3.0 but was in fact merged just afterwards. We need to fix the changelog to move the entries added by this PR into the 0.4.0 section.

PR #365, opened 2022-12-01
Add spends/outputs getter fns to builders for use in change calculation

AloeareV:
The `Input/OutputView` traits are nearly straight-copied from https://github.com/zcash/librustzcash/blob/main/zcash_primitives/src/transaction/components/sapling/fees.rs, I'm assuming that interface is the desired one to this kind of functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,9 @@ and this project adheres to Rust's notion of
   - `impl memuse::DynamicUsage for Nullifier`
 - `orchard::note_encryption`:
   - `impl memuse::DynamicUsage for OrchardDomain`
-- `orchard::builder::SpendInfo::new`
+- `orchard::builder`:
+  - `{SpendInfo::new, InputView, OutputView}`
+  - `Builder::{spends, outputs}`
 - `orchard::circuit::Circuit::from_action_context`
 - impls of `Eq` for:
   - `orchard::zip32::ChildIndex`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to Rust's notion of
 ## [Unreleased]
 ### Added
 - `orchard::builder`:
+  - `{SpendInfo::new, InputView, OutputView}`
+  - `Builder::{spends, outputs}`
   - `SpendError` 
   - `OutputError` 
 
@@ -33,9 +35,6 @@ and this project adheres to Rust's notion of
   - `impl memuse::DynamicUsage for Nullifier`
 - `orchard::note_encryption`:
   - `impl memuse::DynamicUsage for OrchardDomain`
-- `orchard::builder`:
-  - `{SpendInfo::new, InputView, OutputView}`
-  - `Builder::{spends, outputs}`
 - `orchard::circuit::Circuit::from_action_context`
 - impls of `Eq` for:
   - `orchard::zip32::ChildIndex`
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -345,6 +345,16 @@ impl Builder {
         Ok(())
     }
 
+    /// Returns the action spend components that will be produced by the transaction being constructed
+    pub fn spends(&self) -> &Vec<impl InputView<()>> {
+        &self.spends
+    }
+
+    /// Returns the action output components that will be produced by the transaction being constructed
+    pub fn outputs(&self) -> &Vec<impl OutputView> {
+        &self.recipients
+    }
+
     /// The net value of the bundle to be built. The value of all spends,
     /// minus the value of all outputs.
     ///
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -21,7 +21,9 @@ and this project adheres to Rust's notion of
   - `impl memuse::DynamicUsage for Nullifier`
 - `orchard::note_encryption`:
   - `impl memuse::DynamicUsage for OrchardDomain`
-- `orchard::builder::SpendInfo::new`
+- `orchard::builder`:
+  - `{SpendInfo::new, InputView, OutputView}`
+  - `Builder::{spends, outputs}`
```
nuttycom, line 26: This should be moved to the `Unreleased` section.

nuttycom (changes requested):
The changelog needs fixing, otherwise LGTM

CHANGELOG.md
AloeareV: Fixed!

src/builder.rs
```diff
@@ -294,6 +294,16 @@ impl Builder {
         Ok(())
     }
 
+    /// Returns the action spend components that will be produced by the transaction being constructed
+    pub fn spends(&self) -> &Vec<impl InputView<()>> {
+        &self.spends
+    }
+
+    /// Returns the action output components that will be produced by the transaction being constructed
```
nuttycom, line 302:
```suggestion
    /// Returns the action output components that will be produced by the
    /// transaction being constructed
```

src/builder.rs
```diff
@@ -294,6 +294,16 @@ impl Builder {
         Ok(())
     }
 
+    /// Returns the action spend components that will be produced by the transaction being constructed
```
nuttycom, line 297:
```suggestion
    /// Returns the action spend components that will be produced by the 
    /// transaction being constructed
```

src/builder.rs
```diff
@@ -657,6 +667,39 @@ impl<V> Bundle<InProgress<Proof, PartiallyAuthorized>, V> {
...
+pub trait InputView<NoteRef> {
+    /// An identifier for the input being spent.
+    fn note_id(&self) -> &NoteRef;
+    /// The value of the input being spent.
+    fn value<V: From<u64>>(&self) -> V;
+}
+
+impl InputView<()> for SpendInfo {
+    fn note_id(&self) -> &() {
+        // The builder does not make use of note identifiers, so we can just return the unit value.
```
nuttycom, line 681: Could this then be an uninhabited type like `std::convert::Infallible`, such that this method will panic if called?

daira: I'm dubious about this API because `InputView<anything>` does not seem to be the right type. We discussed this in PR gardening and @str4d or @nuttycom were going to open an issue.

AloeareV: I followed how the API was used [here](https://github.com/zcash/librustzcash/blob/903922c59ac85ce56aa0e13a8a62e7128f5c555e/zcash_primitives/src/transaction/components/sapling/builder.rs#L72-L82) without giving it a significant amount of thought. I don't really understand the use-case for note_id, although I see that it's in use in `zcash_client_backend`. I'm concerned about having an explicit panic, if it comes up where someone wants to be generic for `T: InputView`, a method that might noop is a lot easier to work with than a method that might panic.

nuttycom: The panic is actually safe, because if someone is trying to do something in their generic code with the uninhabited return type of that method, the compiler will prevent you from using that operation. This is the great value of parametricity; we can statically exclude that the method will be called in situations where it would be unsafe to do so.

AloeareV: I'm not sure what you mean. From a quick [playground test](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=406eab482094d7ce70881e4172e7c15e), when you actually call the generic code with the type determined, the compiler knows it's a panic, and will give unused code warnings about any code after the panic, but the code compiles fine and panics at runtime. If you comment out line 5, the only compiler warning is an unused variable. There are no compiler warnings associated with the `foo` fn itself. 

It's not a method zingolabs is likely to use, and the compiler does seem to do a reasonable job of warning consumers about panics.. The only case I can think of where this could realistically be missed (without ignoring compiler warnings) would be a library that exposes a generic consumer of `InputView` without having any calls to that function that use the specific concrete type in question. This scenario might be too contrived to bother worrying about. 

If you agree this is too contrived to be worth worrying about, I'm happy to swap out the unit type for an uninhabited type.

AloeareV: @nuttycom If this is waiting on me, let me know. I'm not convinced on uninhabited-vs-unit type, but the downside feels a bit contrived so I'm happy to be overruled here.

nuttycom: My rationale here is that if one is using `InputView` as a constraint, this could potentially cause unexpected behavior if, for example, the note ID is used as a key in a map. Making it panic on an invalid call is safer from that perspective.

nuttycom: I just realized though that this is a clone of what was done for Sapling, so this is fine.

nuttycom (approved):
utACK

src/builder.rs
```diff
@@ -345,6 +345,18 @@ impl Builder {
         Ok(())
     }
 
+    /// Returns the action spend components that will be produced by the
```
nuttycom, line 348:
```suggestion
    /// Returns the action spend components that will be produced by the
```

PR #372, opened 2023-01-31
OrchardZSA reference implementation

PaulLaux:
This PR implements [draft ZIP 226 (Transfer and Burn of Zcash Shielded Assets)](https://qed-it.github.io/zips/zip-0226) and [draft ZIP 227 (Issuance of Zcash Shielded Assets)](https://qed-it.github.io/zips/zip-0227). 

- Burning mechanism: the ability to publicly burn ZSA notes. This is done by adding a `burn` field inside the Orchard `bundle` and the ability to utilize it. In addition the changes to `bvk` and `bsk` to support this change. See https://github.com/QED-it/orchard/pull/35 for a full list of changes.
```rust
/// A bundle of actions to be applied to the ledger.
pub struct Bundle<A: Authorization, V, P: OrchardPrimitives> {
    /// The list of actions that make up this bundle.
    actions: NonEmpty<Action<A::SpendAuth, P>>,
    /// Orchard-specific transaction-level flags for this bundle.
    flags: Flags,
    /// The net value moved out of the Orchard shielded pool.
    ///
    /// This is the sum of Orchard spends minus the sum of Orchard outputs.
    value_balance: V,
    /// Assets intended for burning
    burn: Vec<(AssetBase, NoteValue)>,
    /// The root of the Orchard commitment tree that this bundle commits to.
    anchor: Anchor,
    /// Block height after which this Bundle's Actions are invalid by consensus.
    ///
    /// For the OrchardZSA protocol, `expiry_height` is set to 0, indicating no expiry.
    /// This field is reserved for future use.
    expiry_height: u32,
    /// The authorization for this bundle.
    authorization: A,
}
}
```
- Note Encryption: We extended the `Domain` trait in the `zcash_note_encryption` crate to support various lengths for `COMPACT_NOTE_SIZE` by converting the relevant constants into trait-specific types. Later, implemented `OrchardDomainV3` to support the new note structure.
- Updated the Python implementation to support the changes. See https://github.com/zcash/zcash-test-vectors/pull/108 for details.

Also included:
- Test vectors for `AssetID` derivatio
- Split note mechanism.
- Circuit changes to support `AssetID`, split notes, and all other new functionality. 
- Backward compatibility for `Orchard` Bundles

The proposed changes should be considered **a draft**, not a final version.


src/bundle.rs
```diff
@@ -154,31 +200,38 @@ impl Flags {
...
     anchor: Anchor,
+    /// Block height after which this Bundle's Actions are invalid by consensus.
+    ///
+    /// For the OrchardZSA protocol, `expiry_height` is set to 0, indicating no expiry.
+    /// This field is reserved for future use.
+    expiry_height: u32,
     /// The authorization for this bundle.
-    authorization: T,
+    authorization: A,
 }
```
nuttycom: @str4d and I had an extensive discussion about this type in a pairing. Our conclusion is that parameterizing the `Bundle` type via the `D` parameter causes the downstream code that consumes Orchard bundles to be exposed to details about note encryption that are too low-level for the consumer. Also, the fact that we are parameterizing with `D` but are not fully concretizing the contents of the bundle relative to that parameter enables the representation of mixed states (e.g. `D = OrchardVanilla` but with `burn` nonempty) that we do not want to exist.

Instead, there should be separate `Bundle` and `BundleZsa` types; this then means that:
*  the `OrchardDomainCommon` type may not need to exist, and we can use separate implementations of `zcash_note_encryption::Domain`.
* the new fields, `expiry_height` and `burn` will only exist in the `BundleZsa` type.

The `Action` type then doesn't need to be parameterized by the domain type; it can just be parameterized by the type of the `enc_ciphertext`, and each `Bundle` type can concretely set this to a ciphertext type that directly matches what is used in the domain. This will facilitate the memo changes from ZIP 231 (for which the concrete length will be different than at present, and for which `zcash_note_encryption::Domain::Memo` will be set to the type of the memo key.)

At the top level, what we will do is route scanning of V6 transaction outputs into one batch decryptor, and scanning of V5 outputs into a separate batch decryption.

PaulLaux: Actually, having the parameter `D` does not expose much detail to the downstream consumer (unless explicitly requested).
The consumer have either
```rust
let bundle: Bundle<_, _, Orcahrd> = ...
```
Or
```rust
let bundle: Bundle<_, _, OrcahrdZSA> = ...
```
and can proceed with the `Bundle<_>` operations. But this almost never happens since most of the handling is unified. (We have it mostly in tests.)

The major consideration for this type of parametrization is clean code reuse: in most cases `Bundle<_,_,Orchard>` and `Bundle<_,_,OrchardZSA>` are treated the same and we can have nice abstraction in the form of 
```rust
    let bundle: Bundle<_, i64, D> = builder.build(rng);
    bundle
        .create_proof(&pk, rng)
        .unwrap()
        .apply_signatures(rng, [0; 32], &[])
        .unwrap()
```
In critical places like
```rust
pub fn build<V: TryFrom<i64>, D: OrchardFlavor>(
        self,
        rng: impl RngCore,
    ) -> Result<Bundle<InProgress<Unproven, Unauthorized>, V, D>, BundleMetadata), BuildError> {
```
or
```rust
 pub fn prepare<R: RngCore + CryptoRng>(
        self,
        mut rng: R,
        sighash: [u8; 32],
) -> Bundle<InProgress<P, PartiallyAuthorized>, V, D> {
```
or
```rust
pub fn apply_signatures<R: RngCore + CryptoRng>(
        self,
        mut rng: R,
        sighash: [u8; 32],
        signing_keys: &[SpendAuthorizingKey],
) -> Result<Bundle<Authorized, V, D>, BuildError> {
```
among many others.

Later it is propogated `Action`:
```rust
fn build<D: OrchardDomainCommon>(
        self,
        mut rng: impl RngCore,
) -> (Action<SigningMetadata, D>, Witnesses) {
```
Where also most of the logic is shared between an Orchard action and an OrchardZSA action.

All while the difference between the Bundles or the Actions is encapsulated inside `OrchardFlavor` and called transparently by the consumers.

All while fully supporting backward compatibility.

Removing the parameter `D` and creating two different types will obliterate all these nice properties, and the code will be harder to maintain and harder to reason about, **as most of the logic for both is the same.**

That being said, we can change the handling for the note encryption part to align with ZIP 231 once it is done and we can have it (or we can do it on top).

Also, the burn handling can be changed to have a dedicated empty type for `Orchard` and non-empty type for `OrchardZSA` as we did for Zebra's `ShieldedData`:
```rust
pub struct ShieldedData<FL: Flavor> {
   ...
    /// Assets intended for burning
    /// Denoted as `vAssetBurn` in the spec (ZIP 230).
    pub burn: FL::BurnType,
}
``` 
Forcing the burn to be empty for Orchard.

Other solutions for Burn and note encryption are also possible.

However, we strongly recommend keeping the `D` parameter for `Bundle` for the reasons mentioned above.

Cargo.toml
```diff
@@ -30,26 +30,28 @@ ff = { version = "0.13", default-features = false }
...
-proptest = { version = "1.0.0", optional = true }
+proptest = { version = ">=1.0.0, <1.7.0", optional = true }
 rand = { version = "0.8", default-features = false }
+rand_core = { version = "0.6", default-features = false }
 reddsa = { version = "0.5", default-features = false }
 nonempty = { version = "0.11", default-features = false }
 poseidon = { package = "halo2_poseidon", version = "0.1" }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
-sinsemilla = "0.1"
+sinsemilla = { git = "https://github.com/zcash/sinsemilla", rev = "aabb707e862bc3d7b803c77d14e5a771bcee3e8c" }
```
nuttycom: These dependency bumps should be done via `[patch.crates-io]` directives.

ConstanceBeguier: Done

PaulLaux:
continued in https://github.com/zcash/orchard/pull/471

Issue #371, opened 2023-01-25
Migrate to `ff 0.13`


str4d:
Blocked on a compatible `reddsa` release, which itself is blocked on a compatible `bls12_381` release.

PR #377, opened 2023-03-02
Migrate to `ff 0.13`

str4d:
MSRV is now 1.60.

Closes zcash/orchard#371.


nuttycom (approved):
utACK. I reviewed the ff 0.13 and pasta_curves 0.5.x changes and everything here makes sense to me.

daira (commented):
Post-hoc ACK

PR #379, opened 2023-03-10
Migrate to zcash_note_encryption 0.3.0

nuttycom:
This can be merged prior to a `zcash_note_encryption 0.3.0` release, though it will need to be updated to upgrade to such a release prior to `orchard 0.4.0`.


ebfull (approved):
utACK

PR #394, opened 2023-05-19
Remove redundant checks during note encryption

str4d:
The consistency check between `esk` and `ephemeral_key` is checked inside `zcash_note_encryption::try_output_recovery_with_ock`, and the requirement to check it inside the `Domain` implementation is being lifted in zcash/librustzcash#848.

Removing the check here improves performance, both because we avoid an extra scalar multiplication from `esk.derive_public()`, and because we avoid an unnecessary `spec::diversify_hash()` call which is expensive for Orchard.

### DIFF
--- a/src/note_encryption.rs
+++ b/src/note_encryption.rs
@@ -17,7 +17,6 @@ use crate::{
         OutgoingViewingKey, PreparedEphemeralPublicKey, PreparedIncomingViewingKey, SharedSecret,
     },
     note::{ExtractedNoteCommitment, Nullifier, RandomSeed},
-    spec::diversify_hash,
     value::{NoteValue, ValueCommitment},
     Address, Note,
 };
@@ -52,10 +51,10 @@ pub(crate) fn prf_ock_orchard(
 fn orchard_parse_note_plaintext_without_memo<F>(
     domain: &OrchardDomain,
     plaintext: &[u8],
-    get_validated_pk_d: F,
+    get_pk_d: F,
 ) -> Option<(Note, Address)>
 where
-    F: FnOnce(&Diversifier) -> Option<DiversifiedTransmissionKey>,
+    F: FnOnce(&Diversifier) -> DiversifiedTransmissionKey,
 {
     assert!(plaintext.len() >= COMPACT_NOTE_SIZE);
 
@@ -72,7 +71,7 @@ where
         &domain.rho,
     ))?;
 
-    let pk_d = get_validated_pk_d(&diversifier)?;
+    let pk_d = get_pk_d(&diversifier);
 
     let recipient = Address::from_parts(diversifier, pk_d);
     let note = Option::from(Note::from_parts(recipient, value, domain.rho, rseed))?;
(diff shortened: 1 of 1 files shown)


src/note_encryption.rs
```diff
@@ -209,29 +208,18 @@ impl Domain for OrchardDomain {
...
         })
     }
 
     fn parse_note_plaintext_without_memo_ovk(
         &self,
         pk_d: &Self::DiversifiedTransmissionKey,
-        esk: &Self::EphemeralSecretKey,
-        ephemeral_key: &EphemeralKeyBytes,
+        _esk: &Self::EphemeralSecretKey,
+        _ephemeral_key: &EphemeralKeyBytes,
```
daira, line 219:
```suggestion
```
See https://github.com/zcash/librustzcash/pull/848/files#r1199198303

It may seem as though this is making life more difficult because removing these arguments is a breaking change, but strictly speaking the change to the guarantee of the function is a breaking API change anyway, even if not one that would otherwise be detected by the Rust compiler.

That is, it would be incorrect to mix a version of `orchard` from after this PR with a version of `zcash_note_encryption` from before https://github.com/zcash/librustzcash/pull/848 &mdash; not (as it happens) because it would be insecure per se, but because that usage hasn't been reviewed.

daira (approved):
utACK modulo comment (if [this suggestion](https://github.com/zcash/librustzcash/pull/848/files#r1199198303) is followed).

daira:
> ## [Codecov](https://app.codecov.io/gh/zcash/orchard/pull/394?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=zcash) Report
> 
> Patch coverage: **`100.00`**% and project coverage change: **`-0.37`** ⚠️
> 
> > Comparison is base [(`3619b86`)](https://app.codecov.io/gh/zcash/orchard/commit/3619b86d1c445018e283896dd28ba237f5e970e4?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=zcash) 83.42% compared to head [(`90e64cb`)](https://app.codecov.io/gh/zcash/orchard/pull/394?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=zcash) 83.06%.
> 
> Additional details and impacted files
> 
> [umbrella View full report in Codecov by Sentry](https://app.codecov.io/gh/zcash/orchard/pull/394?src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=zcash). 📢 Do you have feedback about the report comment? [Let us know in this issue](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=zcash).

Looking at the full report, it's actually an *increase* in coverage &mdash; which makes sense, because we previously had a error case that wasn't reachable. I think the reported overall decrease is due to nondeterminism.

src/note_encryption.rs
str4d: Mixing in the way you describe cannot happen, because the traits would not be version-compatible. Rust code is never dynamically linked internally.

nuttycom (approved):
utACK

PR #400, opened 2023-11-26
Small refactors

kdenhartog:
cc @str4d related to the docs updates. LMK if it's fine to just link to the design docs here. I couldn't find any sort of specifics about why the approach defined in the design docs were taken in the protocol spec, so it seemed like directly linking to the design docs was the better approach in most cases.


kdenhartog:
closing since the goal was to just to keep the link shortners in place, but just update so the HTTP 525 code is no longer thrown

PR #402, opened 2023-11-28
CI: Modernise workflows

str4d:
We now check in `Cargo.lock` for testing against MSRV.


.github/workflows/ci.yml
```diff
@@ -44,54 +64,43 @@ jobs:
...
-      - uses: actions/checkout@v2
-      - name: Setup mdBook
-        uses: peaceiris/actions-mdbook@v1
-        with:
-          mdbook-version: '0.4.5'
+      - uses: actions/checkout@v4
+      - name: Install mdbook and mdbook-katex
+        run: |
+          mkdir mdbook
+          curl -sSL https://github.com/rust-lang/mdBook/releases/download/v0.4.35/mdbook-v0.4.35-x86_64-unknown-linux-gnu.tar.gz | tar -xz --directory=./mdbook
```
daira, line 71:
```suggestion
          curl -sSL https://github.com/rust-lang/mdBook/releases/download/v0.4.35/mdbook-v0.4.35-x86_64-unknown-linux-gnu.tar.gz | tar -xz --directory=./mdbook
          curl -sSL https://github.com/lzanini/mdbook-katex/releases/download/v0.5.8pub/mdbook-katex-v0.5.8-x86_64-unknown-linux-gnu.tar.gz | tar -xz --directory=./mdbook
```

str4d: We don't need the `mdbook-katex` plugin to run book tests, as the LaTeX segments are outside the `rust` code blocks.

str4d: Oh wait, I see the actual bug: I copy-pasted the section from the book generator, and then removed the `mdbook-katex` line but forgot to change the name.

str4d:
[Force-pushed](https://github.com/zcash/orchard/compare/09b4eb45d3c9344a869972a2ee503da827e531d2..a12661cf195e316ce7db8e02dd1267ad509e31df) to fix a copy-paste error.

daira (approved):
utACK

PR #403, opened 2023-12-08
Add explicit control of padding to the Builder API.

nuttycom:


src/builder.rs
```diff
@@ -27,6 +27,46 @@ use crate::{
...
 
+/// An enumeration of rules for construction of dummy outputs that may be applied to Orchard bundle
+/// construction.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PaddingRule {
+    /// A rule that requires that at least the specified number of Orchard actions is constructed,
+    /// irrespective of whether any genuine outputs are being included.
+    Require(usize),
+    /// A rule that requires that at least the specified number of Orchard actions are constructed,
+    /// iff any genuine outputs are being included.
```
daira, line 38:
```suggestion
    /// iff any genuine actions are being included.
```

src/builder.rs
```diff
@@ -27,6 +27,46 @@ use crate::{
 
 const MIN_ACTIONS: usize = 2;
 
+/// An enumeration of rules for construction of dummy outputs that may be applied to Orchard bundle
+/// construction.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PaddingRule {
+    /// A rule that requires that at least the specified number of Orchard actions is constructed,
+    /// irrespective of whether any genuine outputs are being included.
```
daira, line 35:
```suggestion
    /// irrespective of whether any genuine actions are being included.
```

src/builder.rs
```diff
@@ -27,6 +27,46 @@ use crate::{
...
+impl PaddingRule {
+    /// The default padding rule, which ensures that the constructed bundle will contain at least 2
+    /// actions if any genuine Orchard spends or outputs are requested.
+    pub const DEFAULT: PaddingRule = PaddingRule::PadTo(MIN_ACTIONS);
+
+    /// Returns the number of logical actions that the builder will add to the bundle
+    /// as a consequence of this padding rule, given the specified number of spends
+    /// and recipients.
+    pub fn num_actions(&self, num_spends: usize, num_recipients: usize) -> usize {
+        let num_real_actions = core::cmp::max(num_spends, num_recipients);
```
daira, line 54:
```suggestion
    /// and outputs.
    pub fn num_actions(&self, num_spends: usize, num_outputs: usize) -> usize {
        let num_real_actions = core::cmp::max(num_spends, num_outputs);
```

daira (dismissed):
utACK

src/builder.rs
```diff
@@ -27,6 +27,46 @@ use crate::{
...
+    /// as a consequence of this padding rule, given the specified number of spends
+    /// and outputs.
+    pub fn num_actions(&self, num_spends: usize, num_outputs: usize) -> usize {
+        let num_real_actions = core::cmp::max(num_spends, num_outputs);
+
+        match self {
+            PaddingRule::Require(n) => core::cmp::max(num_real_actions, *n),
+            PaddingRule::PadTo(n) => {
+                if num_real_actions == 0 {
+                    0
```
str4d, line 60: This introduces the possibility of a panic: if the builder uses this padding rule and is given no spends and no outputs, then `Builder::build` will panic when `NonEmpty` is constructed.

If we want to support this, then we need to change `Builder::build` to return `Option<Bundle>` (which would bring it closer to the `sapling-crypto` API.

src/builder.rs
```diff
@@ -27,6 +27,46 @@ use crate::{
...
+    PadTo(usize),
+    /// A padding rule that specifies that no additional dummy Orchard actions are to be
+    /// constructed.
+    None,
+}
+
+impl PaddingRule {
+    /// The default padding rule, which ensures that the constructed bundle will contain at least 2
+    /// actions if any genuine Orchard spends or outputs are requested.
+    pub const DEFAULT: PaddingRule = PaddingRule::PadTo(MIN_ACTIONS);
```
str4d, line 48: Nit: the old behaviour is equivalent to `PaddingRule::Require(MIN_ACTIONS)` (which is why we could always return a bundle).

src/builder.rs
nuttycom: Done in 99d3d0ab71c3f6c8afbccb3b8f10213fe9dc2199

benches/circuit.rs
```diff
@@ -7,7 +7,7 @@ use criterion::{BenchmarkId, Criterion};
 use pprof::criterion::{Output, PProfProfiler};
 
 use orchard::{
-    builder::Builder,
+    builder::{Builder, PaddingRule},
```
str4d, line 10: Update these files for the refactor.

src/builder.rs
```diff
@@ -27,6 +27,43 @@ use crate::{
 
 const MIN_ACTIONS: usize = 2;
 
+/// An enumeration of rules for construction of dummy actions that may be applied to Orchard bundle
+/// construction.
```
str4d, line 31:
```suggestion
/// An enumeration of rules for Orchard bundle construction.
```

src/builder.rs
```diff
@@ -372,51 +414,51 @@ impl Builder {
...
 
-    /// Builds a bundle containing the given spent notes and recipients.
+    /// Builds a bundle containing the given spent notes and outputs.
     ///
     /// The returned bundle will have no proof or signatures; these can be applied with
     /// [`Bundle::create_proof`] and [`Bundle::apply_signatures`] respectively.
     pub fn build<V: TryFrom<i64>>(
         mut self,
         mut rng: impl RngCore,
+        padding_rule: &BundleType,
```
str4d, line 433:
```suggestion
        bundle_type: &BundleType,
```

src/builder.rs
```diff
@@ -372,51 +414,51 @@ impl Builder {
...
-            let num_spends = self.spends.len();
-            let num_recipients = self.recipients.len();
-            let num_actions = [num_spends, num_recipients, MIN_ACTIONS]
-                .iter()
-                .max()
-                .cloned()
-                .unwrap();
+        let num_real_spends = self.spends.len();
+        let num_real_outputs = self.outputs.len();
+        let num_actions = padding_rule
```
str4d, line 437:
```suggestion
        let num_actions = bundle_type
```

src/builder.rs
```diff
@@ -256,12 +298,12 @@ impl ActionInfo {
...
 
-/// A builder that constructs a [`Bundle`] from a set of notes to be spent, and recipients
+/// A builder that constructs a [`Bundle`] from a set of notes to be spent, and outputs
 /// to receive funds.
 #[derive(Debug)]
 pub struct Builder {
     spends: Vec<SpendInfo>,
-    recipients: Vec<RecipientInfo>,
+    outputs: Vec<OutputInfo>,
     flags: Flags,
```
str4d, line 307: To some extent, the flags are related to the `BundleType`. In particular, if `BundleType::Coinbase` is used, `flags.spends_enabled()` must be `true` (this check is missing from `Builder::build`; the `num_real_spends == 0` check is insufficient). So I'm not sure if this means we should be constructing to builder with `BundleType`, or ensuring we remember to add all necessary checks in `Builder::build`.

str4d (changes requested):
Reviewed as of b6bc39647e80c73e78fcad44421b0a799ae48b63.

src/builder.rs
nuttycom: The bundle type sort of implies the flags; what if `BundleType::Transactional` actually carried a flags payload, with the coinbase bundle type hardcoding the flags to `(false, true)`?

nuttycom: And then we check the flags in `Build` rather than in the `add_*` methods?

PR #408, opened 2023-12-20
Move some changelog entries from 0.3.0 to 0.4.0

### LINKED ISSUE #391: Move relevant changelog entries from 0.3.0 to 0.4.0
The changelog changes in this PR had a non-merge conflict with #363, because this PR was implemented as if it would be part of 0.3.0 but was in fact merged just afterwards. We need to fix the changelog to move the entries added by this PR into the 0.4.0 section.

_Originally posted by @str4d in https://github.com/zcash/orchard/issues/362#issuecomment-1505929607_

daira:
The changelog changes in #362 had a non-merge conflict with #363, because #362 was implemented as if it would be part of the 0.3.0 release but was in fact merged just afterwards.

fixes #391

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,15 @@ and this project adheres to Rust's notion of
   - `Builder::{spends, outputs}`
   - `SpendError`
   - `OutputError`
+- `orchard::keys`:
+  - `PreparedEphemeralPublicKey`
+  - `PreparedIncomingViewingKey`
+- impls of `memuse::DynamicUsage` for:
+  - `orchard::note::nullifier::Nullifier`
+  - `orchard::note_encryption::OrchardDomain`
+- impls of `Eq` for:
+  - `orchard::zip32::ChildIndex`
+  - `orchard::value::ValueSum`
 
 ### Changed
 - MSRV is now 1.60.0.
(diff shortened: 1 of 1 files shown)


nuttycom (dismissed):
utACK, doc-only

CHANGELOG.md
```diff
@@ -57,6 +57,15 @@ and this project adheres to Rust's notion of
   - `Builder::{spends, outputs}`
   - `SpendError`
   - `OutputError`
+- `orchard::keys`:
+  - `PreparedEphemeralPublicKey`
+  - `PreparedIncomingViewingKey`
+- impls of `memuse::DynamicUsage` for:
+  - `orchard::note::nullifier::Nullifier`
```
str4d, line 64:
```suggestion
  - `orchard::note::Nullifier`
```

str4d (approved):
ACK after path fix.

str4d (approved):
Re-ACK

CHANGELOG.md
daira: Oh, I looked at the code and it is actually defined in `orchard::note::nullifier::Nullifier`, but reexported at `orchard::note::Nullifier`, which is why I changed it. I guess it doesn't matter either way, this is fine.

str4d: The `nullifier` module is not public, so to users of this crate it is never part of the module path.

PR #409, opened 2023-12-21
Modify `BundleType` to exclude the anchor & allow no bundle to be produced.

nuttycom:


src/tree.rs
```diff
@@ -58,7 +58,12 @@ impl From<MerkleHashOrchard> for Anchor {
 }
 
 impl Anchor {
-    pub(crate) fn empty_tree() -> Anchor {
+    /// The anchor of the empty Orchard note commitment tree.
+    ///
+    /// This anchor does not correspond to any valid anchor for a spend, so it
+    /// may only be used for coinbase bundles or in circumstances where Orchard
+    /// functionality is not active.
```
str4d, line 65: It is entirely valid to use the empty tree in regular bundles. Doing so is equivalent to setting `enable_spends = false` - it proves publicly that the Orchard bundle spends no notes (but may still have real outputs). I think this comment is fine however because callers should reach for `enable_spends` in this case, rather than the empty anchor.

tests/builder.rs
```diff
@@ -42,7 +42,13 @@ fn bundle_chain() {
         // Use the empty tree.
         let anchor = MerkleHashOrchard::empty_root(32.into()).into();
 
-        let mut builder = Builder::new(BundleType::Transactional(Flags::SPENDS_DISABLED, anchor));
+        let mut builder = Builder::new(
+            BundleType::Transactional {
+                flags: Flags::SPENDS_DISABLED,
```
str4d, line 47: Nit: this publicly announces that spends are disabled, which in a real shielding bundle may be highly undesirable. This doesn't matter in a test, but we should avoid doing this in examples that downstream crate users are likely to look at for reference.

str4d (approved):
utACK 189257391a5726d7f3f2568d547a5ef8603c7156 with nits.

PR #417, opened 2024-02-06
Improve documentation for fixed-base constants

daira:
* Add a pointer to how the `Z` and `U` constants for fixed-base scalar multiplication can be reproduced.
* Correct errors and omissions in the documentation of the fixed-base generators.


str4d (approved):
ACK ae1ba27d8946e1c50de92d58dcdc08302563179f. Doc-only.

PR #428, opened 2024-07-05
Allow SpendValidatingKey to be constructed from bytes

### LINKED ISSUE #427: Allow `SpendValidatingKey` to be constructed from bytes
According to the [FROST book](https://frost.zfnd.org/zcash/technical-details.html) the  ASK is split using FROST  and it's verifying key is the `ak` that should be used to generate the Orchard Full Viewing Key.

In order for this to be possible the `pub(crate) fn from_bytes(bytes: &[u8]) -> Option<Self>` constructor from `SpendValidatingKey` should be pub instead of `pub (crate)`

pacu:
According to the [FROST book](https://frost.zfnd.org/zcash/technical-details.html) the ASK is split using FROST and its verifying key is the ak that should be used to generate the Orchard Full Viewing Key.

Allow SpendValidatingKey to be constructed from bytes

closes #427

As we agreed with @daira at Z|ECC, this adds a FROST feature flag so that public interface of the 
crate is not altered in its meaning for non-FROST applications.

Key derivation is intended to be done on a specific way for usual
applications. Interface changes introduce by FROST can pose a risk
for the rest of the use cases that don't involve the assumptions
of the FROST signature scheme protocol. we don't want non-FROST
cases to make use the helper functions FROST needs to derive the 
needed key elements.

There are other changes pending to make FROST work with Orchard.
Those changes will also be under this same feature flag

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### added
+- Added [Visibility crate](https://crates.io/crates/visibility) to modify
+visibility of methods and struct for the `unstable-frost` feature.
+- Added `SpendValidatingKey` serialization and deserialization from bytes
+visibility under the `unstable-frost` feature
+  - `orchard::keys::SpendValidatingKey`
 
 ## [0.8.0] - 2024-03-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,11 @@ and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-### added
-- Added [Visibility crate](https://crates.io/crates/visibility) to modify
-visibility of methods and struct for the `unstable-frost` feature.
-- Added `SpendValidatingKey` serialization and deserialization from bytes
-visibility under the `unstable-frost` feature
-  - `orchard::keys::SpendValidatingKey`
+### Added
+- `orchard::keys::SpendValidatingKey::{from_bytes, to_bytes}` behind the
+  `unstable-frost` feature flag. These are temporary APIs exposed for development
+  purposes, and will be replaced by type-safe FROST APIs once ZIP 312 key
+  generation is specified (https://github.com/zcash/zips/pull/883).
 
 ## [0.8.0] - 2024-03-25
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ zcash_note_encryption = "0.4"
 incrementalmerkletree = "0.5"
 zcash_spec = "0.1"
 zip32 = "0.1"
+visibility = "0.1.0"
 
 # Logging
 tracing = "0.1"
@@ -71,6 +72,7 @@ bench = false
 
 [features]
 default = ["multicore"]
+unstable-frost = []
 multicore = ["halo2_proofs/multicore"]
 dev-graph = ["halo2_proofs/dev-graph", "image", "plotters"]
 test-dependencies = ["proptest"]
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -186,12 +186,17 @@ impl SpendValidatingKey {
 
     /// Converts this spend validating key to its serialized form,
     /// I2LEOSP_256(ak).
+    #[cfg_attr(feature = "unstable-frost", visibility::make(pub))]
     pub(crate) fn to_bytes(&self) -> [u8; 32] {
         // This is correct because the wrapped point must have ỹ = 0, and
         // so the point repr is the same as I2LEOSP of its x-coordinate.
         <[u8; 32]>::from(&self.0)
     }
 
+    /// Attempts to convert these bytes into a spend validating key
+    /// from its serialized form, I2LEOSP_256(ak). Returns None if
+    /// it can't be created.
+    #[cfg_attr(feature = "unstable-frost", visibility::make(pub))]
     pub(crate) fn from_bytes(bytes: &[u8]) -> Option<Self> {
         <[u8; 32]>::try_from(bytes)
             .ok()
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -190,12 +190,14 @@ impl SpendValidatingKey {
     pub(crate) fn to_bytes(&self) -> [u8; 32] {
         // This is correct because the wrapped point must have ỹ = 0, and
         // so the point repr is the same as I2LEOSP of its x-coordinate.
-        <[u8; 32]>::from(&self.0)
+        let b = <[u8; 32]>::from(&self.0);
+        assert!(b[31] & 0x80 == 0);
+        b
     }
 
-    /// Attempts to convert these bytes into a spend validating key
-    /// from its serialized form, I2LEOSP_256(ak). Returns None if
-    /// it can't be created.
+    /// Attempts to parse a byte slice as a spend validating key, `I2LEOSP_256(ak)`.
+    ///
+    /// Returns `None` if the given slice does not contain a valid spend validating key.
     #[cfg_attr(feature = "unstable-frost", visibility::make(pub))]
     pub(crate) fn from_bytes(bytes: &[u8]) -> Option<Self> {
         <[u8; 32]>::try_from(bytes)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1440,6 +1440,7 @@ dependencies = [
  "serde",
  "subtle",
  "tracing",
+ "visibility",
  "zcash_note_encryption",
  "zcash_spec",
  "zip32",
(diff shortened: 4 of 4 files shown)


src/keys.rs
```diff
@@ -191,7 +191,16 @@ impl SpendValidatingKey {
...
+    /// I2LEOSP_256(ak).
+    #[cfg(feature = "FROST")]
+    pub fn into_bytes(&self) -> [u8; 32] {
+        self.to_bytes()
+    }
 
+    /// Attempts to convert these bytes into a spend validating key
+    /// from its serialized form, I2LEOSP_256(ak). Returns None if
+    /// it can't be created.
     pub(crate) fn from_bytes(bytes: &[u8]) -> Option<Self> {
```
conradoplg, line 204: In our crates we use the `visibility` crate to make things `pub` depending on a feature, so you could add

```suggestion
    #[cfg_attr(feature = "FROST", visibility::make(pub))]
    pub(crate) fn from_bytes(bytes: &[u8]) -> Option<Self> {
```

Cargo.toml
```diff
@@ -73,6 +73,7 @@ bench = false
 default = ["multicore"]
 multicore = ["halo2_proofs/multicore"]
 dev-graph = ["halo2_proofs/dev-graph", "image", "plotters"]
+FROST = []
```
conradoplg, line 76: I think lowercase is the norm for features?

conradoplg (commented):
I'm also interested this since we also need this for our FROST demo signing tool (we are currently using Ywallet code that uses modified versions of the crates so we just edited them to support this, but we're in the process of replacing that code).

I left a couple of random comments while I was taking a look, but it's probably better to wait for the maintainers to chime in first before changing anything.

src/keys.rs
pacu: @daira @nuttycom  @str4d, would you find this acceptable? Given that's a dependency, I'd be leveraging in this case I'd like to make sure you agree with going in such direction.

Cargo.toml
pacu: apparently the Rust Book says no thing about that but it doesn't portrait any uppercase examples so I will change this

src/keys.rs
pacu: @conradoplg would this work for entire `struct`s as well?

Because `NullifierDerivingKey` and some of its functions will have to be exposed as `pub` instead of `pub(crate)`. It's possible that the use of this visibility crate, although not documented enough, will be a good addition for what we need to achieve given that otherwise we would have to manually achieve the same and gate it through the `frost` feature config.

str4d:
I'm fine with adding this kind of API behind an unstable cfg option, but I think that this API is not what we will want in the final production release, because I suspect we will not want people to pass around `ak`s on their own (that will lead to partial loss of necessary details, and sadness).

To figure out the production API, we need to first finish the `Key Generation` part of ZIP 312. I'd started sketching that out in https://github.com/zcash/zips/pull/662#discussion_r1465663928 which I will move into a separate PR; once we've figured that out, we can provide higher-level APIs in this crate that produce the actual types we need.

pacu:
> I suspect we will not want people to pass around `ak`s on their own (that will lead to partial loss of necessary details, and sadness).

Yes this is exactly what we don't want for non-FROST users. 

@str4d, What about adding the visibility crate to handle the API changes as @conradoplg suggested? 

We'd be adding `#[cfg_attr(feature = "frost", visibility::make(pub))]` but this is probably a dependency that is not feature-gated because it has to compile even when the frost feature is not selected

src/keys.rs
daira: `struct SpendValidatingKey` is already pub. Its wrapped field is not and should not be, but as it happens you can obtain it anyway as `svk.randomize(pallas::Scalar::zero())`.

src/keys.rs
```diff
@@ -208,6 +217,14 @@ impl SpendValidatingKey {
...
             .map(SpendValidatingKey)
     }
+
+    /// Attempts to convert these bytes into a spend validating key
+    /// from its serialized form, I2LEOSP_256(ak). Returns None if
+    /// it can't be created.
+    #[cfg(feature = "frost")]
+    pub fn with_bytes(bytes: &[u8]) -> Option<Self> {
+        SpendValidatingKey::from_bytes(bytes)
+    }
```
daira, line 227: This isn't needed if `#[cfg_attr(feature = "frost", visibility::make(pub))]` is used. In any case the method that is conditionally exposed should be called `from_bytes`.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### added
+- Added `SpendValidatingKey` serialization and deserialization from bytes
+visibility under the `FROST` feature
```
daira, line 11:
```suggestion
  with visibility under the "frost" feature.
```

src/keys.rs
pacu: Cool I'll get `visibility` crate in then. Thank you!

src/keys.rs
pacu: I added the visibility crate to implement the required changes on this PR

conradoplg (commented):
LGTM!

str4d:
> To figure out the production API, we need to first finish the `Key Generation` part of ZIP 312. I'd started sketching that out in [zcash/zips#662 (comment)](https://github.com/zcash/zips/pull/662#discussion_r1465663928) which I will move into a separate PR; once we've figured that out, we can provide higher-level APIs in this crate that produce the actual types we need.

I opened https://github.com/zcash/zips/pull/883 as that separate PR. @conradoplg @pacu let's figure out and specify the FROST-Orchard key tree interaction there.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### added
+- Added [Visibility crate](https://crates.io/crates/visibility) to modify
+visibility of methods and struct for the `unstable-frost` feature.
+- Added `SpendValidatingKey` serialization and deserialization from bytes
+visibility under the `unstable-frost` feature
+  - `orchard::keys::SpendValidatingKey`
```
str4d, line 14:
```suggestion
### Added
- `orchard::keys::SpendValidatingKey::{from_bytes, to_bytes}` behind the
  `unstable-frost` feature flag. These are temporary APIs exposed for development
  purposes, and will be replaced by type-safe FROST APIs once ZIP 312 key
  generation is specified (https://github.com/zcash/zips/pull/883).
```

(there is no need to mention the `visibility` crate, as it is an internal implementation detail)

src/keys.rs
```diff
@@ -186,12 +186,17 @@ impl SpendValidatingKey {
 
     /// Converts this spend validating key to its serialized form,
     /// I2LEOSP_256(ak).
+    #[cfg_attr(feature = "unstable-frost", visibility::make(pub))]
     pub(crate) fn to_bytes(&self) -> [u8; 32] {
         // This is correct because the wrapped point must have ỹ = 0, and
         // so the point repr is the same as I2LEOSP of its x-coordinate.
         <[u8; 32]>::from(&self.0)
```
daira, line 193:
```suggestion
        let b = <[u8; 32]>::from(&self.0);
        assert!(b[31] & 0x80 == 0);
        b
```

src/keys.rs
```diff
@@ -186,12 +186,17 @@ impl SpendValidatingKey {
...
     }
 
+    /// Attempts to convert these bytes into a spend validating key
+    /// from its serialized form, I2LEOSP_256(ak). Returns None if
+    /// it can't be created.
```
str4d, line 198:
```suggestion
    /// Attempts to parse a byte slice as spend validating key, `I2LEOSP_256(ak)`.
    ///
    /// Returns `None` if the given slice does not contain a valid spend validating key.
```

str4d (dismissed):
utACK 96bda92e3ed689bdc2fda334b0712ec80d08012d one the following changes are made.

daira (dismissed):
utACK with a non-blocking suggestion.

daira (approved):
utACK 2bb3a08b9fe94eceefe73fb6ed08567221566498

str4d:
> According to the [FROST book](https://frost.zfnd.org/zcash/technical-details.html) the ASK is split using FROST and its verifying key is the ak that should be used to generate the Orchard Full Viewing Key.

Just to be _very_ clear here, the FROST book is _not_ normative, except as it is incorporated by reference into the protocol spec or a ZIP (which to my knowledge it is currently not). It is very _informative_, but anything we put into the public crates needs to be tied to a normative reference. In this case, that is [ZIP 312](https://zips.z.cash/zip-0312), so anything we do needs to be specified there (i.e. in the PR I linked above).

str4d (approved):
Re-utACK 2bb3a08b9fe94eceefe73fb6ed08567221566498

PR #429, opened 2024-07-09
Use generalized Lookup version of Halo2

ConstanceBeguier:
This PR updates the `orchard` crate to use the generalized version of the `halo2` crate ([link](https://github.com/zcash/halo2/pull/823)). It contains two key changes:

- Brings the `LookupRangeCheck` trait into scope (this maintains backward compatibility and allows existing code to function without additional modifications).
- Updates `Cargo.toml` to point to the `improve-backward-compatibility-without-zsa` branch of the `halo2` crate.


Cargo.toml
```diff
@@ -29,8 +29,8 @@ blake2b_simd = "1"
 ff = "0.13"
 fpe = "0.6"
 group = { version = "0.13", features = ["wnaf-memuse"] }
-halo2_gadgets = "0.3"
-halo2_proofs = { version = "0.3", default-features = false, features = ["batch", "floor-planner-v1-legacy-pdqsort"] }
+halo2_gadgets = { git = "https://github.com/QED-it/halo2", branch = "improve-backward-compatability-without-zsa" }
```
str4d, line 32: In general, we do not merge PRs that change dependencies in this way, for two reasons:
- It is a security issue to depend on a branch rather than a commit.
- We use `[patch.crates-io]` instead for patching in-progress changes (and we then propagate those sections through the various downstream repos).

str4d (commented):
Changes here LGTM once the halo2 PR is merged.

ConstanceBeguier:
I just rebased on main branch and updated Cargo.toml to use  patch.crates-io for halo2 repos

Cargo.toml
ConstanceBeguier: I updated Cargo.toml to use patch.crates-io for halo2 repos.

ConstanceBeguier:
Now that the halo2 PR https://github.com/zcash/halo2/pull/823 is merged. This PR could be merged.

ConstanceBeguier:
Replaced by #470

Issue #430, opened 2024-07-29
API changes required for FROST


conradoplg:
I've updated our signing tool to the latest ECC crates and mapped out the API changes that are required in this crate:

- Allowing constructing a SpendValidatingKey from bytes. Done in https://github.com/zcash/orchard/pull/428
- Allow creating a FullViewingKey from a given SpendValidatingKey. The other components can be randomly generated, but there may be some utility in specifying them too.
  - Locally I did this by adding a `FullViewingKey::from_sk_ak(sk: &SpendingKey, ak: SpendValidatingKey)` method, but not sure what is the best approach.
- Allowing getting the generated randomizers (alphas), or preferrably, allow specifying which alphas to use.
  -  I did the former by adding a `alpha()` getter to `SigningParts` (also added `ak()` getter which is useful to sanity check if I'm accessing the right `SigningParts`). The latter would be more complicated and would require passing the alphas in `Builder::build()` (which is awkward) or doing some refactoring.
  - The reason the latter would be preferred is that Re-randomized FROST specifies [a mechanism](https://zips.z.cash/zip-0312#randomizer-generation) to generate alphas from the FROST signing package along with randomness, making sure that different packages will have different alphas. But the crate seems to assume that randomness will always be generated correctly (which is a fair assumption in this context) so this might not be needed.
  - While writing this I also realized that this crate does not seem to strictly follow Zcash's spec on [how the alpha is generated](https://zips.z.cash/protocol/protocol.pdf#concretereddsa) (it doesn't use the H* hash function). Again, not a big deal for the same reason.

My intention is to spark discussion on these API changes but let me know if you'd prefer individual tickets for each of those. I can also create PRs if you're happy with the proposed APIs or can also create PRs with other APIs you suggest.

conradoplg:
I realized there is a chicken-and-egg problem regarding the second approach to alpha handling:

- The sighash, which is the thing FROST signs, depends on rk, which depends on alpha
- The "FROST alpha generation" requires the SigningPackage, which contains the message, which is the sighash

So it seems the "FROST alpha generation" can't be used.

pacu:
please edit:

- Allow creating a UnifiedFullViewingKey from a FullViewingKey. There is a new() method which could be used but it's gated under the test-dependencies feature. https://github.com/zcash/librustzcash/pull/1478

pacu:
please edit:
[...]

PR #433, opened 2024-07-31
add SigningParts getters

conradoplg:
Depends on #428

See #430 for rationale

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -751,6 +751,18 @@ pub struct SigningParts {
     alpha: pallas::Scalar,
 }
 
+#[cfg(feature = "unstable-frost")]
+impl SigningParts {
+    /// Return the spend validating key for this action.
+    pub fn ak(&self) -> &SpendValidatingKey {
+        &self.ak
+    }
+    /// Return the randomization for this action.
+    pub fn alpha(&self) -> pallas::Scalar {
+        self.alpha
+    }
+}
+
 /// Marker for an unauthorized bundle with no signatures.
 #[derive(Clone, Debug)]
 pub struct Unauthorized {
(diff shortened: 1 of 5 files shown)


src/keys.rs
```diff
@@ -334,6 +334,15 @@ impl From<FullViewingKey> for SpendValidatingKey {
 }
 
 impl FullViewingKey {
+    /// Create FVK from the given sk and ak.
+    pub fn from_sk_ak(sk: &SpendingKey, ak: SpendValidatingKey) -> Self {
```
str4d, line 338: The documentation here is very insufficient:
- It makes no mention of how `sk` and `ak` are checked or combined (in particular, that you get a different FVK than if you just go from `sk`).
- It makes no mention of FROST (or why you would want to use this API).
- It is not behind a feature flag so is exposing a normative public API which I do not want to be present without a corresponding specification (which will eventually be ZIP 312 once https://github.com/zcash/zips/pull/883 is finished).

Also, not documented as an addition in the changelog.

src/builder.rs
```diff
@@ -917,6 +917,18 @@ pub struct SigningParts {
...
 }
 
+#[cfg(feature = "unstable-frost")]
+impl SigningParts {
+    /// Return the spend validating key for this action.
+    pub fn ak(&self) -> &SpendValidatingKey {
+        &self.ak
+    }
+    /// Return the randomization for this action.
+    pub fn alpha(&self) -> pallas::Scalar {
```
str4d, line 927: These getters are only necessary for the older compile-time state machine builder API. Since this PR was opened, we've deployed PCZTs which already include these getters implicitly, and PCZTs (and runtime state machine builder APIs in general) are the general direction we will be heading, so these aren't technically necessary. That being said, as these are behind an unstable feature flag I wouldn't mind them going in for testing purposes, on the understanding that they likely will eventually be removed (along with the corresponding builder logic).

str4d (changes requested):
Not including this PR in `orchard 0.11.0`

conradoplg:
Closing in favor of https://github.com/zcash/orchard/pull/475

PR #440, opened 2024-11-27
Implement PCZT support

str4d:


src/pczt.rs
```diff
@@ -0,0 +1,406 @@
...
+
+    /// The ZIP 32 derivation path at which the spending key can be found for the note
+    /// being spent.
+    pub(crate) zip32_derivation: Option<Zip32Derivation>,
+
+    /// The spending key for this spent note, if it is a dummy note.
+    ///
+    /// - This is chosen by the Constructor.
+    /// - This is required by the IO Finalizer, and is cleared by it once used.
+    /// - Signers MUST reject PCZTs that contain `dummy_sk` values.
```
str4d, line 185: - I both named this `dummy_sk` and added a "MUST reject" here because I do not want PCZTs to get serialized with non-dummy spending keys.
- I stored the dummy note's `sk` instead of the smaller-scoped `ask`, because in Orchard we impose validity requirements at parse time of `sk` (specifically to ensure that a valid `ivk` is produced), and we already had APIs for handling that parsing here. If desired this can be scoped down to `dummy_ask`, but given that (again) this field is only for dummy notes, I think `sk` should be fine here.

src/pczt/parse.rs
```diff
@@ -0,0 +1,329 @@
...
+    value::{NoteValue, Sign, ValueCommitTrapdoor, ValueCommitment, ValueSum},
+    Address, Anchor, Proof, NOTE_COMMITMENT_TREE_DEPTH,
+};
+
+impl Bundle {
+    /// Parses a PCZT bundle from its component parts.
+    pub fn parse(
+        actions: Vec<Action>,
+        flags: u8,
+        value_sum: (u64, bool),
```
str4d, line 25: I decided on this serialization format for `ValueSum` (rather than e.g. `i128`) because it has no edge cases; all values are valid.

src/pczt/tx_extractor.rs
```diff
@@ -0,0 +1,156 @@
...
+    MissingProof,
+    /// The Transaction Extractor role requires all `zkproof` fields to be set.
+    MissingSpendAuthSig,
+    /// The value sum does not fit into a `valueBalance`.
+    ValueSumOutOfRange,
+}
+
+/// Authorizing data for a bundle of actions that is just missing a binding signature.
+#[derive(Debug)]
+pub struct Unbound {
```
nuttycom, line 122: nit: this name isn't particularly intuitive to me, but I don't have a better suggestion; maybe `BindingSigInputs` or something?

src/pczt/tx_extractor.rs
```diff
@@ -0,0 +1,156 @@
...
+
+impl Authorization for Unbound {
+    type SpendAuth = redpallas::Signature<SpendAuth>;
+}
+
+impl<V> crate::Bundle<Unbound, V> {
+    /// Verifies the given sighash with every `spend_auth_sig`, and then binds the bundle.
+    ///
+    /// Returns `None` if the given sighash does not validate against every `spend_auth_sig`.
+    pub fn bind<R: RngCore + CryptoRng>(
```
nuttycom, line 135: Maybe `apply_binding_sig` or `bind_to_sighash`?

nuttycom (approved):
utACK e7a2945333c65818fdd2d7233377cce1ea92f6f0 with minor questions / naming nits. Overall this looks excellent.

src/builder.rs
```diff
@@ -552,19 +643,110 @@ impl Builder {
...
+
+            // Create the actions.
+            let (actions, circuits): (Vec<_>, Vec<_>) =
+                pre_actions.into_iter().map(|a| a.build(&mut rng)).unzip();
+
+            // Verify that bsk and bvk are consistent.
+            let bvk = (actions.iter().map(|a| a.cv_net()).sum::<ValueCommitment>()
+                - ValueCommitment::derive(value_balance, ValueCommitTrapdoor::zero()))
+            .into_bvk();
+            assert_eq!(redpallas::VerificationKey::from(&bsk), bvk);
```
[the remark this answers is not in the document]
str4d: Internal consistency check; there's a straight line between where we sample `rcv` and compute `cv_net`, and here.

Note how in the PCZT version I turned this check into an error because we're serializing in between and thus there's a chance that a Constructor put bad `cv_net` or `rcv` data into the PCZT that would cause this to fail.

src/pczt.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// This is set by the Creator. The Constructor MUST only add spends and outputs that
+    /// are consistent with these flags (i.e. are dummies as appropriate).
+    pub(crate) flags: Flags,
+
+    /// The sum of the values of all `actions`.
+    ///
+    /// This is initialized by the Creator, and updated by the Constructor as spends or
+    /// outputs are added to the PCZT. It enables per-spend and per-output values to be
+    /// redacted from the PCZT after they are no longer necessary.
+    pub(crate) value_sum: ValueSum,
```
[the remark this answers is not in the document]
str4d: The mut getter here is actually for the Fun Third Option: mutating existing actions so that signatures can be applied. In the PCZT format PR I implemented all the roles as consuming / transforming, but the `orchard` APIs are what get used under the hood to do the transformations, and for the scale of PCZTs a mapping-based API with the necessary granularity would be ridiculously verbose and onerous to work with.

If we don't want to expose the dual capabilities of "modify existing Actions" and "add / remove Actions" here, then I can replace this with a method that returns `&mut [Action]` instead (hiding the `Vec` nature and only exposing the modification capability). That would be sufficient for now as the "add Actions" capability is currently provided by `Builder` (I have no plans to enable piece-wise PCZT building in this first pass).

src/pczt/tx_extractor.rs
str4d: The name appears primarily in type signatures, i.e. `Bundle<Unbound, ZatBalance>`; I chose it to make sense there.

src/pczt.rs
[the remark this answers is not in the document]
nuttycom: Hiding the `Vec` nature sounds like the right approach to me. Right now it presents a footgun, which we can eliminate.

src/pczt/io_finalizer.rs
```diff
@@ -0,0 +1,67 @@
...
+            - ValueCommitment::derive(self.value_sum, ValueCommitTrapdoor::zero()))
+        .into_bvk();
+        if redpallas::VerificationKey::from(&bsk) != bvk {
+            return Err(IoFinalizerError::ValueCommitMismatch);
+        }
+        self.bsk = Some(bsk);
+
+        // Add signatures to dummy spends.
+        for action in self.actions.iter_mut() {
+            if let Some(sk) = action.spend.dummy_sk.take() {
```
str4d, line 45: Point the `take()` out.

src/pczt.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// the values and change involved in the transaction.
+    ///
+    /// This exposes the value to all participants. For Signers who don't need this
+    /// information, we can drop the values and compress the rcvs into the bsk global.
+    pub(crate) value: Option<NoteValue>,
+
+    /// The seed randomness for the output.
+    ///
+    /// - This is set by the Constructor.
+    /// - This is required by the Prover, instead of disclosing `shared_secret` to them.
```
str4d, line 225: - Remove `shared_secret` reference.
- Describe how the Signer can use `rseed` and `recipient` to decrypt and verify `enc_ciphertext` (in the case where `ock` is not present, or `OvkPolicy::None` was used to make the data unrecoverable from chain by the sender).

src/pczt.rs
```diff
@@ -0,0 +1,406 @@
...
+                auth_path[..].try_into().unwrap(),
+            );
+            let anchor = root.into();
+            assert_eq!(anchor, merkle_path.root(cmx));
+            (anchor, merkle_path)
+        };
+
+        // Run the Creator and Constructor roles.
+        let mut builder = Builder::new(BundleType::DEFAULT, anchor);
+        builder.add_spend(fvk, note, merkle_path).unwrap();
```
str4d, line 369: Modify this test to send an output to a different recipient (with change), and then verify that the PCZT contains sufficient information to decrypt and check `enc_ciphertext`.

src/pczt/tx_extractor.rs
```diff
@@ -0,0 +1,156 @@
...
+        self,
+        sighash: [u8; 32],
+        rng: R,
+    ) -> Option<crate::Bundle<Authorized, V>> {
+        let bound = self.map_authorization(
```
str4d, line 140: Move this to after the check so we don't bother making the signature if the binding would be wrong.

src/pczt/tx_extractor.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+/// Errors that can occur while extracting a regular Orchard bundle from a PCZT bundle.
+#[derive(Debug)]
+pub enum TxExtractorError {
+    /// The Transaction Extractor role requires `bsk` to be set.
+    MissingBindingSignatureSigningKey,
+    /// The Transaction Extractor role requires `zkproof` to be set.
+    MissingProof,
+    /// The Transaction Extractor role requires all `zkproof` fields to be set.
```
str4d, line 114:
```suggestion
    /// The Transaction Extractor role requires all `spend_auth_sig` fields to be set.
```

src/pczt/tx_extractor.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// [regular `Bundle`]: crate::Bundle
+    pub fn extract<V: TryFrom<i64>>(
+        self,
+    ) -> Result<Option<crate::Bundle<Unbound, V>>, TxExtractorError> {
+        self.to_tx_data(
+            |action| {
+                Ok(redpallas::Signature::from(
+                    action
+                        .spend
+                        .spend_auth_sig
```
str4d, line 36: Copy-pasta: this is already a `redpallas::Signature`.

src/pczt/parse.rs
```diff
@@ -0,0 +1,329 @@
...
+                .try_into()
+                .map_err(|_| ParseError::InvalidOutCiphertext)?,
+        };
+
+        let recipient = recipient
+            .as_ref()
+            .map(|r| {
+                Address::from_raw_address_bytes(r)
+                    .into_option()
+                    .ok_or(ParseError::InvalidSpendRecipient)
```
str4d, line 239: Rename this error to `InvalidRecipient`.

src/pczt/io_finalizer.rs
```diff
@@ -0,0 +1,67 @@
...
+                    .sign(sighash, &ask, &mut rng)
+                    .map_err(IoFinalizerError::DummySignature)?;
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// Errors that can occur while extracting a regular Orchard bundle from a PCZT bundle.
```
str4d, line 57: Fix this comment.

str4d:
[Force-pushed](https://github.com/zcash/orchard/compare/e7a2945333c65818fdd2d7233377cce1ea92f6f0..782d3abfe763dd38c0378e83268bf3df20758116) to address comments and add missing serialization APIs.

str4d:
[Force-pushed](https://github.com/zcash/orchard/compare/782d3abfe763dd38c0378e83268bf3df20758116..9b1140c607a1140c538d2eaeeb4a8e632546b452) to address another comment I missed.

str4d:
[Force-pushed](https://github.com/zcash/orchard/compare/9b1140c607a1140c538d2eaeeb4a8e632546b452..cb404f817480f267d2e138a268a55bac6474794a) to address another comment I missed, and fix clippy lints.

src/builder.rs
```diff
@@ -261,6 +262,48 @@ impl SpendInfo {
             &path_root == anchor
         }
     }
+
+    /// Builds the spend half of an action.
+    ///
+    /// Defined in [Zcash Protocol Spec § 4.7.3: Sending Notes (Orchard)][orchardsend].
```
daira, line 268:
```suggestion
    /// The returned values are chosen as in [Zcash Protocol Spec § 4.7.3: Sending Notes (Orchard)][orchardsend].
```

src/builder.rs
```diff
@@ -301,6 +344,54 @@ impl OutputInfo {
...
+    ) -> crate::pczt::Output {
+        let (note, cmx, encrypted_note) = self.build(cv_net, nf_old, rng);
+
+        crate::pczt::Output {
+            cmx,
+            encrypted_note,
+            recipient: Some(self.recipient),
+            value: Some(self.value),
+            rseed: Some(*note.rseed()),
+            // TODO: Save this?
```
daira, line 389: Clarify this TODO.

nuttycom (approved):
utACK cb404f817480f267d2e138a268a55bac6474794a with nonblocking question about how to handle `TransmittedNoteCiphertext` in the future.

src/pczt.rs
```diff
@@ -0,0 +1,434 @@
...
+    /// A commitment to the new note being created.
+    pub(crate) cmx: ExtractedNoteCommitment,
+
+    /// The transmitted note ciphertext.
+    ///
+    /// This contains the following PCZT fields:
+    /// - `ephemeral_key`
+    /// - `enc_ciphertext`
+    /// - `out_ciphertext`
+    pub(crate) encrypted_note: TransmittedNoteCiphertext,
```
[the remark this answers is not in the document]
str4d: This is the parsed typed type, so it can be whatever we want. The PCZT encoding uses `Vec<u8>` for both ciphertexts in anticipation of NU7.

src/pczt.rs
```diff
@@ -0,0 +1,434 @@
...
+    pub(crate) dummy_sk: Option<SpendingKey>,
+
+    /// Proprietary fields related to the note being spent.
+    pub(crate) proprietary: BTreeMap<String, Vec<u8>>,
+}
+
+/// Information about an Orchard output within a transaction.
+#[derive(Getters)]
+#[getset(get = "pub")]
+pub struct Output {
```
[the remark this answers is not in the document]
str4d: The transaction version is runtime-specified for PCZTs, so this probably should be as well.

src/builder.rs
str4d:
```suggestion
            // TODO: Extract ock from the encryptor and save it so
            // Signers can check `out_ciphertext`.
```

nuttycom (approved):
re-utACK d132b5b02cac57ce5185a4d454ec7c991b6939a5

src/pczt/parse.rs
```diff
@@ -0,0 +1,331 @@
...
+    keys::{FullViewingKey, SpendingKey},
+    note::{ExtractedNoteCommitment, Nullifier, RandomSeed, Rho, TransmittedNoteCiphertext},
+    primitives::redpallas::{self, SpendAuth},
+    tree::{MerkleHashOrchard, MerklePath},
+    value::{NoteValue, Sign, ValueCommitTrapdoor, ValueCommitment, ValueSum},
+    Address, Anchor, Proof, NOTE_COMMITMENT_TREE_DEPTH,
+};
+
+impl Bundle {
+    /// Parses a PCZT bundle from its component parts.
```
daira, line 21:
```suggestion
    /// Parses a PCZT bundle from its component parts.
    /// `value_sum` is represented as `(magnitude, is_negative)`.
```

daira (approved):
utACK

daira (approved):
Doc-only self-utACK commits/5b6cc68b746c4dab37e43d09314da98d5ee99c21

PR #443, opened 2024-12-12
Add methods for validating aspects of PCZT bundles

str4d:


src/pczt/verify.rs
```diff
@@ -0,0 +1,176 @@
...
+            (_, Some(fvk), Some(value)) if value.inner() == 0 => Ok(fvk),
+            (_, None, Some(value)) if value.inner() == 0 => Err(VerifyError::MissingFullViewingKey),
+            // If the FVK field has been pruned, assume the caller provided the correct FVK.
+            (Some(expected_fvk), None, _) => Ok(expected_fvk),
+            // This is not a dummy note; if the FVK field is present, it must match.
+            (Some(expected_fvk), Some(fvk), _) if fvk == expected_fvk => Ok(fvk),
+            (Some(_), Some(_), _) => Err(VerifyError::MismatchedFullViewingKey),
+            (None, Some(fvk), _) => Ok(fvk),
+            (None, None, _) => Err(VerifyError::MissingFullViewingKey),
+        }
```
daira, line 51: What is the motivation for treating dummy notes differently? The main effect seems to be that if we pass a different `expected_fvk` than `self.fvk` when both are `Some`, when we will allow that for a dummy note. But why?

If not for that case, we might as well do the simpler:
```suggestion
        match (expected_fvk, self.fvk.as_ref()) {
            (Some(expected_fvk), Some(fvk)) if fvk == expected_fvk => Ok(fvk),
            (Some(_), Some(_)) => Err(VerifyError::MismatchedFullViewingKey),
            (Some(expected_fvk), None) => Ok(expected_fvk),
            (None, Some(fvk)) => Ok(fvk),
            (None, None) => Err(VerifyError::MissingFullViewingKey),
        }
```

src/pczt/verify.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+}
+
+impl super::Output {
+    /// Verifies that the `cmx` field is consistent with the note fields.
+    ///
+    /// Requires that the following optional fields are set:
+    /// - `recipient`
+    /// - `value`
+    /// - `rseed`
```
daira, line 124:
```suggestion
    /// - `rseed`
    ///
    /// `spend` must be the Spend from the same Orchard action.
```

daira (dismissed):
utACK with suggestion.

nuttycom (dismissed):
utACK 3d951b4201a63f3c07cba8b179dd9abde142cf33 modulo @daira's suggestions

daira (approved):
utACK 9b99350b38f309676a85bb2779bb9173b42b398e

PR #444, opened 2024-12-13
pczt: Add output field for storing the user-facing address

str4d:
This is necessary in order for Signers to display the address encoding that a user is expecting to confirm.


nuttycom (approved):
utACK 6011de11dfd9b62cad9e04aba1cabfe2a87893d3 (in pairing)

nuttycom (approved):
utACK 7a44e3279b5747819022c4d8f4474fa79b2d9746

PR #447, opened 2024-12-16
Keystone preparations

str4d:
This includes everything for `no-std` support except the `std` feature flag.


nuttycom (approved):
utACK f99b6565a78763b58dac792d7492c55067bae680

PR #450, opened 2024-12-19
Add CI workflow for verifying no_std compatibility.

nuttycom:


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to Rust's notion of
 - `orchard::pczt::Zip32Derivation::extract_account_index`
 
 ### Changed
+- MSRV is now 1.66
```
str4d, line 14: What requires or motivates the MSRV bump in `orchard`? Is it just "parity with `sapling-crypto`"? (And what was the motivation there?)

str4d (approved):
utACK d983ca0d5e73c7739d6b26db3705b5e8781f95e0 with question.

CHANGELOG.md
nuttycom: This is the easiest workaround for the `constant_time_eq` MSRV breaking bump; I figure that `1.66` is old enough now that it should basically be fine to do this.

PR #451, opened 2024-12-30
Install libfontconfig1-dev to fix CI build on updated Ubuntu runner

dmidem:
GitHub updated ubuntu-latest from Ubuntu 22.04 to 24.04, which looks like it no longer includes fontconfig by default. This PR adds installation of libfontconfig1-dev so the yeslogic-fontconfig-sys crate (a transitive dependency) can build successfully again.


dmidem:
Closed because it was mistakenly opened here instead of the QED-it fork.

PR #460, opened 2025-03-19
CI: add `--features circuit` for `no_std` platforms, and rename that step to `build-nostd`

daira:


daira:
This is harder than I thought, and will require releases of at least `pasta_curves` and the `halo2` crates (to avoid std dependencies via `rand_core`). Going to leave this on the back burner for now.

PR #473, opened 2025-12-04
More release preparations

str4d:


CHANGELOG.md
```diff
@@ -13,6 +13,25 @@ and this project adheres to Rust's notion of
 ### Changed
 - `orchard::pczt::SignerError` has added variants:
   - `InvalidExternalSignature`
+- All error enums in this crate are now `#[non_exhaustive]`, to allow future
+  error variants to be added without a SemVer break:
```
str4d, line 30: Note this is not all *errors*, as there are some zero-sized structs used as errors (e.g. `orchard::builder::OutputError`. Making those non-exhaustive would let us add fields to them, but not turn them into enums; we'd either need to make them one-variant errors right now, or just defer that until the first time we need to distinguish a second error variant.

nuttycom: Hmmm... If they're intended to be errors, they should probably be enums rather than structs. I would tend to prefer that we do that all at once with this change.

nuttycom (dismissed):
utACK 9bf58f066ee68e018ab2ef45e41583c6835ef5b1

nuttycom (approved):
utACK 3661d41da09b155253d71e32cfd0a03f7180beda

PR #475, opened 2025-12-12
Add FullViewingKey::from_sk_ak_incompatible_with_quantum_recoverability_and_will_be_removed()

conradoplg:
A revisit of https://github.com/zcash/orchard/pull/433/

This is motivated by us adding PCZT support to https://github.com/ZcashFoundation/frost-tools/ with the final goal of creating a PoC of FROST with zcash-devtool.

This allows FROST usage under the `unstable-frost` feature. After https://zips.z.cash/draft-ecc-quantum-recoverability we can probably keep the same API and just replace how the derivation is done.


src/keys.rs
```diff
@@ -334,6 +334,25 @@ impl From<FullViewingKey> for SpendValidatingKey {
...
+    ///
+    /// When using FROST or any threshold signature scheme, we might need to
+    /// create a FullViewingKey from an specific SpendValidatingKey which was
+    /// not derived from a SpendingKey (since it might be generated using a DKG
+    /// protocol, which generates a random SpendValidatingKey). This method
+    /// achieves this by creating the FullViewingKey from the SpendingKey as
+    /// usual, with the exception of SpendValidatingKey which is provided by the
+    /// caller.
+    #[cfg(feature = "unstable-frost")]
+    pub fn from_sk_ak(sk: &SpendingKey, ak: SpendValidatingKey) -> Self {
```
daira:
```suggestion
    pub fn from_sk_ak_incompatible_with_quantum_recoverability_and_will_be_removed(sk: &SpendingKey, ak: SpendValidatingKey) -> Self {
```

src/keys.rs
```diff
@@ -334,6 +334,25 @@ impl From<FullViewingKey> for SpendValidatingKey {
...
+    /// Create a FullViewingKey from the given SpendingKey `sk` and the given
+    /// SpendValidatingKey `ak`.
+    ///
+    /// When using FROST or any threshold signature scheme, we might need to
+    /// create a FullViewingKey from an specific SpendValidatingKey which was
+    /// not derived from a SpendingKey (since it might be generated using a DKG
+    /// protocol, which generates a random SpendValidatingKey). This method
+    /// achieves this by creating the FullViewingKey from the SpendingKey as
+    /// usual, with the exception of SpendValidatingKey which is provided by the
+    /// caller.
```
daira: This is inherently incompatible with ZIP 2005, section [Usage with FROST](https://zips.z.cash/zip-2005#usagewithfrost).

ZIP 2005 is intended to ensure that when keys are derived in the way it specifies, *which for non-FROST keys matches the way previously specified*, notes sent to those keys are recoverable. However, the spec didn't define how to generate FROST keys.

I anticipated this problem and designed a way for FROST keys to be compatible with QR. The problem is that $\mathsf{H^{qk}}$ hasn't been decided on yet.

It wouldn't take much work to specify. I am strongly inclined to make $\mathsf{H^{qk}}$ a (domain-separated) use of BLAKE3, rather than being an algebraic hash, because the latter would make the circuit field-dependent. BLAKE3 is good enough for this purpose. However, I am loath to do cryptographic design on the fly and potentially end up with funds sent to keys that can't be supported in the long term. So if we add this now, it will have to be on the basis that it is purely temporary and we WILL remove this constructor.

```suggestion
    /// Create a `FullViewingKey` intended *ONLY FOR EXPERIMENTAL USE* with
    /// FROST, and that *DOES NOT SUPPORT QUANTUM RECOVERABILITY*, from
    /// the given `SpendingKey` `sk` and the given `SpendValidatingKey` `ak`.
    ///
    /// When using FROST or any threshold signature scheme, we might need to
    /// create a `FullViewingKey` from an specific `SpendValidatingKey` which was
    /// not derived from a `SpendingKey` (since it might be generated using a
    /// Distributed Key Generation protocol, which generates a random
    /// `SpendValidatingKey`). This method achieves this by creating the
    /// `FullViewingKey` from the `SpendingKey` as usual, with the exception of
    /// `SpendValidatingKey` which is provided by the caller.
    ///
    /// # Why this method of constructing FVKs will not be supported in the long term
    ///
    /// See [ZIP 2005](https://zips.z.cash/zip-2005) for the motivation behind
    /// Quantum Recoverability. A Quantum Recovery protocol needs to be able to
    /// check that $\mathsf{ivk}$ commits to the full viewing key $(\mathsf{ak}, \mathsf{nk})$.
    /// Because the commitment scheme used in Orchard to derive the FVK is not
    /// post-quantum binding, the Recovery circuit has to check (in zk) how the ivk
    /// was constructed, which requires a known key derivation to be used. This
    /// constructor bypasses the normal key derivation for non-FROST keys, and
    /// the key derivation for FROST keys is yet to be fully specified (the choice of
    /// $\mathsf{H^{qk}}$ has not been made).
    ///
    /// The impact of this is not limited to funds sent to the corresponding address
    /// being quantum-vulnerable. Because FVKs generated by this constructor will
    /// not conform to future mandatory conformance requirements, future wallets
    /// might not support them at all, possibly resulting in the funds becoming
    /// inaccessible. That is why this constructor is only intended for experimental
    /// usage, and SHOULD NOT be exposed in public-facing wallet software without
    /// strong warnings to the user.
```

nuttycom:
We've decided not to include this in 0.13; as it's additive, we can add what's needed in a patch release once the issues that @daira identifies are resolved.

daira (changes requested):
NACK, see comment.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Added
+
+- `orchard::keys::FullViewingKey::from_sk_ak()` behind the `unstable-frost` flag.
```
daira:
```suggestion
- `orchard::keys::FullViewingKey::from_sk_ak_incompatible_with_quantum_recoverability_and_will_be_removed()`,
  behind the `unstable-frost` flag. Intended *ONLY FOR EXPERIMENTAL USE* with
  FROST; see the doc comment for further caveats.
```

PR #479, opened 2026-04-02
Return DepthOverflow error instead of panicking at depth 255

MavenRain:


nuttycom (approved):
utACK 44de07dab1a41a9728a4b5bd57983feee65583da

nullcopy (approved):
utACK 44de07dab1a41a9728a4b5bd57983feee65583da

PR #482, opened 2026-04-17
Migrate from yanked `core2` library to `corez`

### LINKED ISSUE #484: Migrate from yanked `core2` crate to `corez`

nuttycom:


nullcopy (dismissed):
utACK e94ade8dd1e4dff0538555152fd20bead14cf0b0 modulo rustfmt errors

nullcopy (approved):
utACK 872a2192238f4085ce4c8dd2f0222bd6ce26fba7

daira (approved):
utACK

Cargo.toml
```diff
@@ -59,7 +59,7 @@ getset = "0.1"
 tracing = { version = "0.1", default-features = false }
 
 # No-std support
-core2 = { version = "0.3", default-features = false, features = ["alloc"] }
+corez = { version = "0.1.1", default-features = false, features = ["alloc"] }
```
czarcas7ic, line 62: I checked to see if corez 0.1.1 was a 1:1 match of the yanked core2 0.3, but it seems to be a slimmed down version. I checked the diff and the things that were removed appear safe to me.

nuttycom: Correct; in particular, some code paths that could result in UB were removed; this is the minimal API surface that our crates actually need.

PR #486, opened 2026-04-17
Allow the identity for `rk` in halo2 proof instance construction.

nuttycom:
According to the `orchard` specification, the identity is a valid value for `rk`; it was inadvertently disallowed as a consequence of an API design that did not readily allow decomposition into affine coordinates for the identity.


nuttycom:
Closing in favor of #487

PR #487, opened 2026-04-17
Disallow identity `rk` in Orchard proof verification.

nuttycom:
Previously, proof verification could crash if a set of public inputs to the circuit contained an `rk` value that mapped to the identity.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,12 +60,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -271,16 +265,15 @@ dependencies = [
 
 [[package]]
 name = "chrono"
-version = "0.4.30"
+version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "defd4e7873dbddba6c7c91e199c7fcb946abc4a6a4ac3195400bcfb01b5de877"
+checksum = "c673075a2e0e5f4a1dde27ce9dee1ea4558c7ffe648f576438a20ca1d2acc4b0"
 dependencies = [
- "android-tzdata",
  "iana-time-zone",
  "js-sys",
  "num-traits",
  "wasm-bindgen",
- "windows-targets",
+ "windows-link",
 ]
 
 [[package]]
@@ -343,25 +336,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "cmake"
-version = "0.1.50"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a31c789563b815f77f4250caee12365734369f942439b7defd71e18a48197130"
-dependencies = [
- "cc",
-]
-
-[[package]]
 name = "color_quant"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3d7b894f5411737b7867f4827955924d7c254fc9f4d91a6aad6b097804b1018b"
-
-[[package]]
-name = "const-cstr"
-version = "0.3.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ed3d0b5ff30645a68f35ece8cea4556ca14ef8a1651455f789a099a0513532a6"
 
 [[package]]
 name = "constant_time_eq"
@@ -371,9 +349,9 @@ checksum = "21a53c0a4d288377e7415b53dcfc3c04da5cdc2cc95c8d5ac178b58f0b861ad6"
 
 [[package]]
 name = "core-foundation"
-version = "0.9.3"
+version = "0.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "194a7a9e6de53fa55116934067c844d9d749312f75c6f6d0980e8c252f8c2146"
+checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
 dependencies = [
  "core-foundation-sys",
  "libc",
@@ -381,15 +359,15 @@ dependencies = [
 
 [[package]]
 name = "core-foundation-sys"
-version = "0.8.4"
+version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e496a50fda8aacccc86d7529e2c1e0892dbd0f898a6b5645b5561b89c3210efa"
+checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
 name = "core-graphics"
-version = "0.22.3"
+version = "0.23.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2581bbab3b8ffc6fcbd550bf46c355135d16e9ff2a6ea032ad6b9bf1d7efe4fb"
+checksum = "c07782be35f9e1140080c6b96f0d44b739e2278479f64e02fdab4e32dfd8b081"
 dependencies = [
  "bitflags 1.3.2",
  "core-foundation",
@@ -400,9 +378,9 @@ dependencies = [
 
 [[package]]
 name = "core-graphics-types"
-version = "0.1.2"
+version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2bb142d41022986c1d8ff29103a1411c8a3dfad3552f87a4f8dc50d61d4f4e33"
+checksum = "45390e6114f68f718cc7a830514a96f903cccd70d02a8f6d9f643ac4ba45afaf"
 dependencies = [
  "bitflags 1.3.2",
  "core-foundation",
@@ -411,9 +389,9 @@ dependencies = [
 
 [[package]]
 name = "core-text"
-version = "19.2.0"
+version = "20.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "99d74ada66e07c1cefa18f8abfba765b486f250de2e4a999e5727fc0dd4b4a25"
+checksum = "c9d2790b5c08465d49f8dc05c8bcae9fea467855947db39b0f8145c091aaced5"
 dependencies = [
  "core-foundation",
  "core-graphics",
@@ -632,24 +610,24 @@ dependencies = [
 ]
 
 [[package]]
-name = "dirs-next"
-version = "2.0.0"
+name = "dirs"
+version = "6.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b98cf8ebf19c3d1b223e151f99a4f9f0690dca41414773390fc824184ac833e1"
+checksum = "c3e8aa94d75141228480295a7d0e7feb620b1a5ad9f12bc40be62411e38cce4e"
 dependencies = [
- "cfg-if",
- "dirs-sys-next",
+ "dirs-sys",
 ]
 
 [[package]]
-name = "dirs-sys-next"
-version = "0.1.2"
+name = "dirs-sys"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4ebda144c4fe02d1f7ea1a7d9641b6fc6b580adcfa024ae48797ecdeb6825b4d"
+checksum = "e01a3366d27ee9890022452ee61b2b63a67e6f13f58900b651ff5665f0bb1fab"
 dependencies = [
  "libc",
+ "option-ext",
  "redox_users",
- "winapi",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -663,9 +641,9 @@ dependencies = [
 
 [[package]]
 name = "dwrote"
-version = "0.11.0"
+version = "0.11.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "439a1c2ba5611ad3ed731280541d36d2e9c4ac5e7fb818a27b604bdc5a6aa65b"
+checksum = "9e1b35532432acc8b19ceed096e35dfa088d3ea037fe4f3c085f1f97f33b4d02"
 dependencies = [
  "lazy_static",
  "libc",
@@ -687,7 +665,7 @@ checksum = "136526188508e25c6fef639d7927dfb3e0e3084488bf202267829cf7fc23dbdd"
 dependencies = [
  "errno-dragonfly",
  "libc",
- "windows-sys",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -766,9 +744,9 @@ dependencies = [
 
 [[package]]
 name = "float-ord"
-version = "0.2.0"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7bad48618fdb549078c333a7a8528acb57af271d0433bdecd523eb620628364e"
+checksum = "8ce81f49ae8a0482e4c55ea62ebbd7e5a686af544c00b9d090bba3ff9be97b3d"
 
 [[package]]
 name = "flume"
@@ -791,19 +769,19 @@ checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
 name = "font-kit"
-version = "0.11.0"
+version = "0.14.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "21fe28504d371085fae9ac7a3450f0b289ab71e07c8e57baa3fb68b9e57d6ce5"
+checksum = "2c7e611d49285d4c4b2e1727b72cf05353558885cc5252f93707b845dfcaf3d3"
 dependencies = [
- "bitflags 1.3.2",
+ "bitflags 2.4.0",
  "byteorder",
  "core-foundation",
  "core-graphics",
  "core-text",
- "dirs-next",
+ "dirs",
  "dwrote",
  "float-ord",
- "freetype",
+ "freetype-sys",
  "lazy_static",
  "libc",
  "log",
@@ -816,18 +794,30 @@ dependencies = [
 
 [[package]]
 name = "foreign-types"
-version = "0.3.2"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f6f339eb8adc052cd2ca78910fda869aefa38d22d5cb648e6485e4d3fc06f3b1"
+checksum = "d737d9aa519fb7b749cbc3b962edcf310a8dd1f4b67c91c4f83975dbdd17d965"
 dependencies = [
+ "foreign-types-macros",
  "foreign-types-shared",
 ]
 
 [[package]]
-name = "foreign-types-shared"
-version = "0.1.1"
+name = "foreign-types-macros"
+version = "0.2.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00b0228411908ca8685dba7fc2cdd70ec9990a6e753e89b6ac91a84c40fbaf4b"
+checksum = "1a5c6c585bc94aaf2c7b51dd4c2ba22680844aba4c687be581871a6f518c5742"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
+name = "foreign-types-shared"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aa9a19cbb55df58761df49b23516a86d432839add4af60fc256da840f66ed35b"
 
 [[package]]
 name = "fpe"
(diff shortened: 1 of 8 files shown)


ValarDragon (commented):
Lgtm 

Pr added changing an error from unwrap to passing it up

src/circuit.rs
```diff
@@ -835,7 +835,7 @@ impl Instance {
         }
     }
 
-    fn to_halo2_instance(&self) -> [[vesta::Scalar; 9]; 1] {
+    fn to_halo2_instance(&self) -> Option<[[vesta::Scalar; 9]; 1]> {
```
daira: Why make this return an `Option` rather than prevent the creation of invalid `Instance`s?

Two options: exclude the identity from `VerificationKey<SpendAuth>`, or make at least `rk` private with an accessor. I'd be inclined to do the latter, with all fields private and a `from_parts` that returns `Option<Self>`. Only field accesses in this crate and callers of `from_parts` need to change.

src/circuit.rs
```diff
@@ -906,18 +914,27 @@ impl Proof {
     /// just want to batch-validate Orchard bundles, use [`bundle::BatchValidator`].
     ///
     /// [`bundle::BatchValidator`]: crate::bundle::BatchValidator
-    pub fn add_to_batch(&self, batch: &mut BatchVerifier<vesta::Affine>, instances: Vec<Instance>) {
+    pub fn add_to_batch(
+        &self,
+        batch: &mut BatchVerifier<vesta::Affine>,
+        instances: Vec<Instance>,
```
daira: Note that the approach with `Instance`s valid-by-construction moves the public API change from this method to `Instance::from_parts`.

daira (dismissed):
utACK. I'm okay with either this approach or the `Instance` valid-by-construction one, your choice.

daira:
I think we had consensus not to use this approach (before the [force-push](https://github.com/zcash/orchard/compare/288cb064620ab9ecd0a88c1a266c2e51efdd343a..211663c7c3d050d1c1451465edea874bc9db9899)), and instead to change `redpallas::VerificationKey<SpendAuth>` so that it disallows the identity. ~~(Note that the Pallas version of `VerificationKey` is defined in `redpallas`, not in `orchard`.)~~

There is a wrinkle with the randomization: it makes randomization partial, because if $\alpha = -\mathsf{ask}$ then $\mathsf{rk}$ will randomize to the identity. However, that's ok because the probability of it happening with honestly chosen $\alpha$ is negligible, and it's actually fine for `redpallas::VerificationKey<SpendAuth>::randomize` to be partial.

A possibly more significant reason to have pause is that this approach diverges from Sapling, which allows `redjubjub::VerificationKey<SpendAuth>` to be the identity (more generally a small-order point), and checks for small-order points in `SaplingVerificationContext::check_spend()`.

reddsa-0.5.1 src/verification_key.rs:93-116:
```rust
impl<T: SigType> TryFrom<VerificationKeyBytes<T>> for VerificationKey<T> {
    type Error = Error;

    fn try_from(bytes: VerificationKeyBytes<T>) -> Result<Self, Self::Error> {
        // XXX-jubjub: this should not use CtOption
        // XXX-jubjub: this takes ownership of bytes, while Fr doesn't.
        // This checks that the encoding is canonical...
        let mut repr = <T::Point as GroupEncoding>::Repr::default();
        repr.as_mut().copy_from_slice(&bytes.bytes);
        let maybe_point = T::Point::from_bytes(&repr);
        if maybe_point.is_some().into() {
            let point = maybe_point.unwrap();
            // Note that small-order verification keys (including the identity) are not
            // rejected here. Previously they were rejected, but this was a bug as the
            // RedDSA specification allows them. Zcash Sapling rejects small-order points
            // for the RedJubjub spend authorization key rk; this now occurs separately.
            // Meanwhile, Zcash Orchard uses a prime-order group, so the only small-order
            // point would be the identity, which is allowed in Orchard.
            Ok(VerificationKey { point, bytes })
        } else {
            Err(Error::MalformedVerificationKey)
        }
    }
}
```

nuttycom:
> I think we had consensus not to use this approach, and instead to change `redpallas::VerificationKey<SpendAuth>` so that it disallows the identity. (Note that the Pallas version of `VerificationKey` is defined in `redpallas`, not in `orchard`.)

Why would we change this in `redpallas` instead of just changing it in `orchard`?

daira (changes requested):
This fixes the consensus issue but it leaves a footgun in the form of the `randomize` method, which can produce an identity point (and that is easy for the tx creator to provoke, which might be a problem for hardware wallets for example). The `randomize` method should be partial; this is a breaking API change but there are not many callers, and it is correct to require them to handle the partiality.

Also, there should be tests.

src/builder.rs
```diff
@@ -302,7 +302,9 @@ impl SpendInfo {
         let nf_old = self.note.nullifier(&self.fvk);
         let ak: SpendValidatingKey = self.fvk.clone().into();
         let alpha = pallas::Scalar::random(&mut rng);
-        let rk = ak.randomize(&alpha);
+        let rk = ak.randomize(&alpha).expect(
+            "alpha was generated randomly; randomization producing the identity is cryptographically negligible"
```
daira:
```suggestion
            "alpha was generated randomly; randomization producing the identity is vanishingly unlikely"
```

src/builder.rs
```diff
@@ -1068,7 +1070,7 @@ impl<P: fmt::Debug, V> Bundle<InProgress<P, PartiallyAuthorized>, V> {
             &mut signature_valid_for,
             |valid_for, partial, maybe| match maybe {
                 MaybeSigned::SigningMetadata(parts) => {
-                    let rk = parts.ak.randomize(&parts.alpha);
+                    let rk = parts.ak.randomize(&parts.alpha).expect("SigningParts is constructed only with random alpha; the randomized rk being the identity is cryptographically negligible.");
```
daira: Document on `SigningParts` that it always has a random alpha.

src/builder.rs
```diff
@@ -921,6 +923,8 @@ pub struct SigningParts {
     /// actions they can create signatures for.
     ak: SpendValidatingKey,
     /// The randomization needed to derive the actual signing key for this note.
+    ///
+    /// This is guaranteed (internal to the `orchard` crate) to be derived from randomness.
```
ebfull, line 927:
```suggestion
    /// This is guaranteed (internally) to be derived from randomness.
```

src/primitives/redpallas.rs
```diff
@@ -122,8 +137,16 @@ impl VerificationKey<SpendAuth> {
     /// Randomizes this verification key with the given `randomizer`.
     ///
     /// Randomization is only supported for `SpendAuth` keys.
-    pub fn randomize(&self, randomizer: &pallas::Scalar) -> Self {
-        VerificationKey(self.0.randomize(randomizer))
+    ///
+    /// Returns the randomized key, or an error if the [`pallas::Point`] corresponding to the
+    /// randomized verification key is the identity.
+    pub fn randomize(&self, randomizer: &pallas::Scalar) -> Result<Self, reddsa::Error> {
```
nuttycom: Design question: since we're returning `Option` for `SpendValidatingKey::randomize`, should we also return `Option` here?

src/primitives/redpallas.rs
```diff
@@ -78,11 +79,25 @@ impl<T: SigType> From<&VerificationKey<T>> for [u8; 32] {
...
 
-impl<T: SigType> TryFrom<[u8; 32]> for VerificationKey<T> {
+impl TryFrom<[u8; 32]> for VerificationKey<SpendAuth> {
     type Error = reddsa::Error;
 
     fn try_from(bytes: [u8; 32]) -> Result<Self, Self::Error> {
-        bytes.try_into().map(VerificationKey)
+        // Do not permit construction of `VerificationKey<SpendAuth>` if the underlying
+        // `pallas::Point` is the identity.
+        if bool::from(pallas::Point::from_bytes(&bytes).is_none()) {
```
daira: This is not a correct way to check for the identity.

[Claude]
In [`pasta_curves-0.5.1/src/curves.rs` lines 663–686](https://github.com/zcash/pasta_curves/blob/7e3fc6a4919f6462a32b79dd226cb2587b7961eb/src/curves.rs#L663-L686), the affine `from_bytes` explicitly decodes the all-zeros encoding as the identity:

```rust
fn from_bytes(bytes: &[u8; 32]) -> CtOption<Self> {
    let mut tmp = *bytes;
    let ysign = Choice::from(tmp[31] >> 7);
    tmp[31] &= 0b0111_1111;

    $base::from_repr(tmp).and_then(|x| {
        CtOption::new(Self::identity(), x.is_zero() & (!ysign)).or_else(|| {
            // ... non-identity path
        })
    })
}
```

For `bytes = [0u8; 32]`:

- `ysign = Choice::from(0)`
- `x = 0` (decodes cleanly via `from_repr`)
- `x.is_zero() & !ysign` evaluates to `true`
- The function returns `CtOption::new(Self::identity(), Choice::from(1u8))`, i.e. `is_some() == true`.

The projective `Point::from_bytes` just forwards to the affine version (curves.rs:258–260), so the behavior is identical for `pallas::Point`.

src/primitives/redpallas.rs
```diff
@@ -122,8 +137,16 @@ impl VerificationKey<SpendAuth> {
...
     ///
     /// Randomization is only supported for `SpendAuth` keys.
-    pub fn randomize(&self, randomizer: &pallas::Scalar) -> Self {
-        VerificationKey(self.0.randomize(randomizer))
+    ///
+    /// Returns the randomized key, or an error if the [`pallas::Point`] corresponding to the
+    /// randomized verification key is the identity.
+    pub fn randomize(&self, randomizer: &pallas::Scalar) -> Result<Self, reddsa::Error> {
+        let randomized = self.0.randomize(randomizer);
+        if bool::from(pallas::Point::from_bytes(&<[u8; 32]>::from(randomized)).is_none()) {
```
daira: This has the same problem as above.

daira (changes requested):
NACK; see https://github.com/zcash/orchard/pull/487/changes#r3119617155

nuttycom:
[force-pushed](https://github.com/zcash/orchard/compare/b0669cda44fd16ff9cc9d1562165b0b88a8a9e0d..3c583e33d5e40f2710b56fa536be04a8693b6370) to address comments from review.

src/primitives/redpallas.rs
```diff
@@ -78,11 +79,28 @@ impl<T: SigType> From<&VerificationKey<T>> for [u8; 32] {
...
     type Error = reddsa::Error;
 
     fn try_from(bytes: [u8; 32]) -> Result<Self, Self::Error> {
-        bytes.try_into().map(VerificationKey)
+        // Do not permit construction of `VerificationKey<SpendAuth>` if the underlying
+        // `pallas::Point` is the identity. Note that `pallas::Point::from_bytes` decodes
+        // the all-zeros encoding successfully (as the identity), so we must explicitly
+        // check `is_identity()` rather than relying on the decoding to fail.
+        let maybe_point = pallas::Point::from_bytes(&bytes);
+        if bool::from(maybe_point.is_none()) || bool::from(maybe_point.unwrap().is_identity()) {
```
daira:
```suggestion
        if bool::from(maybe_point.unwrap_or(pallas::Point::identity()).is_identity()) {
```
would be constant-time except in the error case, although I'm not sure that matters here.

src/primitives/redpallas.rs
```diff
@@ -122,8 +140,17 @@ impl VerificationKey<SpendAuth> {
...
     /// Randomization is only supported for `SpendAuth` keys.
-    pub fn randomize(&self, randomizer: &pallas::Scalar) -> Self {
-        VerificationKey(self.0.randomize(randomizer))
+    ///
+    /// Returns `None` if the [`pallas::Point`] corresponding to the randomized verification
+    /// key is the identity.
+    pub fn randomize(&self, randomizer: &pallas::Scalar) -> Option<Self> {
+        let randomized = self.0.randomize(randomizer);
+        let maybe_point = pallas::Point::from_bytes(&<[u8; 32]>::from(randomized));
+        if bool::from(maybe_point.is_none()) || bool::from(maybe_point.unwrap().is_identity()) {
```
daira:
```suggestion
        if bool::from(maybe_point.unwrap_or(pallas::Point::identity()).is_identity()) {
```

src/builder.rs
```diff
@@ -302,7 +302,9 @@ impl SpendInfo {
         let nf_old = self.note.nullifier(&self.fvk);
         let ak: SpendValidatingKey = self.fvk.clone().into();
         let alpha = pallas::Scalar::random(&mut rng);
-        let rk = ak.randomize(&alpha);
+        let rk = ak.randomize(&alpha).expect(
+            "alpha was generated randomly; randomization producing the identity is vanishingly unlikely"
```
daira:
```suggestion
            "alpha was generated randomly, so identity rk is vanishingly unlikely"
```
Non-blocking, consistency/conciseness.

src/builder.rs
```diff
@@ -1068,7 +1072,7 @@ impl<P: fmt::Debug, V> Bundle<InProgress<P, PartiallyAuthorized>, V> {
             &mut signature_valid_for,
             |valid_for, partial, maybe| match maybe {
                 MaybeSigned::SigningMetadata(parts) => {
-                    let rk = parts.ak.randomize(&parts.alpha);
+                    let rk = parts.ak.randomize(&parts.alpha).expect("SigningParts is constructed only with random alpha; the randomized rk being the identity is vanishingly unlikely.");
```
daira:
```suggestion
                    let rk = parts.ak.randomize(&parts.alpha).expect("SigningParts is constructed only with random alpha, so identity rk is vanishingly unlikely.");
```
Non-blocking, consistency/conciseness.

src/circuit.rs
```diff
@@ -949,7 +947,9 @@ mod tests {
         let rho = Rho::from_nf_old(nf_old);
         let ak: SpendValidatingKey = fvk.into();
         let alpha = pallas::Scalar::random(&mut rng);
-        let rk = ak.randomize(&alpha);
+        let rk = ak
+            .randomize(&alpha)
+            .expect("randomization does not yield the identity for `rk`");
```
daira:
```suggestion
            .expect("alpha was generated randomly, so identity rk is vanishingly unlikely");
```
Blocking because the current expect message is arguably incorrect (or if "does not" is interpreted more loosely, it doesn't say why not).

src/primitives/redpallas.rs
```diff
@@ -226,3 +253,20 @@ pub mod testing {
...
+        // The all-zeros encoding is the canonical encoding of the identity point on Pallas.
+        // `VerificationKey<SpendAuth>` must reject it, since an identity `rk` would allow
+        // forgery of spend authorization signatures.
+        let result = VerificationKey::<SpendAuth>::try_from([0u8; 32]);
+        assert!(
+            result.is_err(),
+            "VerificationKey::<SpendAuth>::try_from([0u8; 32]) must reject the identity",
+        );
+    }
+}
```
daira, line 375: Add a test of randomization that produces the identity (use the relevant base point with -1 as the randomizer), for both `SpendAuth` and `Binding`. The former should return `None` and the latter should return the identity public key.

daira (dismissed):
The requested tests are important, otherwise utACK with suggestions.

nuttycom:
[force-pushed](https://github.com/zcash/orchard/compare/9f128439392371415930dfa251f2a8578d0ff1e6..a5acdd2b6d4335524fb6f9e0dad2b8e2c5bb9349) to address additional review comments.

Cargo.lock
```diff
@@ -1593,9 +1578,9 @@ checksum = "26072860ba924cbfa98ea39c8c19b4dd6a4a25423dbdf219c1eca91aa0cf6964"
 
 [[package]]
 name = "plotters"
-version = "0.3.5"
+version = "0.3.7"
```
ebfull, line 1581: I was able to confirm there wasn't a less aggressive bump, though the impact is minimal since it is just a lockfile in this repository.

src/primitives/redpallas.rs
```diff
@@ -122,8 +144,21 @@ impl VerificationKey<SpendAuth> {
...
-        VerificationKey(self.0.randomize(randomizer))
+    ///
+    /// Returns `None` if the [`pallas::Point`] corresponding to the randomized verification
+    /// key is the identity.
+    pub fn randomize(&self, randomizer: &pallas::Scalar) -> Option<Self> {
+        let randomized = self.0.randomize(randomizer);
+        let maybe_point = pallas::Point::from_bytes(&<[u8; 32]>::from(randomized));
+        if bool::from(
+            maybe_point
+                .unwrap_or(pallas::Point::identity())
```
ebfull: Shouldn't this be `.expect("VerificationKey bytes are always a canonical Pallas point encoding")`? `from_bytes` should not fail, since we just encoded it. (This was probably copied from the `try_from` logic above, where the bytes may or may not be a valid encoding.)

ebfull: `CtOption` doesn't have an `expect` method so you'll have to change this function:

```rust
pub fn randomize(&self, randomizer: &pallas::Scalar) -> Option<Self> {
    let randomized = self.0.randomize(randomizer);
    let bytes = <[u8; 32]>::from(randomized);
    let point = Option::from(pallas::Point::from_bytes(&bytes))
        .expect("VerificationKey bytes are always a canonical Pallas point encoding");

    if bool::from(point.is_identity()) {
        None
    } else {
        Some(VerificationKey(randomized))
    }
}
````

Comment (nuttycom) [reply src/primitives/redpallas.rs]: The `self.0.randomize(randomizer)` call produces a `reddsa::VerificationKey`, which *does* permit the identity. That API doesn't give us any way to obtain the point except via the byte encoding, so we're plumbing through bytes, then checking. Since `reddsa::VerificationKey` admits the identity, we can't `expect` here.

Comment (ebfull) [reply src/primitives/redpallas.rs]: That is conflating two failure modes; the existing `unwrap_or` is covering an unreachable condition (a false encoding, regardless of whether it's the identity or not, will not be produced by `<[u8; 32]>::from(randomized)`.)

Comment (daira) [reply src/primitives/redpallas.rs]: @ebfull's suggestion looks correct and well-motivated to me (and Claude).

Comment (nuttycom) [reply src/primitives/redpallas.rs]: Oh, never mind, I get it - the Pallas decoding won't fail because it was an encoded valid Pallas point, we just need to subsequently check it against the identity. Misread, sorry.

File: src/primitives/redpallas.rs [hunk 17]
```diff
@@ -226,3 +260,67 @@ pub mod testing {
...
+
+    #[test]
+    fn orchard_spendauth_basepoint() {
+        use group::GroupEncoding;
+        use pasta_curves::arithmetic::CurveExt;
+        assert_eq!(
+            pallas::Point::hash_to_curve("z.cash:Orchard")(b"G").to_bytes(),
+            ORCHARD_SPENDAUTHSIG_BASEPOINT_BYTES,
+        );
+    }
```
daira:
```suggestion
    use super::{Binding, SpendAuth, SigningKey, VerificationKey};
```
Simpler: we can generate a `VerificationKey` for the base point using `SigningKey` (it is the `VerificationKey` corresponding to a `SigningKey` with scalar 1).

src/primitives/redpallas.rs
```diff
@@ -226,3 +260,67 @@ pub mod testing {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use group::ff::Field;
```
daira:
```suggestion
    use group::ff::{Field, PrimeField};
```

src/primitives/redpallas.rs
```diff
@@ -226,3 +260,67 @@ pub mod testing {
...
+        );
+    }
+
+    #[test]
+    fn spendauth_randomize_to_identity_returns_none() {
+        // Construct a `VerificationKey<SpendAuth>` equal to the SpendAuthSig basepoint
+        // `G`, then randomize by `alpha = -1`. The resulting key is
+        // `rk = G + (-1) * G = identity`, which must be rejected.
+        let ak = VerificationKey::<SpendAuth>::try_from(ORCHARD_SPENDAUTHSIG_BASEPOINT_BYTES)
+            .expect("the SpendAuthSig basepoint is not the identity");
```
daira:
```suggestion
        // Construct a `VerificationKey<SpendAuth>` corresponding to the
        // signing key with scalar 1 (so, equal to the `SpendAuthSig` base
        // point `G`), then randomize by `alpha = -1`. The resulting key is
        // `rk = [1] G + [-1] G = identity`, which must be rejected.
        let ask_bytes: [u8; 32] = pallas::Scalar::ONE.to_repr().into();
        let ask: SigningKey<SpendAuth> = ask_bytes.try_into().expect("1 is a valid scalar");
        let ak: VerificationKey<SpendAuth> = ask.into();
```

src/primitives/redpallas.rs
```diff
@@ -226,3 +260,52 @@ pub mod testing {
...
+        // point `G`), then randomize by `alpha = -1`. The resulting key is
+        // `rk = [1] G + [-1] G = identity`, which must be rejected.
+        let ask_bytes: [u8; 32] = pallas::Scalar::ONE.to_repr().into();
+        let ask = <SigningKey<SpendAuth>>::try_from(ask_bytes).expect("1 is a valid scalar");
+        let ak = <VerificationKey<SpendAuth>>::from(&ask);
+        let alpha = -pallas::Scalar::ONE;
+        assert!(
+            ak.randomize(&alpha).is_none(),
+            "randomizing the SpendAuthSig basepoint by -1 produces the identity, which must be rejected",
+        );
```
daira, line 373: Actually this exposes a problem. There are various backdoors in `reddsa`'s `src/verification_key.rs` allowing a `VerificationKey<T>` (for any `T: SigType`) to be constructed from a `SigningKey<T>` or a `T::Scalar`. And we have not prevented a `SigningKey<SpendAuth>` from having scalar 0. The method that constructs a `VerificationKey<T>` from a `T::Scalar` is `pub(crate)`, but I think that just using `ZERO` instead of `ONE` above would create a forbidden identity `VerificationKey<SpendAuth>`.

nuttycom: I don't understand the problem here; as far as I can tell, all of the methods that can produce a `redpallas::VerificationKey<SpendAuth>` are guarded, and there's no public function from `reddsa::VerificationKey -> redpallas::VerificationKey`, so what is the issue?

nuttycom: Ah, actually, I see it.

nuttycom: @daira does this mean that we need to check the signing key at construction? Also, as we discussed, I want to get rid of the `From` impl that converts `SigningKey` to `VerificationKey`; that shouldn't be a bound.

src/keys.rs
```diff
@@ -204,11 +209,10 @@ impl SpendValidatingKey {
...
             .ok()
             .and_then(|b| {
-                // Structural validity checks for ak_P:
-                // - The point must not be the identity
-                //   (which for Pallas is canonically encoded as all-zeroes).
-                // - The sign of the y-coordinate must be positive.
-                if b != [0; 32] && b[31] & 0x80 == 0 {
+                // Structural validity check for ak_P: the sign of the y-coordinate must
+                // be positive. Rejection of the identity point is handled by
+                // `<redpallas::VerificationKey<SpendAuth>>::try_from`.
```
daira, line 220: This change depends on an identity `VerificationKey<SpendAuth>` being unconstructable, but as we saw above, that isn't a valid assumption.

nuttycom:
Closing in favor of #492

PR #18, opened 2021-02-12
book: Update commitment tree section with the design decision

str4d:

PR #488, opened 2026-04-20
feat: add unstable-voting-circuits feature to widen internals

czarcas7ic:
## Summary

Adds an opt-in feature flag `unstable-voting-circuits` that surgically exposes the orchard internals needed to build a downstream voting circuit against this crate, following the same pattern as the existing `unstable-frost` flag.

All visibility changes only upgrade to `pub` under `--features unstable-voting-circuits`. The `unstable-` prefix signals that any items exposed under this flag are not covered by the crate's semver stability guarantees and may change in any future release.


src/keys.rs
```diff
@@ -121,7 +121,7 @@ pub struct SpendAuthorizingKey(redpallas::SigningKey<SpendAuth>);
 
 impl SpendAuthorizingKey {
     /// Derives ask from sk. Internal use only, does not enforce all constraints.
-    fn derive_inner(sk: &SpendingKey) -> pallas::Scalar {
+    pub fn derive_inner(sk: &SpendingKey) -> pallas::Scalar {
```
czarcas7ic, line 124: Flagging this used to be private, now public

ebfull (dismissed):

ebfull:
Note that this should be merged after we do the `0.13` release.

src/primitives/redpallas.rs
```diff
@@ -172,9 +172,11 @@ impl<T: SigType> From<&Signature<T>> for [u8; 64] {
     }
 }
 
-pub(crate) mod private {
+/// Sealed trait for signature types.
+pub mod private {
```
nuttycom: This is an unacceptable change, it breaks the sealed-ness of the above traits. Is it actually needed for the voting circuit?

nuttycom (changes requested):
NACK pending detailed discussion, breaking the sealed-ness of the `SigType` trait needs strong justification. Also, are you certain that all of this surface area actually needs to be exposed?

czarcas7ic:
@nuttycom I had a more confined version that only exposed the items that voting circuits needed, but then got guidance from @ebfull to convert anything with `pub(crate)` to `pub` instead. 

To move things forward, I can open a second PR that is only the items that we actually need for voting circuits if that is desired.

nuttycom:
> @nuttycom I had a more confined version that only exposed the items that voting circuits needed, but then got guidance from @ebfull to convert anything with `pub(crate)` to `pub` instead.
> 
> To move things forward, I can open a second PR that is only the items that we actually need for voting circuits if that is desired.

Yes, please. We can always expose more if needed, but it's more of a hassle to close things off.

czarcas7ic:
@nuttycom on it now, thanks!

ebfull:
@nuttycom The purpose of the crate split was to enable us to avoid having to make deliberate stable API decisions every time someone needs to repurpose an internal abstraction. As a general rule, anything that _was_ `pub(crate)` should be safe to expose from the `orchard_internal` crate, since we're using the `orchard` crate as a _stable_ guard. (I explicitly told @czarcas7ic that anything that was visibility-scoped to a module needed to have specific intention, but as a rule `pub(crate) -> pub` is fair game IMO.)

If we really needed to close down visibility afterward for some reason we could do it with a major (breaking) release of `orchard_internal`, which we should feel free to do whenever we want.

The specific change that was a sealed trait pattern that was turned into a `pub` is a mistake but not because of the PR author. The _existing_ code is wrong, because a sealed trait pattern should almost never be `pub(crate)`, it should be `pub` inside of a _private_ module.

ebfull (approved):

ebfull (approved):

src/primitives/redpallas.rs
czarcas7ic: Fixed in my squashed commit, thanks

czarcas7ic:
@nuttycom just for clarity, I do have the minimal change staged on our fork as promised (https://github.com/valargroup/orchard/pull/15), but it seems like the full widening might still be desired (aside from the error you flagged, which is now fixed)

src/primitives/redpallas.rs
```diff
@@ -172,7 +172,7 @@ impl<T: SigType> From<&Signature<T>> for [u8; 64] {
     }
 }
 
-pub(crate) mod private {
+mod private {
     use super::{Binding, SpendAuth};
 
     pub trait Sealed {}
```
ebfull, line 186: The earlier concern about the sealed trait pattern was resolved here by making the module private, which is what the existing code should have been doing.

nuttycom:
> @nuttycom The purpose of the crate split was to enable us to avoid having to make deliberate stable API decisions every time someone needs to repurpose an internal abstraction. As a general rule, anything that _was_ `pub(crate)` should be safe to expose from the `orchard_internal` crate, since we're using the `orchard` crate as a _stable_ guard. (I explicitly told @czarcas7ic that anything that was visibility-scoped to a module needed to have specific intention, but as a rule `pub(crate) -> pub` is fair game IMO.)
> 
> If we really needed to close down visibility afterward for some reason we could do it with a major (breaking) release of `orchard_internal`, which we should feel free to do whenever we want.
> 
> The specific change that was a sealed trait pattern that was turned into a `pub` is a mistake but not because of the PR author. The _existing_ code is wrong, because a sealed trait pattern should almost never be `pub(crate)`, it should be `pub` inside of a _private_ module.

I'm not sure that everything that was `pub(crate)` was well-justified in having that level of visibility; as you noted, the sealed trait shouldn't have had it. My thought was that it's trivially easy to make additive API changes, and so having just the exposure required for the voting circuit would be sufficient, and then we could make non-semver-breaking changes for whatever additions would be needed whenever they're actually needed, and that way we're not unintentionally exposing something where the `pub(crate)` was spurious. I don't feel super strongly about it though.

ebfull:
IMO anything that is `pub(crate)` and accessible from the crate root is by definition an internal API that isn't module-scoped, so should be part of the public API of the `orchard_internal` crate. (In a sense, it is already a "public API" from the perspective of the code within the `orchard_internal` crate.) Any examples where this is incorrect are only surfacing existing visibility bugs, and we should just fix those.

nuttycom:
> IMO anything that is `pub(crate)` and accessible from the crate root is by definition an internal API that isn't module-scoped, so should be part of the public API of the `orchard_internal` crate. (In a sense, it is already a "public API" from the perspective of the code within the `orchard_internal` crate.) Any examples where this is incorrect are only surfacing existing visibility bugs, and we should just fix those.

@czarcas7ic had had another PR up at one point that was the minimal set of changes required for the voting protocol; I had wanted to do a diff against that just to check whether that could help us find such visibility bugs, but I can't see that PR now, it looks like it was not just closed but was fully deleted?

src/constants/sinsemilla.rs
```diff
@@ -64,7 +64,7 @@ pub const Q_MERKLE_CRH: ([u8; 32], [u8; 32]) = (
     ],
 );
 
-pub(crate) fn lebs2ip_k(bits: &[bool]) -> u32 {
+pub fn lebs2ip_k(bits: &[bool]) -> u32 {
```
nuttycom: This is only used in this module, it could be private.

nuttycom: Also, if it's going to be public the panic needs to be documented.

src/constants/sinsemilla.rs
```diff
@@ -73,7 +73,7 @@ pub(crate) fn lebs2ip_k(bits: &[bool]) -> u32 {
 
 /// The sequence of K bits in little-endian order representing an integer
 /// up to `2^K` - 1.
-pub(crate) fn i2lebsp_k(int: usize) -> [bool; K] {
+pub fn i2lebsp_k(int: usize) -> [bool; K] {
```
nuttycom: This is used in orchard_primitives::tree, but if it's going to be made public the panic needs to be documented.

src/constants/util.rs
```diff
@@ -6,7 +6,7 @@ pub fn gen_const_array<Output: Copy + Default, const LEN: usize>(
     gen_const_array_with_default(Default::default(), closure)
 }
 
-pub(crate) fn gen_const_array_with_default<Output: Copy, const LEN: usize>(
+pub fn gen_const_array_with_default<Output: Copy, const LEN: usize>(
```
nuttycom: This is used only in this module, and could be private.

src/note/nullifier.rs
```diff
@@ -12,7 +12,7 @@ use crate::{
 
 /// A unique nullifier for a note.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub struct Nullifier(pub(crate) pallas::Base);
+pub struct Nullifier(pub pallas::Base);
```
nuttycom: The point of the wrapper here is to make Nullifiers safe by construction and immutable; this should never have had a pub(crate) field. Please make the internals here private and provide constructors and accessors.

src/primitives/redpallas.rs
```diff
@@ -115,7 +115,7 @@ impl<T: SigType> Ord for VerificationKey<T> {
 impl VerificationKey<SpendAuth> {
     /// Used in the note encryption tests.
     #[cfg(test)]
-    pub(crate) fn dummy() -> Self {
```
nuttycom, line 118: This is cfg(test), it doesn't need to change.

src/address.rs
```diff
@@ -23,7 +23,8 @@ pub struct Address {
 }
 
 impl Address {
-    pub(crate) fn from_parts(d: Diversifier, pk_d: DiversifiedTransmissionKey) -> Self {
+    /// Constructs an [`Address`] from a diversifier and a diversified transmission key.
+    pub fn from_parts(d: Diversifier, pk_d: DiversifiedTransmissionKey) -> Self {
```
nuttycom: These changes to the Address type are visible in the `orchard` public API; they need to be documented in the CHANGELOG if they're needed.

src/builder.rs
```diff
@@ -229,11 +229,16 @@ impl std::error::Error for OutputError {}
 /// Information about a specific note to be spent in an [`Action`].
 #[derive(Debug)]
 pub struct SpendInfo {
-    pub(crate) dummy_sk: Option<SpendingKey>,
```
nuttycom, line 232: These should never have been pub(crate) fields; also, these changes are visible in the orchard public API.

src/circuit.rs
```diff
@@ -101,25 +103,44 @@ pub struct Config {
 /// The Orchard Action circuit.
 #[derive(Clone, Debug, Default)]
 pub struct Circuit {
```
nuttycom, line 109: This is a huge expansion to the `orchard` public API; this change isn't confined to `orchard_internal`.

czarcas7ic:
> @czarcas7ic had had another PR up at one point that was the minimal set of changes required for the voting protocol; I had wanted to do a diff against that just to check whether that could help us find such visibility bugs, but I can't see that PR now, it looks like it was not just closed but was fully deleted?

@nuttycom that PR is here https://github.com/valargroup/orchard/pull/15, I opened it against our fork to reduce noise until we came to a conclusion here

ebfull (changes requested):
Decided to use `unstable-voting-circuits` and a targeted visibility export process.

czarcas7ic:
@nuttycom I appreciate the in depth review. As Sean mentioned above, I changed this PR to instead use a `unstable-voting-circuits` feature flag approach to both unblock the feature and keep things safe.

src/circuit.rs
```diff
@@ -58,9 +58,17 @@ use halo2_gadgets::{
...
 mod commit_ivk;
+#[cfg(feature = "unstable-voting-circuits")]
+#[allow(missing_docs)]
+pub mod commit_ivk;
 pub mod gadget;
+#[cfg(not(feature = "unstable-voting-circuits"))]
 mod note_commit;
+#[cfg(feature = "unstable-voting-circuits")]
+#[allow(missing_docs)]
+pub mod note_commit;
```
ebfull, line 69: Things like this are necessary because the `visibility` crate has some limitations for non-inline modules and tuple structs.

ebfull (changes requested):
Looks good, but this should update the `CHANGELOG.md` the same way that we've updated it before for the `unstable-frost` feature.

czarcas7ic:
Thanks @ebfull, just added the `CHANGELOG.md` entry

ebfull (dismissed):

src/circuit/gadget.rs
```diff
@@ -24,7 +24,11 @@ use halo2_proofs::{
     plonk::{self, Advice, Assigned, Column},
 };
 
+#[cfg(not(feature = "unstable-voting-circuits"))]
 pub(in crate::circuit) mod add_chip;
+/// Addition chip for constraining `a + b = c` in-circuit.
+#[cfg(feature = "unstable-voting-circuits")]
+pub mod add_chip;
```
daira, line 31: Does `#[cfg_attr(feature = "unstable-voting-circuits", visibility::make(pub))]` not work here?

ebfull: [It doesn't.](https://github.com/zcash/orchard/pull/488#discussion_r3127502547) I find this limitation extremely annoying; it almost defeats the purpose of the crate.

ebfull:
```
error[E0658]: non-inline modules in proc macro input are unstable
  --> src/lib.rs:44:1
   |
44 | pub(crate) mod spec;
   = note: see issue #54727 <https://github.com/rust-lang/rust/issues/54727>
```

ebfull: We might be able to get away with it using a `path` attribute on an inner module, and then glob importing the contents of the module out. But then that's infinitely worse than just using `cfg` attributes.

src/note/nullifier.rs
daira: I agree, we simply shouldn't have public wrappers like this. They are such an obstacle to fixing anything.

src/address.rs
```diff
@@ -35,10 +35,14 @@ impl Address {
         self.d
     }
 
+    /// Returns the diversified base point for this address.
+    #[cfg_attr(feature = "unstable-voting-circuits", visibility::make(pub))]
```
daira, line 39: How is this useful without also making `Address::from_parts` conditionally visible?

src/circuit/gadget.rs
nuttycom: This is fine as it is, it's obvious what it's doing.

src/note/nullifier.rs
```diff
@@ -11,9 +11,15 @@ use crate::{
...
 
 /// A unique nullifier for a note.
+#[cfg(not(feature = "unstable-voting-circuits"))]
 #[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
 pub struct Nullifier(pub(crate) pallas::Base);
 
+/// A unique nullifier for a note.
+#[cfg(feature = "unstable-voting-circuits")]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Nullifier(pub pallas::Base);
```
daira:
```suggestion
#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
pub struct Nullifier(pallas::Base);

impl Nullifier {
    /// Constructs a `Nullifier` from the given Pallas base field element.
    #[cfg_attr(feature = "unstable-voting-circuits", visibility::make(pub))]
    pub(crate) fn from_inner(inner: pallas::Base) -> Self {
        Self(inner)
    }

    /// Returns the inner Pallas base field element.
    #[cfg_attr(feature = "unstable-voting-circuits", visibility::make(pub))]
    pub(crate) fn inner(&self) -> pallas::Base {
        self.0
    }
}
```
Note that this will require some straightforward `.0` -> `.inner()` fix-ups elsewhere.

src/constants.rs
daira: These are arguably part of the documented protocol design, not an implementation detail, and can perfectly reasonably be public unconditionally.

(Yet again GitHub has messed up what lines the suggestion applies to. It's pissing me off because it slows everything down when suggestions, a really basic functionality of this webbed site, don't work.)

```rust
pub const L_ORCHARD_BASE: usize = 255;

/// $\ell^\mathsf{Orchard}_\mathsf{scalar}$
pub const L_ORCHARD_SCALAR: usize = 255;

/// $\ell_\mathsf{value}$
pub const L_VALUE: usize = 64;
```

src/note.rs
```diff
@@ -15,10 +15,21 @@ use crate::{
...
 
+#[cfg(not(feature = "unstable-voting-circuits"))]
 pub(crate) mod commitment;
+#[cfg(feature = "unstable-voting-circuits")]
+#[allow(missing_docs)]
+pub mod commitment;
+#[cfg(feature = "unstable-voting-circuits")]
+pub use self::commitment::{ExtractedNoteCommitment, NoteCommitTrapdoor, NoteCommitment};
+#[cfg(not(feature = "unstable-voting-circuits"))]
 pub use self::commitment::{ExtractedNoteCommitment, NoteCommitment};
```
daira, line 24:
```suggestion
pub use self::commitment::{ExtractedNoteCommitment, NoteCommitment};
#[cfg(feature = "unstable-voting-circuits")]
pub use self::commitment::NoteCommitTrapdoor;
```

nuttycom (dismissed):
utACK 1a1a05725cf17f39dcf31c75f60651a7914f0222; I'd prefer if a constructor were exposed for `Nullifier` instead of making the internals public but I won't block on it.

src/address.rs
czarcas7ic: voting-circuits never constructs an Address from its component fields. It only consumes addresses produced by already public APIs (FullViewingKey::address_at, Note::recipient) and then destructures them via g_d / pk_d for circuit witnessing. So exposing the read accessors without from_parts is sufficient for our needs. 

Example usage in our impl:

- https://github.com/valargroup/voting-circuits/blob/e60a19ae0e70d17c51c27daac95d8a49dcae05ac/voting-circuits/src/delegation/circuit.rs#L371-L386 (via note.recipient())
- https://github.com/valargroup/voting-circuits/blob/e60a19ae0e70d17c51c27daac95d8a49dcae05ac/voting-circuits/src/vote_proof/builder.rs#L464-L466 (via fvk.address_at(...))

That being said I am happy to make from_parts conditionally visible for completeness if you would like, but I am fairly sure our current implementation does not need it.

src/value.rs
```diff
@@ -100,6 +100,8 @@ impl std::error::Error for BalanceError {}
 pub struct NoteValue(u64);
 
 impl NoteValue {
+    /// Returns a zero note value.
+    #[cfg_attr(feature = "unstable-voting-circuits", visibility::make(pub))]
     pub(crate) fn zero() -> Self {
```
daira: This can be `pub`, it's equivalent to `NoteValue::from_raw(0)` which is already `pub`. (Alternatively, that means it doesn't actually need to be changed by this PR.)
```suggestion
    pub fn zero() -> Self {
```

daira (dismissed):
utACK with suggestions; like Kris I would strongly prefer the suggested change to `Nullifier`. Note that anything I suggested to make unconditionally `pub` must be documented as an addition.

czarcas7ic:
Some of the suggestions I accepted seems to have been reverted on my force push, I am re-applying them now. Will post here when ready for re-review. Thanks all!

src/value.rs
nuttycom: Instead of making this a pub method, I'd like to make this a pub constant ZERO (and remove the pub(crate) method).

src/constants.rs
czarcas7ic: Done, and also added `pub use constants::{L_ORCHARD_BASE, L_ORCHARD_SCALAR, L_VALUE};` at the crate root so they're reachable as `orchard::L_ORCHARD_BASE` etc. regardless of the `unstable-voting-circuits` feature.

daira:
I have some suggestions that I want to just push onto the branch, please hold off force-pushing for a bit.

Edit: ok for you to push now (I didn't have permission to push to this branch in any case), see https://github.com/valargroup/orchard/pull/18 for my suggested changes.

daira:
I've opened https://github.com/valargroup/orchard/pull/18 on to this PR, please review @nuttycom @czarcas7ic.

daira (commented):
## Review — current state (assuming doc PR #18 merged)

**Resolved since the first round of comments:**

- Sealed-trait break in `src/primitives/redpallas.rs` — fixed by the squashed commit; ebfull confirmed at line 186.
- `Nullifier` opaque-wrapper (Kris's and Daira-Emma's request) — applied.
- `L_ORCHARD_{BASE,SCALAR}` / `L_VALUE` unconditionally `pub` (Daira-Emma's suggestion) — applied, plus `pub use` at crate root.
- `src/note.rs` re-exports (Daira-Emma's suggestion) — applied.
- `NoteValue::zero` unconditionally `pub` (Daira-Emma's suggestion) — applied.
- `Address::{g_d, pk_d}` documented in CHANGELOG (Kris's request) — done.
- `#[allow(missing_docs)]` on the six newly-public modules — replaced with real docs in https://github.com/valargroup/orchard/pull/18 (pending merge into this branch).
- `src/primitives/redpallas.rs:118` (`cfg(test) fn dummy`) — current state matches Kris's "doesn't need to change."
- Daira-Emma's "Address::from_parts?" — reply clarified voting-circuits only reads, doesn't construct; acceptable.
- `src/circuit/gadget.rs:31` `visibility::make` attempt — ebfull linked the `non-inline modules in proc macro input` unstable-issue showing it can't work; Kris agreed "fine as it is."

[Removed false-positive discussion of clippy failures caused by version skew between my local clippy and CI's.]

**Still-live design discussion:**

Kris suggested a `pub const ZERO: NoteValue` in `src/value.rs` rather than `pub fn zero()`. The current code went with the `pub fn`. A `const` is marginally more idiomatic for a zero-argument invariant, and usable in const contexts; the `fn` matches `NoteValue::from_raw(0)`'s shape. Maintainer call.

**Minor cleanup (discretionary, pre-existing):**

- `gen_const_array` in `src/constants/util.rs` is `pub` but used only within the crate (`util.rs`, `spec.rs`). Kris flagged this. Not introduced by this PR, but since the PR is already in that file, a `pub(crate)` narrowing fits.
- Kris's three sinsemilla.rs concerns (items could be private / panic needs docs) are pre-existing; not introduced by this PR. Out of scope.
- SpendInfo `pub(crate)` fields in `src/builder.rs` — same story, pre-existing.

**Structural (Kris's main concern):**

"Huge expansion of public API" — still true in absolute terms (~30 items conditionally graduated to `pub` under the feature, mapped 1:1 against `valargroup/voting-circuits/UPSTREAM.md`'s checklist). Mitigations in place: feature flag, `unstable-` prefix, CHANGELOG flagging non-semver-stable. Each exposure justified by a concrete downstream consumer. Still worth being conservative about which items migrate from `unstable-voting-circuits` to unconditional `pub` in future releases.

🤖 This comment was prepared with the assistance of Claude Opus 4.7 (1M context).

czarcas7ic:
> I've opened https://github.com/valargroup/orchard/pull/18 on to this PR, please review @nuttycom @czarcas7ic.

Merged it in, thanks!

czarcas7ic:
I am now mobile but I will address Kris' concern at my soonest availability

daira (dismissed):
utACK

ValarDragon:
For the one remaining item:
> Kris suggested a pub const ZERO: NoteValue in src/value.rs rather than pub fn zero(). The current code went with the pub fn. A const is marginally more idiomatic for a zero-argument invariant, and usable in const contexts; the fn matches NoteValue::from_raw(0)'s shape. Maintainer call.

I'll just remove this as a change in this PR, and do that in an isolated follow-up

ebfull (approved):

PR #489, opened 2026-04-20
feat: add SpendAuthG fixed-base multiplication support

czarcas7ic:
## Summary

- Extends `orchard_internal::constants` under the `circuit` feature with `OrchardBaseFieldBases::SpendAuthGBase` and `OrchardShortScalarBases::SpendAuthGShort`
- Adds precomputed `Z_SHORT` / `U_SHORT` tables for SpendAuthG (generated by `find_zs_and_us`)
- Rewires internal callers (`value_commit_orchard`, `derive_nullifier`) with `.into()` adapters

Of the ~1,000 inserted lines, **~740 are machine-generated precomputed lookup tables** (byte arrays). The actual logic change is ~140 lines of enum definitions, trait routing, and tests.

### Why this matters

Without these variants, a downstream circuit that needs to multiply SpendAuthG by a base-field scalar or a short scalar must fall back to variable-base multiplication, which is significantly more expensive in constraint rows. Exposing SpendAuthG through the existing `FixedPointBaseField` and `FixedPointShort` gadgets follows the same pattern already established for `NullifierK` (base-field) and `ValueCommitV` (short-scalar).

### Verification

- Existing `src/circuit.rs::round_trip` still passes, so the pinned Orchard VK is unchanged
- New SpendAuthG fixed-base routing tests pass for both base-field and short-scalar paths
- New `Z_SHORT` / `U_SHORT` tables are validated with the existing `test_zs_and_us` helpers
- `cargo check -p orchard_internal --no-default-features` still succeeds


czarcas7ic:
This PR should only be merged after we do the `0.13` release.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Added
+- Under the `circuit` feature, `orchard_internal::constants::OrchardBaseFieldBases`
+  and `OrchardShortScalarBases` expose `SpendAuthG` as a fixed base for
+  multiplication by base-field and short signed scalars.
```
ebfull, line 13: This changelog is not tracking changes to the `orchard_internal` crate, or at least if it is, then #488 should also contain changelog entries for its API-affecting changes. I've addressed this in #490. Let's adapt this PR to the outcome of that discussion.

ebfull (changes requested):

CHANGELOG.md
czarcas7ic: This makes sense to me, addressed with the assumption that 490 gets merged as is. I will resolve this issue once 490 gets in, or follow up if 490 evolves from discussion.

ebfull (approved):
I've checked this thoroughly.

ebfull (changes requested):
Although this code piggybacks on much of the existing code, and I've checked it for correctness, _technically_ this PR currently is missing an end-to-end test for the newly supported bases. We should add those just to avoid leaving coverage gaps. (I know this is exercised downstream in `voting-circuits`.)

ebfull (approved):
Checked pretty thoroughly.

src/lib.rs
```diff
@@ -26,7 +26,8 @@ pub mod builder;
 pub mod bundle;
 #[cfg(feature = "circuit")]
 pub mod circuit;
-mod constants;
+#[allow(missing_docs)]
+pub mod constants;
```
ebfull: This visibility change takes place in #488 and will conflict once merged.

src/constants/fixed_bases.rs
```diff
@@ -52,13 +52,51 @@ pub const NUM_WINDOWS: usize = L_ORCHARD_SCALAR.div_ceil(FIXED_BASE_WINDOW_SIZE)
...
+/// `OrchardFixedBasesFull::SpendAuthG` (same 85-window structure over the
+/// 255-bit pallas base field), allowing `FixedPointBaseField::mul` to accept
+/// an `AssignedCell` directly — no variable-base `NonIdentityPoint` witness
+/// needed.
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
-// A sum type for both full-width and short bases. This enables us to use the
-// shared functionality of full-width and short fixed-base scalar multiplication.
-pub enum OrchardFixedBases {
-    Full(OrchardFixedBasesFull),
+pub enum OrchardBaseFieldBases {
```
ebfull, line 72: Let's make this `#[non_exhaustive]` right now while we're at it.

ebfull (changes requested):

src/constants/fixed_bases.rs
czarcas7ic: Just added on both `OrchardBaseFieldBases` and `OrchardShortScalarBases`

src/lib.rs
czarcas7ic: Thanks for flagging. Plan is to accept upstream change once #488 is merged. Lets keep this as unresolved until that happens.

daira:
I would like to review this before it goes in, since it's load-bearing for the security of the voting protocol. (It shouldn't take long.)

src/lib.rs
ValarDragon: Fixed in rebase

ValarDragon:
Rebased with #488 's merge

src/circuit/gadget.rs
```diff
@@ -130,7 +130,7 @@ pub(in crate::circuit) fn value_commit_orchard<
     // commitment = [v] ValueCommitV
     let (commitment, _) = {
         let value_commit_v = ValueCommitV;
-        let value_commit_v = FixedPointShort::from_inner(ecc_chip.clone(), value_commit_v);
```
ebfull: Not sure why this isn't inlined like the `NullifierK.into()` below in this file.

ebfull (dismissed):
This matches what I reviewed before, and it looked correct to me last time.

src/circuit/gadget.rs
ValarDragon: Yeah something odd from what was already in the crate. Fixed

ebfull (dismissed):

ebfull:
@daira There are a lot of things that are load-bearing for the security of the voting protocol. In this particular case, I've asked my engineer @TalDerei to additionally review the code. I will be doing a second review pass later today. That is more than adequate review for this particular change to the `orchard` crate.

src/constants.rs
```diff
@@ -7,6 +7,8 @@ pub mod util;
 pub use self::sinsemilla::{OrchardCommitDomains, OrchardHashDomains};
 #[cfg(feature = "circuit")]
 pub use fixed_bases::{NullifierK, OrchardFixedBases, OrchardFixedBasesFull, ValueCommitV};
+#[cfg(all(feature = "circuit", feature = "unstable-voting-circuits"))]
+pub use fixed_bases::{OrchardBaseFieldBases, OrchardShortScalarBases};
```
TalDerei, line 11: ACK – post-hoc traced (interstitial) visibility changes in https://github.com/zcash/orchard/pull/488 and proper downstream consumption in [voting-circuits](https://github.com/valargroup/voting-circuits).

src/constants/fixed_bases.rs
```diff
@@ -184,3 +277,362 @@ impl FixedPoint<pallas::Affine> for ValueCommitV {
...
+        /// `FixedPointShort::mul`.
+        Short,
+    }
+
+    /// Shared MockProver harness for both SpendAuthG fixed-base routes.
+    ///
+    /// Factoring this out keeps `configure`, the 10-bit lookup-table load, and
+    /// the MockProver wiring common between the per-base tests below. Only the
+    /// case-specific `mul` invocation and native-arithmetic cross-check differ.
+    fn run_spend_auth_g_fixed_base_mul_e2e(case: SpendAuthGCase) {
```
TalDerei, line 459: I locally extended these roundtrip tests to more exhaustively mirror the `halo2_gadgets/src/ecc/chip/mul_fixed/` battery, and it passes edge cases and random fuzzing.

src/constants/fixed_bases.rs
```diff
@@ -58,6 +58,45 @@ pub const NUM_WINDOWS: usize = L_ORCHARD_SCALAR.div_ceil(FIXED_BASE_WINDOW_SIZE)
...
+/// The ECC chip's `FixedPoints::Base` associated type must be a single type,
+/// so both `NullifierK` and `SpendAuthGBase` are wrapped in this enum.
+/// `SpendAuthGBase` reuses the same generator and U/Z tables as
+/// `OrchardFixedBasesFull::SpendAuthG` (same 85-window structure over the
+/// 255-bit pallas base field), allowing `FixedPointBaseField::mul` to accept
+/// an `AssignedCell` directly — no variable-base `NonIdentityPoint` witness
+/// needed.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum OrchardBaseFieldBases {
```
TalDerei, line 72: ACK – `OrchardBaseFieldBases` / `OrchardShortScalarBases` (gated under "unstable-voting-circuits") for voting circuits to call `FixedPointBaseField::mul` on `SpendAuthGBase` and `FixedPointShort::mul` on `SpendAuthGShort` (to avoid expensive variable base mul as the readme points out). same underlying `SpendAuthG` generator though with U/Z precomputed tables.

`NullifierK` / `ValueCommitV` relocation seems fine to me.

src/constants/fixed_bases.rs
TalDerei, line 218: nit; looking through this file, are `impl FixedPoint for NullifierK / ValueCommitV` blocks necessary anymore?

src/constants/fixed_bases.rs
```diff
@@ -184,3 +277,362 @@ impl FixedPoint<pallas::Affine> for ValueCommitV {
...
+        assert_eq!(
+            short.z().len(),
+            NUM_WINDOWS_SHORT,
+            "SpendAuthGShort Z table must have NUM_WINDOWS_SHORT entries"
+        );
+        // The first NUM_WINDOWS_SHORT - 1 windows are identical (same generator
+        // and window layout); the final window differs because the short-scalar
+        // path accounts for the sign bit of a signed scalar. Asserting the
+        // difference at exactly that window documents the invariant, instead of
+        // relying on any mismatch in the slice.
```
TalDerei: hmm validate this "final window" claim.

TalDerei:
> @daira There are a lot of things that are load-bearing for the security of the voting protocol. In this particular case, I've asked my engineer @TalDerei to additionally review the code. I will be doing a second review pass later today. That is more than adequate review for this particular change to the `orchard` crate.

left some observations that aren't blocking, overall looks good!

src/constants/fixed_bases.rs
p0mvn: Good catch.

They aren't needed for in-circuit dispatch anymore (the chip routes through `OrchardBaseFieldBases` / `OrchardShortScalarBases` after this refactor).

Filed a mechanical follow-up that drops both impls and symmetrizes the one regression test that was leaning on the `ValueCommitV` impl as ground truth:
https://github.com/valargroup/orchard/pull/20

I'll retarget it to `zcash/orchard:main` once #489 lands

src/constants/fixed_bases.rs
p0mvn: Validated, and the "sign bit" reason was wrong.

Sign is applied at the [gadget level](https://github.com/zcash/halo2/blob/32a87582dfb0ad9364ef3ffe71751ceab2a502ea/halo2_gadgets/src/ecc/chip/mul_fixed/short.rs#L65), not in `U_SHORT`/`Z_SHORT`.

For `SpendAuthG: Z[0..=20] == Z_SHORT[0..=20]`, `Z[21] ≠ Z_SHORT[21]`. Same pattern on `U`.

Algorithmically, `find_zs_and_us` [builds windows `0..n-1` from `(base, w)`](https://github.com/zcash/halo2/blob/32a87582dfb0ad9364ef3ffe71751ceab2a502ea/halo2_gadgets/src/ecc/chip/constants.rs#L43-L52) only, so the prefix is `num_windows`-independent. The last window has a [num_windows-dependent offset](https://github.com/zcash/halo2/blob/32a87582dfb0ad9364ef3ffe71751ceab2a502ea/halo2_gadgets/src/ecc/chip/constants.rs#L61-L73).

Updated the comment and tightened the test to assert prefix-equality and last-window-inequality on both `U` and `Z`.

[1117335](https://github.com/zcash/orchard/pull/489/commits/11173351a9d028b0656e70e7a82c72259004767e)

ebfull (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to Rust's notion of
...
+- `orchard::constants::{OrchardBaseFieldBases, OrchardShortScalarBases}`
+  enums adding `SpendAuthG` as a supported fixed-base generator for
+  base-field scalar multiplication (`SpendAuthGBase`) and short-scalar
+  multiplication (`SpendAuthGShort`). Both are marked `#[non_exhaustive]`.
+  `orchard::constants::OrchardFixedBases` now uses
+  `Base(OrchardBaseFieldBases)` and `Short(OrchardShortScalarBases)`
+  wrapping variants in place of the previous `NullifierK` and `ValueCommitV`
+  variants; existing `From<NullifierK>` and `From<ValueCommitV>` conversions
+  continue to route correctly. Reachable externally only under the
+  `unstable-voting-circuits` feature.
```
ebfull:
```suggestion
### Changed
- The following APIs have changed or been made available behind the
  `unstable-voting-circuits` feature flag, and are not covered by the
  crate's semver stability guarantees and may change in any future
  release:
  - `orchard::constants::OrchardFixedBases` has been reorganized to wrap
    two new fixed-base enums, each adding `SpendAuthG` as a supported
    generator alongside the existing `NullifierK` / `ValueCommitV`
    generators. Existing `From<NullifierK>` and `From<ValueCommitV>`
    conversions to `OrchardFixedBases` are preserved.
  - `orchard::constants::OrchardBaseFieldBases`: fixed bases for scalar
    multiplication by a base-field scalar.
  - `orchard::constants::OrchardShortScalarBases`: fixed bases for scalar
    multiplication by a short signed scalar.
```

ebfull (changes requested):

ebfull (approved):

src/constants/fixed_bases.rs
```diff
@@ -115,8 +178,8 @@ pub struct ValueCommitV;
 #[cfg(feature = "circuit")]
 impl FixedPoints<pallas::Affine> for OrchardFixedBases {
     type FullScalar = OrchardFixedBasesFull;
-    type Base = NullifierK;
-    type ShortScalar = ValueCommitV;
+    type Base = OrchardBaseFieldBases;
+    type ShortScalar = OrchardShortScalarBases;
```
nuttycom, line 182: These associated types raise a question for me: is it the case that any combination of base field base and short scalar base is valid? My concern as a pattern is that this looks like the kind of place where enums are often misused; you have to consider the cartesian produce of the states that are admitted. Invalid states should be made unrepresentable by construction.

czarcas7ic: I don't think this creates a cartesian-product state. I believe this design already follows that principle, since the two associated types are the mechanism that makes invalid pairings unrepresentable at the type level.

`FixedPoints::Base` and `FixedPoints::ShortScalar` are consumed by different gadget entrypoints. `FixedPointBaseField::mul` dispatches to `mul_fixed_base_field_elem`, which takes only `FixedPoints::Base`; `FixedPointShort::mul` dispatches to `mul_fixed_short`, which takes only `FixedPoints::ShortScalar`. So a `FixedPointBaseField` can only be constructed with `OrchardBaseFieldBases`, and a `FixedPointShort` can only be constructed with `OrchardShortScalarBases`.

So there isn't a value-level pair like `(base_field_base, short_scalar_base)` whose combinations need to be validated. The invariant is scoped to each enum independently: every `OrchardBaseFieldBases` variant is valid for base-field scalar multiplication, and every `OrchardShortScalarBases` variant is valid for short-scalar multiplication. The chip supports both categories, but an individual multiplication still chooses from exactly one category.

It's possible I misunderstood your point though, so please lmk if this satisfies your concern or not.

ebfull: This concern was addressed in #496.

src/constants/fixed_bases.rs
ebfull: This concern was addressed in #496 as well.

PR #490, opened 2026-04-21
orchard_internal split clarifications

ebfull:
1. `orchard_internal` should not have the same version convention as `orchard`, since the public API of `orchard_internal` is _intentionally_ being modified and exposed in ways that would decouple the two when following strict semantic versioning conventions, or require needless major version bumps to `orchard`.
2. Let's _not_ strictly pin from `orchard` to `orchard_internal`. (Just respect semver!)
3. Let's clarify that the `CHANGELOG.md` file applies to the `orchard` crate specifically to avoid mixing the two.


nuttycom (approved):
ACK

PR #492, opened 2026-04-22
consensus: Reject identity `rk` in `Action::from_parts` and `Instance::from_parts`

daira:
## Summary

zcashd v6.12.1 and Zebra 4.3.1 both added a consensus rule rejecting transactions whose Orchard actions have an identity `rk`; the Zcash protocol specification will be updated to match. This PR aligns the `orchard` crate with that rule.

It is an alternative to #487. The approach taken in #487 attempted to exclude the identity point $\mathcal{O}_{\mathbb{P}}$ from `primitives::redpallas::VerificationKey<SpendAuth>`, but currently has a soundness gap that would require more disruptive changes in order to fix.

References for the disclosed vulnerability and the consensus-rule addition:

- <https://zodl.com/zcashd-zebra-april-2026-disclosure/>
- <https://zfnd.org/zebra-4-3-1-critical-security-fixes-dockerized-mining-and-ci-hardening/>

## Design

Enforcement lives at the `Action` and `Instance` constructors rather than at the `primitives::redpallas` or `reddsa` layer, because RedDSA (and by extension RedPallas) does not forbid the identity as a public key. Forbidding identity `rk` is an Orchard-specific consensus concern, not a signature-scheme concern, so the narrowed Orchard types are the right enforcement point.

## Changes

Commit 1 — consensus fix:
- `Action::from_parts` and `Instance::from_parts` now return `Option<Self>`, yielding `None` when `rk` is the identity.
- These checks use a new `pub fn is_identity` method on `VerificationKey<T>`, implemented by comparing the cached byte encoding to `[0u8; 32]`. `reddsa::VerificationKey<T>` caches the byte encoding, so the check is O(1).
- Both `Action` and `Instance` carry a `# Invariants` section documenting that `rk` is non-identity by construction.
- Callers updated:
  - `builder.rs` and the `note_encryption` test use `.expect(...)` — the randomizer $\alpha$ is generated from crypto-quality randomness, so a zero-scalar result is vanishingly unlikely.
  - `pczt/tx_extractor.rs` adds `TxExtractorError::IdentityRk` and propagates on identity `rk` in parsed PCZT actions.
  - `pczt/prover.rs` adds `ProverError::IdentityRk` and propagates on identity `rk` in the actions being proven.
  - The `circuit.rs` test-vector helper uses `.expect(...)` — the serialized test vectors were generated with non-identity `rk`.
  - `Instance::to_halo2_instance`'s `.unwrap()`s on `pallas::Point::from_bytes` and `.coordinates()` become `.expect(...)` with invariant-specific messages.
- Added a new `arb_valid_spendauth_keypair` proptest generator that produces uniformly-distributed valid `(rsk, rk)` pairs with non-identity `rk`, and updated the existing `arb_unauthorized_action` and `arb_action` helpers to use it. This is more clarifying than strictly necessary: it is consistent with enforcing non-identity `rk` as invariants of `Action` and `Instance` rather than of `VerificationKey<SpendAuth>`.
- Unit tests for `VerificationKey<SpendAuth>::is_identity`, for both `Action::from_parts` outcomes, and for both `Instance::from_parts` outcomes.

Commit 2 — tighten `Instance` field visibility:
- Fields are now fully private, with `pub(crate)` accessor methods matching the pattern already in use on `Action`, preventing in-crate code from bypassing `Instance::from_parts` and accidentally violating the invariant.
- `Action::to_instance` now routes through `Instance::from_parts`, with an `.expect(...)` justified by the Action-side invariant.

Commit 3 — PCZT `IdentityRk` test coverage:
- Two tests in `src/pczt.rs` exercise the error wiring on the pczt boundary: `create_proof` returns `ProverError::IdentityRk` when an action has an identity `rk`, and `extract` returns `TxExtractorError::IdentityRk` in the same situation. Both inject identity `rk` via `pczt::Bundle::actions_mut()` after the bundle has reached a valid state, then invoke the method under test.

## Unchanged

- No breaking changes to `primitives::redpallas` module; no need for changes to the `reddsa` crate.
- The `epk_bytes` field in `TransmittedNoteCiphertext` is not checked here; its type in an Action description (§4.6, §5.4.5.5) is `KA^{Orchard}.Public = ℙ*`, but the consensus-level type check lives at the transaction deserializer in a consumer crate such as `zcash_primitives`.

🤖 Filed with [Claude Code](https://claude.com/claude-code).

### DIFF
--- a/src/action.rs
+++ b/src/action.rs
@@ -10,6 +10,10 @@ use crate::{
 ///
 /// This both creates a note (adding a commitment to the global ledger), and consumes
 /// some note created prior to this action (adding a nullifier to the global ledger).
+///
+/// # Invariants
+///
+/// Every `Action` has a non-identity `rk`.
 #[derive(Debug, Clone)]
 pub struct Action<A> {
     /// The nullifier of the note being spent.
@@ -28,6 +32,16 @@ pub struct Action<A> {
 
 impl<T> Action<T> {
     /// Constructs an `Action` from its constituent parts.
+    ///
+    /// Returns `None` if `rk` is the identity [`pasta_curves::pallas::Point`].
+    /// zcashd v6.12.1 and Zebra 4.3.1 both added a consensus rule rejecting
+    /// transactions whose Orchard actions have an identity `rk`; the Zcash
+    /// protocol specification will be updated to match, and this crate aligns
+    /// with that rule.
+    ///
+    /// See:
+    /// - <https://zodl.com/zcashd-zebra-april-2026-disclosure/>
+    /// - <https://zfnd.org/zebra-4-3-1-critical-security-fixes-dockerized-mining-and-ci-hardening/>
     pub fn from_parts(
         nf: Nullifier,
         rk: redpallas::VerificationKey<SpendAuth>,
@@ -35,15 +49,15 @@ impl<T> Action<T> {
         encrypted_note: TransmittedNoteCiphertext,
         cv_net: ValueCommitment,
         authorization: T,
-    ) -> Self {
-        Action {
+    ) -> Option<Self> {
+        (!rk.is_identity()).then_some(Action {
             nf,
             rk,
             cmx,
             encrypted_note,
             cv_net,
             authorization,
-        }
+        })
     }
 
     /// Returns the nullifier of the note being spent.
@@ -132,10 +146,7 @@ pub(crate) mod testing {
             commitment::ExtractedNoteCommitment, nullifier::testing::arb_nullifier,
             testing::arb_note, TransmittedNoteCiphertext,
         },
-        primitives::redpallas::{
-            self,
-            testing::{arb_spendauth_signing_key, arb_spendauth_verification_key},
-        },
+        primitives::redpallas::{self, testing::arb_valid_spendauth_keypair},
         value::{NoteValue, ValueCommitTrapdoor, ValueCommitment},
     };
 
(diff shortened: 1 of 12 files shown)


src/action.rs
```diff
@@ -28,22 +32,35 @@ pub struct Action<A> {
...
         encrypted_note: TransmittedNoteCiphertext,
         cv_net: ValueCommitment,
         authorization: T,
-    ) -> Self {
-        Action {
+    ) -> Option<Self> {
+        if rk.is_identity() {
+            return None;
+        }
+        Some(Action {
```
nuttycom:
```suggestion
        rk.is_identity().then_some(Action {
```
nit: let's avoid early returns when we can.

src/circuit.rs
```diff
@@ -823,16 +833,54 @@ impl Instance {
         cmx: ExtractedNoteCommitment,
         enable_spend: bool,
         enable_output: bool,
-    ) -> Self {
-        Instance {
+    ) -> Option<Self> {
+        if rk.is_identity() {
+            return None;
+        }
+        Some(Instance {
```
nuttycom:
```suggestion
        rk.is_identity().then_some(Instance {
```

nuttycom (dismissed):
utACK 8e89c1551728ce8a2a3bb4d369f1a3edd4321e57 with nonblocking suggestions

src/action.rs
daira: Should be `(!rk.is_identity()).then_some(...)`, but yes I'd independently changed this.

src/circuit.rs
```diff
@@ -844,10 +893,10 @@ impl Instance {
         instance[NF_OLD] = self.nf_old.0;
 
         let rk = pallas::Point::from_bytes(&self.rk.clone().into())
-            .unwrap()
+            .expect("the cached byte encoding of a VerificationKey<_> is canonical")
```
nuttycom, line 896: It's annoying that we have to plumb through the byte encoding here in the first place; `VerificationKey` should just expose the point it corresponds to.

nuttycom (approved):
utACK 0fce1f26a865ac1e96e3bff417b3da9c912dafef

PR #480, opened 2026-04-10
Split orchard into `orchard_internal` (wide visibility) + `orchard` (API-preserving shim)

ebfull:
Downstream extensions to Orchard, such as the new voting protocol, will want to create their own circuits while reusing components of this crate (especially its internal circuits). But, we would prefer that the API is not enlarged on `orchard` itself.

This PR turns this repository into a workspace. The existing `orchard` crate becomes `orchard_internal` (crate name reserved) and `orchard` is published from the `public/` subdirectory, which contains a shim crate implementation that exposes an identical API (and documentation) as the current `orchard` crate.

Later, `orchard_internal` will have many components that are currently `pub(crate)` turned `pub`, and some very minor refactors to enable the voting circuits will also be folded in.

PR #493, opened 2026-04-22
Revert `orchard_internal` crate split

ebfull:
This split ends up being pointless, revert it for now. (Reverts most of #480, #490.)

The idea was to allow APIs that are internal only to the `orchard` crate to be exposed by an internal crate; the external `orchard` crate then just re-exports them. **Unfortunately, quite a few of the APIs that have to be "exposed" are inherent methods that cannot be hidden through re-exports in the public facade.** (There are several existing visibility issues in the code as well, which we should fix, but I"ll save this for another time.)

The folks working on `voting-circuits` will just have to modify #488 and #489 so that they use an `unstable-voting-circuits` feature just like the `unstable-frost` feature in `orchard` today, which grappled with the same issues (see also #428). These changes can be added in an `0.13.1` point release.

**Blocks 0.13.**


nuttycom (dismissed):
~utACK db94124a76bb01b60859192fefba84e152aca9e7. I have reviewed https://github.com/zcash/orchard/compare/efb13500fb6594e2faba26eaf6b3f334060bef8d..db94124a76bb01b60859192fefba84e152aca9e7 and agree that this is the correct reversion of the split.~

nuttycom (approved):
utACK ae56960. I have reviewed https://github.com/zcash/orchard/compare/efb13500fb6594e2faba26eaf6b3f334060bef8d..ae56960  and agree that this is the correct reversion of the split.

PR #495, opened 2026-04-24
value: Replace `NoteValue::zero()` with `NOTE_VALUE_ZERO` const

ValarDragon:
## Summary

- Introduces `pub const orchard::value::NOTE_VALUE_ZERO: NoteValue`, replacing the crate-private `NoteValue::zero()` associated function.
- Rewires all internal call sites in `src/builder.rs`, `src/bundle.rs` (testing helpers), and `src/note.rs` to use the new constant.
- Removes the `NoteValue::zero()` method entirely.
- Adds a CHANGELOG entry under `[Unreleased] → Added` for the new public symbol.

## Why

A `const` is usable in `const` contexts, avoids the function-call indirection, and gives downstream consumers a direct name for the zero value without going through `NoteValue::from_raw(0)`. The previous `zero()` constructor was `pub(crate)`, so exposing the replacement as `pub` is the only external-facing change.

This is follow-up from Daira and Kris' feedback: https://github.com/zcash/orchard/pull/488#discussion_r3132767762

As this is a nice to have, but does not block our voting circuit work. 

## Verification

- [x] `cargo fmt -- --check` — clean
- [x] `cargo build --all-features --benches` — clean, no warnings
- [x] `cargo test` — all tests pass (61 + 1 + 1)
- [x] `cargo doc --all-features --document-private-items --no-deps` — clean
- [x] `cargo clippy --all-features --all-targets` — no new lints (2 pre-existing `useless_conversion` warnings in `src/action.rs` and `src/circuit.rs` are present on `main` and unrelated)

🤖 This PR was prepared with the assistance of Claude Opus 4.7 (1M context).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Added
+- `orchard::value::NoteValue::ZERO`, a `const NoteValue` equal to zero.
+
 ### Changed
 - The following APIs have changed or been made available behind the
   `unstable-voting-circuits` feature flag, and are not covered by the
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -276,7 +276,7 @@ impl SpendInfo {
     }
 
     fn has_matching_anchor(&self, anchor: &Anchor) -> bool {
-        if self.note.value() == NoteValue::zero() {
+        if self.note.value() == NoteValue::ZERO {
             true
         } else {
             let cm = self.note.commitment();
@@ -360,7 +360,7 @@ impl OutputInfo {
         let fvk: FullViewingKey = (&SpendingKey::random(rng)).into();
         let recipient = fvk.address_at(0u32, Scope::External);
 
-        Self::new(None, recipient, NoteValue::zero(), [0u8; 512])
+        Self::new(None, recipient, NoteValue::ZERO, [0u8; 512])
     }
 
     /// Builds the output half of an action.
@@ -639,11 +639,11 @@ impl Builder {
         let value_balance = self
             .spends
             .iter()
-            .map(|spend| spend.note.value() - NoteValue::zero())
+            .map(|spend| spend.note.value() - NoteValue::ZERO)
             .chain(
                 self.outputs
                     .iter()
-                    .map(|output| NoteValue::zero() - output.value),
+                    .map(|output| NoteValue::ZERO - output.value),
             )
             .try_fold(ValueSum::zero(), |acc, note_value| acc + note_value)
             .ok_or(BalanceError::Overflow)?;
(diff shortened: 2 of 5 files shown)


src/value.rs
```diff
@@ -63,6 +63,11 @@ use crate::{
 /// Maximum note value.
 pub const MAX_NOTE_VALUE: u64 = u64::MAX;
 
+/// The zero note value.
+///
+/// Equivalent to `NoteValue::from_raw(0)`.
+pub const NOTE_VALUE_ZERO: NoteValue = NoteValue(0);
```
ebfull: Can't this be an associated const instead?

```rust
impl NoteValue {
    pub const ZERO: Self = NoteValue(0);
}
```

ebfull (changes requested):

src/value.rs
ValarDragon: On it

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Added
+- `orchard::value::NOTE_VALUE_ZERO`, a `pub const NoteValue` equal to zero.
```
ebfull:
```suggestion
- `orchard::value::NOTE_VALUE_ZERO`, a `const NoteValue` equal to zero.
```

visibility is redundant; this is a public API changelog

ebfull (changes requested):

src/builder.rs
```diff
@@ -276,7 +279,7 @@ impl SpendInfo {
     }
 
     fn has_matching_anchor(&self, anchor: &Anchor) -> bool {
-        if self.note.value() == NoteValue::zero() {
+        if self.note.value() == NOTE_VALUE_ZERO {
```
daira: Why not `NoteValue::ZERO`, given that `NoteValue` is public? That seems more idiomatic and matches other instances like `pallas::Base::ZERO`.

ebfull: Duplicate of https://github.com/zcash/orchard/pull/495#discussion_r3135299633.

src/value.rs
p0mvn: Addressed in [0062868](https://github.com/zcash/orchard/pull/495/commits/0062868dddeaffa675aa33451259f04af25aa6b1). Thank you

src/builder.rs
p0mvn: https://github.com/zcash/orchard/pull/495#discussion_r3140962456

ebfull (approved):

p0mvn (commented):
Force-pushed with a clean commit, addressing the comment and updating the changelog.

ebfull:
Actually, let's merge #489 first in case the changelogs conflict. I don't want you to have to rebase #489 since it's the uglier PR.

ebfull (dismissed):

ebfull:
> New changes require approval from someone other than ebfull because they were the last pusher.

Not only is this apparently buggy, it discourages good development hygiene by forcing us to play antics to route around it.

ebfull (approved):

PR #496, opened 2026-04-26
refactor: collapse OrchardFixedBases and drop dead FixedPoint impls

czarcas7ic:
Two follow-up cleanups after the 0.13.1 fixed-base refactor:

- Drop the standalone `impl FixedPoint for NullifierK` / `for ValueCommitV` — dispatch routes through `OrchardBaseFieldBases` / `OrchardShortScalarBases` now, leaving these dead.
- Collapse `OrchardFixedBases` to a unit struct and drop its five dead `From<X>` impls. Nothing constructed it as a value; it's a `FixedPoints` trait carrier only.

All affected items live behind `unstable-voting-circuits`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Removed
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK 53014bea355f35a61e52762dc651f2f5996f3bfc

PR #499, opened 2026-05-08
Add QR Orchard note version support

ValarDragon:
## Summary

- add Orchard note plaintext versions `NoteVersion::V2` and `NoteVersion::V3`,
  and expose `NoteVersion` from the crate root
- parse both Orchard note plaintext lead bytes `0x02` and `0x03`
- keep existing note constructors and builders defaulting to
  `DEFAULT_NOTE_VERSION` / `NoteVersion::V2`
- add explicit version selectors via `Note::from_parts_with_version`,
  `OutputInfo::new_with_version`, and `Builder::add_output_with_version`
- derive V3 note commitment randomness with the ZIP 2005 preimage, including
  `AssetBase = V^Orchard`
- verify QR `rcm` / `cmx` against generated Orchard key-component test vectors

## Test Vectors

This PR now consumes the generated QR fields proposed in
zcash/zcash-test-vectors#129:

- `note_qr_rcm`
- `note_qr_cmx`

The old ad hoc Orchard-only QR regression vector file has been removed in
favor of those generated vectors.

## Notes

For `NoteVersion::V3`, `rcm` is derived as
`ToScalar(PRF^expand(rseed, pre_rcm))`, where `pre_rcm` is
`[0x0B, leadByte] || LEBS2OSP_256(repr_P(g_d)) ||
LEBS2OSP_256(repr_P(pk_d)) || I2LEOSP_64(v) || rho ||
I2LEOSP_256(psi) || LEBS2OSP_256(repr_P(AssetBase))`.

The `qr_rcm` helper takes the `g_d` and `pk_d` elliptic-curve point values
and performs the ZIP 2005 point encoding internally. `AssetBase` is encoded
from the existing `V^Orchard` fixed-base source, with a regression test
checking that the constant equals `value_commit_v::generator().to_bytes()`.

## Tests

- `cargo fmt --check`
- `cargo test qr_rcm`
- `cargo test keys`
- `cargo test`
- `git diff --check`


src/builder.rs
```diff
@@ -335,10 +335,13 @@ pub struct OutputInfo {
     recipient: Address,
     value: NoteValue,
     memo: [u8; 512],
+    note_version: NoteVersion,
 }
 
 impl OutputInfo {
     /// Constructs a new OutputInfo from its constituent parts.
+    /// This defaults to NoteVersion::V2, for compatability with existing wallets.
+    /// After a coordinated upgrade height for default sends being QR is decided upon, this will be updated to NoteVersion::Qr.
```
ebfull: * I prefer for all Rust comments in our codebases to be wrapped to make it easier for others to read them in different editors. I think other maintainers share the same feelings. If you use VScode there's an extension called [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) that I use; just `Alt+q` to wrap everything and clean it up as a habit, using default 80-char settings.
* The first paragraph is always rendered in module-level documentation and should generally be a shorter and more concise explanation of the API's purpose and behavior. The new lines you're adding should be in another paragraph.
* When mentioning specific enum variants, we should doc-link to them (`[NoteVersion::V2].
* "for default sends being" is a little confusing. "This constructor may be changed in a future (breaking) upgrade to use `NoteVersion::Qr`." is more clear.

src/builder.rs
```diff
@@ -350,6 +353,24 @@ impl OutputInfo {
             recipient,
             value,
             memo,
+            note_version: NoteVersion::V2,
```
ebfull: If there are multiple places in the code that will in the future need to be changed in unison (like which is considered the current default note version) then there should be a constant.

src/builder.rs
```diff
@@ -595,20 +617,40 @@ impl Builder {
     }
 
     /// Adds an address which will receive funds in this transaction.
+    /// Defaults to the note version that is standard in the current protocol;
```
ebfull: Easiest to just identify the constant (see above) here representing the enum variant that is the default note version the API is configured to produce.

src/builder.rs
```diff
@@ -595,20 +617,40 @@ impl Builder {
...
+    /// this default may change across wallet SDK updates as the protocol evolves.
     pub fn add_output(
         &mut self,
         ovk: Option<OutgoingViewingKey>,
         recipient: Address,
         value: NoteValue,
         memo: [u8; 512],
+    ) -> Result<(), OutputError> {
+        self.add_versioned_output(ovk, recipient, value, memo, NoteVersion::V2)
```
ebfull: Second example of `NoteVersion::V2` appearing as a default when it should be a shared constant.

src/note.rs
```diff
@@ -23,6 +23,23 @@ pub mod commitment;
...
+/// Note plaintext version, determining rcm derivation strategy.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[allow(non_camel_case_types)]
+pub enum NoteVersion {
+    /// Original Orchard (lead byte 0x02). rcm = PRF(rseed, [0x05] || rho)
+    V2,
+    /// Quantum-recoverable (lead byte 0x03).
+    ///
+    /// rcm = PRF(rseed, [0x0B, 0x03] || g_d || pk_d || v || rho || psi || AssetBase)
+    V3_Qr,
```
ebfull: I don't know whether we should call this `V3_Qr` or just `V3`. **The note plaintext version does more in theory than just determining the `rcm` derivation strategy, even though in this particular upgrade that's all it will do.** The danger in coupling the two is making the change more confusing to reason about, and may lead to logical errors or semantic misconceptions elsewhere.

Aside: In this PR there are some places where you say `V3`, `QR` and `V3_Qr` in doc comments, because you renamed this several times during development. Doc tests would fail if those docs were linking to the variant, which is another good reason to practice that.

src/note.rs
```diff
@@ -134,6 +151,51 @@ impl RandomSeed {
             PrfExpand::ORCHARD_RCM.with(&self.0, &rho.to_bytes()),
         ))
     }
+
+    /// QR rcm derivation per [ZIP 2005].
```
ebfull: I don't mind using `QR` where it is unambiguous that something means "quantum resistant" -- especially if it's documented -- but for clarity let's try to explicitly write "quantum recoverable" in places where we can afford it.

src/note.rs
```diff
@@ -134,6 +151,51 @@ impl RandomSeed {
...
+    ///
+    /// Binds rcm to all note fields for post-quantum commitment binding:
+    ///   pre_rcm = [0x0B, lead_byte] || g_d || pk_d || v_le || rho || psi || AssetBase
+    ///   rcm = ToScalar(PRF^expand(rseed, pre_rcm))
+    ///       = ToScalar(BLAKE2b-512("Zcash_ExpandSeed", rseed || pre_rcm))
```
ebfull: This will render like crap in rustdoc; use codeblocks or use katex math formulas ([supported in this repo](https://github.com/zcash/orchard/blob/main/katex-header.html)).

src/note.rs
```diff
@@ -184,12 +248,26 @@ impl Note {
         value: NoteValue,
         rho: Rho,
         rseed: RandomSeed,
+    ) -> CtOption<Self> {
+        Self::from_parts_versioned(recipient, value, rho, rseed, NoteVersion::V2)
```
ebfull: This method documentation was not updated to mention the default note version, which is inconsistent with the other method(s). This is the third example of a default version that could share the constant.

src/note.rs
```diff
@@ -134,6 +151,51 @@ impl RandomSeed {
...
+
+    /// QR rcm derivation per [ZIP 2005].
+    ///
+    /// Binds rcm to all note fields for post-quantum commitment binding:
+    ///   pre_rcm = [0x0B, lead_byte] || g_d || pk_d || v_le || rho || psi || AssetBase
+    ///   rcm = ToScalar(PRF^expand(rseed, pre_rcm))
+    ///       = ToScalar(BLAKE2b-512("Zcash_ExpandSeed", rseed || pre_rcm))
+    ///
+    /// [ZIP 2005]: https://zips.z.cash/zip-2005
+    pub(crate) fn qr_rcm(
```
ebfull: Does this need to be `pub(crate)`? (Maybe it does, just flagging it. There's plenty of crap with loose `pub(crate)` visibility throughout the codebase already.)

src/note_encryption.rs
```diff
@@ -17,13 +17,18 @@ use crate::{
...
     value::{NoteValue, ValueCommitment},
     Address, Note,
 };
 
 const PRF_OCK_ORCHARD_PERSONALIZATION: &[u8; 16] = b"Zcash_Orchardock";
 
+/// Lead byte for quantum-recoverable Orchard note plaintexts.
+/// This signals that rcm was derived binding all note fields.
+/// Matches the draft ZIP (0x03 = v3/QR note plaintext).
+pub const QR_NOTE_PLAINTEXT_LEAD_BYTE: u8 = 0x03;
```
ebfull: We should probably add a constant for the original `0x02` lead byte and put it here. This kind of a refactoring encourages us to spot all the places where `0x02` might exist that we explicitly (or accidentally) neglected to update to account for the new QR logic.

ebfull (commented):
First pass, mostly high level comments to reduce review burden on others and improve maintainability. I haven't checked through the derivations.

src/note.rs
ValarDragon: Going with just V3 throughout for now then

src/note.rs
ValarDragon: It was not needed, thank you!

src/note_encryption.rs
ValarDragon: Applied!

ValarDragon:
All comments addressed, and added to my local linters! Thank you

New constants made, comments improved, V3_QR -> V3, and equations adapted to katex

src/note_encryption.rs
```diff
@@ -17,13 +17,22 @@ use crate::{
...
 const PRF_OCK_ORCHARD_PERSONALIZATION: &[u8; 16] = b"Zcash_Orchardock";
 
+/// Lead byte for original Orchard note plaintexts.
+pub const ORIGINAL_NOTE_PLAINTEXT_LEAD_BYTE: u8 = 0x02;
+
+/// Lead byte for quantum recoverable Orchard note plaintexts.
+///
+/// This signals that rcm was derived binding all note fields, as defined in
+/// ZIP 2005.
+pub const QR_NOTE_PLAINTEXT_LEAD_BYTE: u8 = 0x03;
```
ebfull: These can just be `pub(crate)`, users of the crate never interact with these constants directly.

src/note.rs
```diff
@@ -205,9 +304,26 @@ impl Note {
         value: NoteValue,
         rho: Rho,
         mut rng: impl RngCore,
+    ) -> Self {
+        Self::new_versioned(recipient, value, rho, &mut rng, DEFAULT_NOTE_VERSION)
+    }
+
+    /// Generates a new note with a specific version.
```
ebfull:
```suggestion
    /// Generates a new note with a specified [`NoteVersion`].
```

src/note.rs
```diff
@@ -23,6 +23,30 @@ pub mod commitment;
 pub use self::commitment::NoteCommitTrapdoor;
 pub use self::commitment::{ExtractedNoteCommitment, NoteCommitment};
 
+/// `LEBS2OSP_256(repr_P(V^Orchard))`, used as `AssetBase` in ZIP 2005
+/// quantum recoverable rcm derivation.
+const ORCHARD_ASSET_BASE_BYTES: [u8; 32] =
+    crate::constants::fixed_bases::value_commit_v::GENERATOR.0;
```
ebfull: This is extracting the $x$-coordinate serialization of $\mathcal{V}^{\text{Orchard}}$ and interpreting it as a compressed encoding of a Pallas point, but this is only accidentally correct because the sign bit of the $y$-coordinate happens to be zero for this base. The `qr_asset_base_encoding_matches_value_commitment_generator` test in this PR guards against this edge case, but we should probably leave a comment here to point it out.

src/test_vectors/qr_note.rs
```diff
@@ -0,0 +1,228 @@
...
+/// Domain separator for quantum recoverable rcm derivation per ZIP 2005.
+///
+/// The PRF input is:
+///
+/// $$
+/// [\mathtt{0x0B}, \mathsf{leadByte}] \mathbin\Vert \mathsf{g}\_{\mathsf{d}}
+/// \mathbin\Vert \mathsf{pk}\_{\mathsf{d}} \mathbin\Vert \mathsf{I2LEOSP}\_{64}(\mathsf{v})
+/// \mathbin\Vert \rho \mathbin\Vert \psi \mathbin\Vert \mathsf{AssetBase}
+/// $$
+pub(crate) const QR_RCM_DOMAIN: [u8; 2] = [0x0B, QR_NOTE_PLAINTEXT_LEAD_BYTE];
```
ebfull: This is not used anywhere in this PR.

`h.update(&[0x0B, QR_NOTE_PLAINTEXT_LEAD_BYTE]);` in `qr_rcm` inlines its definition, so perhaps you could make a constant and refer to it, but I haven't thought this through.

src/test_vectors/qr_note.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// Note recipient: diversifier (11 bytes)
+    pub(crate) d: [u8; 11],
+    /// Note recipient: diversified transmission key pk_d (32 bytes)
+    pub(crate) pk_d: [u8; 32],
+    /// Note value
+    pub(crate) v: u64,
+    /// Note rho (32 bytes, canonical encoding of a Pallas base field element)
+    pub(crate) rho: [u8; 32],
+    /// Random seed (32 bytes)
+    pub(crate) rseed: [u8; 32],
```
ebfull: There are several pieces of dead code in this file that are covered up by `pub(crate)` visibility. This entire module is `cfg(test)` so I think everything in this file should either be `pub` or private, which would cause these all to be warnings caught by CI. (And then, they become test coverage bugs, which must also be fixed.)

src/test_vectors/qr_note.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub(crate) rcm_qr: [u8; 32],
+    /// Note commitment (cmx) using original rcm
+    pub(crate) cmx_old: [u8; 32],
+    /// Note commitment (cmx) using quantum recoverable rcm
+    pub(crate) cmx_qr: [u8; 32],
+}
+
+pub(crate) fn test_vectors() -> Vec<TestVector> {
+    vec![
+        TestVector {
```
ebfull: All of the other test vectors we have in this crate (should) originate from https://github.com/zcash/zcash-test-vectors, which means all of these are currently regression tests, not test vectors.

src/note.rs
```diff
@@ -184,18 +267,34 @@ impl Note {
...
         rho: Rho,
         rseed: RandomSeed,
+    ) -> CtOption<Self> {
+        Self::from_parts_versioned(recipient, value, rho, rseed, DEFAULT_NOTE_VERSION)
+    }
+
+    /// Creates a `Note` from its component parts with a specific version.
+    ///
+    /// Returns `None` if a valid [`NoteCommitment`] cannot be derived from the note.
+    pub fn from_parts_versioned(
```
ebfull: nit: possibly we could use `_with_version` for consistency with most other Rust APIs that do this kind of thing

src/note.rs
```diff
@@ -359,7 +485,229 @@ pub mod testing {
...
+                version: DEFAULT_NOTE_VERSION,
             }
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    extern crate std;
```
ebfull: This probably doesn't do anything.

src/note_encryption.rs
```diff
@@ -59,10 +68,13 @@ where
...
 
-    // Check note plaintext version
-    if plaintext[0] != 0x02 {
-        return None;
-    }
+    // Check note plaintext version; accept both original and quantum
+    // recoverable versions.
+    let note_version = match plaintext[0] {
+        ORIGINAL_NOTE_PLAINTEXT_LEAD_BYTE => NoteVersion::V2,
+        b if b == QR_NOTE_PLAINTEXT_LEAD_BYTE => NoteVersion::V3,
```
ebfull: I think you can pattern on the constant.

```suggestion
        ORIGINAL_NOTE_PLAINTEXT_LEAD_BYTE => NoteVersion::V2,
        QR_NOTE_PLAINTEXT_LEAD_BYTE => NoteVersion::V3,
```

src/test_vectors/qr_note.rs
```diff
@@ -0,0 +1,228 @@
...
+//! \mathbin\Vert \mathsf{g}\_{\mathsf{d}} \mathbin\Vert \mathsf{pk}\_{\mathsf{d}}
+//! \mathbin\Vert \mathsf{I2LEOSP}\_{64}(\mathsf{v}) \mathbin\Vert \rho
+//! \mathbin\Vert \psi \mathbin\Vert \mathsf{AssetBase}
+//! $$
+//!
+//! Lead byte 0x03 denotes the quantum recoverable note version.
+//!
+//! These vectors are derived from the first 3 entries in the Orchard key component test
+//! vectors (https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/orchard_key_components.py),
+//! using the same (sk, note_v, note_rho, note_rseed).
```
ebfull: Those test vectors (currently) simply share those particular inputs, but nothing specifically about QR.

src/note.rs
```diff
@@ -359,7 +485,229 @@ pub mod testing {
...
+            println!("            pk_d: {:02x?},", pk_d_bytes);
+            println!("            v: {},", tv.note_v);
+            println!("            rho: {:02x?},", rho_bytes);
+            println!("            rseed: {:02x?},", tv.note_rseed);
+            println!("            g_d: {:02x?},", g_d_bytes);
+            println!("            rcm_old: {:02x?},", rcm_old_repr);
+            println!("            rcm_qr: {:02x?},", rcm_new_repr);
+            println!("            cmx_old: {:02x?},", cmx_old_bytes);
+            println!("            cmx_qr: {:02x?},", cmx_qr_bytes);
+            println!("        }},");
```
ebfull: These are suppressed test line outputs used to regenerate the test vectors, they should come from `zcash-test-vectors` instead and these print lines should be removed.

src/note_encryption.rs
```diff
@@ -17,13 +17,22 @@ use crate::{
...
     value::{NoteValue, ValueCommitment},
     Address, Note,
 };
 
 const PRF_OCK_ORCHARD_PERSONALIZATION: &[u8; 16] = b"Zcash_Orchardock";
 
+/// Lead byte for original Orchard note plaintexts.
+pub const ORIGINAL_NOTE_PLAINTEXT_LEAD_BYTE: u8 = 0x02;
+
+/// Lead byte for quantum recoverable Orchard note plaintexts.
```
ebfull: (puts on @daira hat) I believe ZIP 2005 consistently uses "quantum-recoverable" as a qualified adjective, so we should here as well. (I distinctly remember us debating this language last year, lol...)


```suggestion
/// Lead byte for quantum-recoverable Orchard note plaintexts.
```

src/note_encryption.rs
ebfull: (actually they could possibly be just private to the module, I'd prefer that)

ValarDragon: Here we need QR_NOTE_PLAINTEXT_LEAD_BYTE in both note.rs and note_encryption.rs , which are two different modules.

To make it private, we'd have to hardcode 0x03 in note.rs, defeating the goal.

Made ORIGINAL_NOTE_PLAINTEXT_LEAD_BYTE private.

src/test_vectors/qr_note.rs
ValarDragon: Just deleted it!

ebfull (commented):
Need clarifications.

src/note.rs
```diff
@@ -134,6 +162,61 @@ impl RandomSeed {
...
+
+        // BLAKE2b-512 with personalization "Zcash_ExpandSeed" (= PRF^expand)
+        let mut h = Params::new()
+            .hash_length(64)
+            .personal(b"Zcash_ExpandSeed")
+            .to_state();
+        // rseed: raw bytes (32 bytes)
+        h.update(&self.0);
+        // domain separator: [0x0B, lead_byte] (2 bytes, literal)
+        h.update(&[0x0B, QR_NOTE_PLAINTEXT_LEAD_BYTE]);
```
ebfull: ZIP 2005 is written under the assumption that `leadByte = 0x03` implies a new note plaintext format change which includes $\mathsf{AssetBase}$, which requires a consensus change to accommodate the larger note ciphertexts:

> Whether or not ZSAs are deployed at the same time, the proposal anticipates that an $\mathsf{AssetBase}$ field will be added to note plaintexts with lead byte $\mathtt{0x03}$. This field will be set to the 32-byte constant $\mathsf{LEBS2OSP}\_{\ell_{\mathbb{P}}}​​(\mathsf{repr}_{\mathbb{P}​}(\mathcal{V}^{\mathsf{Orchard}}))$ as long as ZSAs are not deployed.

The `rcm` derivation in ZIP 2005 goes further and incorporates the `leadByte` into the derivation itself. This PR faithfully reproduces the derivation change as though the note plaintext format change occurred, but configures recipients to parse `0x03` plaintexts as though it never had.

I'm guessing this is to unify the two derivations (they are both used by the quantum recovery circuit) because some of us still anticipate the note plaintext format changes including `AssetBase` are going to happen eventually. But in that case, there's no way we would be okay with using `0x03` for a future note plaintext format that did include `AssetBase`.

This means the ZSA/memo bundle ZIPs would need to change to specify a new `leadByte`. But if that happened, it requires us to keep track of the `leadByte` for recovery, which is silly. It's just a wire format tag, why is that in the `rcm` derivation at all?

src/note.rs
```diff
@@ -152,6 +235,8 @@ pub struct Note {
     rho: Rho,
     /// The seed randomness for various note components.
     rseed: RandomSeed,
+    /// The note plaintext version, determining rcm derivation strategy.
+    version: NoteVersion,
```
ebfull, line 255: Interesting tension that I want to flag somewhere, but I don't know that there's a better option than what you've implemented.

The spec really wants us to continue defining `Note` the way it currently is: ZIP 2005 edits the plaintext definition and the rcm derivation but deliberately leaves notes untouched. This makes sense, since `leadByte` is a property of the plaintext format (and a parameter for the derivation), so this change would add coupling that the spec avoids. There technically isn't currently a concept of a "version" for a note.

In theory, we would introduce some kind of `RcmDerivation` parameter and thread it through the APIs, leaving `Note` untouched. But if we did that, it becomes much more invasive. Even for something like the nullifier computation (strictly speaking) you would need to know the `RcmDerivation`. So you'd have to dig out of this with some major API surgery.

src/note_encryption.rs
```diff
@@ -59,10 +68,13 @@ where
 {
     assert!(plaintext.len() >= COMPACT_NOTE_SIZE);
 
-    // Check note plaintext version
-    if plaintext[0] != 0x02 {
-        return None;
-    }
+    // Check note plaintext version; accept both original and
+    // quantum-recoverable versions.
```
ebfull, line 63: Also flagging for future review pass: the logic for restricting which leadBytes are allowed for particular transaction versions is missing here, which is probably what we want in this PR, just thought I'd record it somewhere.

src/note.rs
```diff
@@ -134,6 +162,61 @@ impl RandomSeed {
...
+        psi: &pallas::Base,
+    ) -> commitment::NoteCommitTrapdoor {
+        use crate::note_encryption::QR_NOTE_PLAINTEXT_LEAD_BYTE;
+        use blake2b_simd::Params;
+
+        // BLAKE2b-512 with personalization "Zcash_ExpandSeed" (= PRF^expand)
+        let mut h = Params::new()
+            .hash_length(64)
+            .personal(b"Zcash_ExpandSeed")
+            .to_state();
```
ebfull: Technically we should be modifying the `zcash_spec` crate and calling into that here, though there's nothing to suggest this code is wrong in the mean time.

src/note.rs
```diff
@@ -228,11 +348,12 @@ impl Note {
         let fvk: FullViewingKey = (&sk).into();
         let recipient = fvk.address_at(0u32, Scope::External);
 
-        let note = Note::new(
+        let note = Note::new_with_version(
             recipient,
             NoteValue::ZERO,
             rho.unwrap_or_else(|| Rho::from_nf_old(Nullifier::dummy(rng))),
             rng,
+            DEFAULT_NOTE_VERSION,
```
ebfull: This is for the dummy notes, I think we want to do this but it raises another concern about if we change `DEFAULT_NOTE_VERSION` someday and there's a mixture of notes produced using `DEFAULT_NOTE_VERSION` (because it's not configurable, as here with dummy notes) and a different (future) plaintext version.

ebfull (commented):
I'm just leaving some reminders for places to look in a more extensive review later.

src/note.rs
ValarDragon: Yeah :/ 

Marking as resolved per signal chat with ZIP authors. Seems like everyone is on board with later on converting this lead byte to a constant in a subsequent update if we have a new note plaintext version that doesn't change the note commitment

src/note.rs
ValarDragon: Moved to spec.rs!

src/note.rs
ValarDragon: True. This only works here because QR and non-QR are simultaneously both valid in V5 transactions.

And when the default note version updates to QR, were already fine sending QR notes.

src/note.rs
```diff
@@ -228,11 +321,12 @@ impl Note {
...
```
TalDerei: referencing #499 (comment), could optionally add a comment here for visibility next to the `Note::dummy` and `OutputInfo::dummy` call sites:

```ts
  // TODO(zip-2005): once V3 emission unlocks, dummy notes must track the
  // bundle's target version
```

src/spec.rs
```diff
@@ -321,10 +396,19 @@ pub fn i2lebsp<const NUM_BITS: usize>(int: u64) -> [bool; NUM_BITS] {
...
+    use group::GroupEncoding;
     use rand::{rngs::OsRng, RngCore};
 
+    #[test]
+    fn qr_asset_base_encoding_matches_value_commitment_generator() {
+        assert_eq!(
+            ORCHARD_ASSET_BASE_BYTES,
+            crate::constants::fixed_bases::value_commit_v::generator().to_bytes(),
+        );
+    }
```
TalDerei: referencing https://github.com/zcash/orchard/pull/499#discussion_r3213750956, could optionally `const_assert` on the `ORCHARD_ASSET_BASE_BYTES` sign bit, but maybe that's unnecessary since we have that regression test.

src/note_encryption.rs
```diff
@@ -59,10 +68,13 @@ where
...
+    let note_version = match plaintext[0] {
+        ORIGINAL_NOTE_PLAINTEXT_LEAD_BYTE => NoteVersion::V2,
+        QR_NOTE_PLAINTEXT_LEAD_BYTE => NoteVersion::V3,
+        _ => return None,
+    };
```
TalDerei: nit; think there should be a module doc covering the _proposed_ activation height contract for in-band secret distribution of orchard bundles (ie. something referencing that callers performing trial decryption need to filter by the activation height / tx version), even though it's not enforced in this PR.

src/spec.rs
```diff
@@ -245,6 +247,79 @@ pub(crate) fn prf_nf(nk: pallas::Base, rho: pallas::Base) -> pallas::Base {
...
+    g_d: &NonIdentityPallasPoint,
+    pk_d: &NonIdentityPallasPoint,
+    value: u64,
+    rho: &pallas::Base,
+    psi: &pallas::Base,
+) -> pallas::Scalar {
+    let mut h = Blake2bParams::new()
+        .hash_length(64)
+        .personal(PRF_EXPAND_PERSONALIZATION)
+        .to_state();
```
TalDerei: validated internal fields against spec.

nit; noticed there was already an existing `PrfExpand`, which could've been used for the QR rcm hash instead of reimplementing BLAKE2b manually, but that requires an upstream change to 
https://github.com/zcash/zcash_spec. not suggesting blocking on that simplification.

src/note.rs
```diff
@@ -134,6 +156,40 @@ impl RandomSeed {
...
+    ///
+    /// [ZIP 2005]: https://zips.z.cash/zip-2005
+    fn qr_rcm(
+        &self,
+        rho: &Rho,
+        g_d: &NonIdentityPallasPoint,
+        pk_d: &NonIdentityPallasPoint,
+        value: u64,
+        psi: &pallas::Base,
+    ) -> commitment::NoteCommitTrapdoor {
```
TalDerei, line 214: nit; by typing `g_d` and `pk_d` as `NonIdentityPallasPoint`, we're serializing these points twice downstream unnecessarily. `qr_rcm` and `h_rcm_orchard_qr` can take bytes, and then `commitment_inner` serializes once, and we can delete the `DiversifiedTransmissionKey` impl.

src/note.rs
```diff
@@ -359,7 +462,143 @@ pub mod testing {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::keys::{FullViewingKey, Scope, SpendingKey};
+    use ff::PrimeField;
+    use group::GroupEncoding;
+
+    #[test]
+    fn qr_rcm_differs_from_old_rcm() {
```
TalDerei: `qr_rcm_differs_from_old_rcm` and `qr_rcm_verify_stored_vectors` have duplicated test logic, and can be extracted into a shared helper that collapses the diff almost entirely.

src/note.rs
```diff
@@ -23,6 +26,25 @@ pub mod commitment;
...
+    /// [ZIP 2005]: https://zips.z.cash/zip-2005
+    V3,
+}
+
+/// The note version produced by constructors that do not take an explicit
+/// [`NoteVersion`].
+///
+/// This is planned to be updated to [`NoteVersion::V3`] once V3 parsing is
+/// sufficiently integrated into wallets.
+pub const DEFAULT_NOTE_VERSION: NoteVersion = NoteVersion::V2;
```
TalDerei: one suggestion is to drop the the `ORIGINAL_NOTE_PLAINTEXT_LEAD_BYTE` and `QR_NOTE_PLAINTEXT_LEAD_BYTE` constants and replace them with a mapping on `NoteVersion`

```ts
  impl NoteVersion {
      /// The note version produced by constructors that do not take an explicit
      /// `NoteVersion`.
      ///
      /// This is planned to be updated to [`Self::V3`] once V3 parsing is
      /// sufficiently integrated into wallets.
      pub const DEFAULT: Self = Self::V2;

      /// Returns the note plaintext lead byte signaling this version.
      pub(crate) const fn lead_byte(self) -> u8 {
          match self {
              Self::V2 => 0x02,
              Self::V3 => 0x03,
          }
      }

      /// Parses a note plaintext lead byte into the corresponding version,
      /// returning `None` if the byte is not a recognized version.
      pub(crate) fn from_lead_byte(b: u8) -> Option<Self> {
          match b {
              0x02 => Some(Self::V2),
              0x03 => Some(Self::V3),
              _ => None,
          }
      }
  }
```

and then the `let note_version = match plaintext[0] { ... }` parsing simplifies to one line:

```ts
let note_version = NoteVersion::from_lead_byte(plaintext[0])?;
```

src/note.rs
TalDerei, line 266: given some semver / versioning ambiguities that were flagged by @ebfull, can we double check this `PartialEq` impl? does anything about its correctness change?

TalDerei (commented):
some general observations, nothing blocking.

src/spec.rs
ebfull: @ValarDragon this is what I was referring to earlier about updating the `zcash_spec` crate. I haven't looked but I didn't understand why things were moved into a `spec` module in response to my earlier feedback.

ValarDragon: Oh thank you! I misunderstood that as meaning the spec.rs file here, my bad.

src/note.rs
ValarDragon: I wanted these to both specifically take in the curve points to remove any risk of serializing incorrectly, or additional serialization checks needed here

ValarDragon:
Thank you for review! Will reply / address each properly in AM my timezone!

src/note.rs
```diff
@@ -23,6 +26,25 @@ pub mod commitment;
...
+    ///
+    /// [ZIP 2005]: https://zips.z.cash/zip-2005
+    V3,
+}
+
+/// The note version produced by constructors that do not take an explicit
+/// [`NoteVersion`].
+///
+/// This is planned to be updated to [`NoteVersion::V3`] once V3 parsing is
+/// sufficiently integrated into wallets.
```
TalDerei: tracing iiuc the intended dependency chain here:
- this first phase in this PR allows wallets to serialize and read v3 notes (parser support at crate level), but not minting / emission support,
- then we'll wait for parsing support to land in wallets,
- then we can wire up the orchard action machinery for constructing V3 notes (`from_action_context_unchecked` would need to dispatch on the note version when witnessing rcm, PCZT would need to carry a version field, and dummy notes would need to become version aware), 
- then wallets opt-in and can actually mint V3 notes,
- then V3 emission eventually becomes the default by flipping the default note version field,
- recovery protocol + qsk/qk spend authority impl (far into the future),
- some kind of auto-migration tooling support (far into the future)

src/note_encryption.rs
daira: Yes this should be in this PR.

daira:
[Claude]
[zcash/zips#1275](https://github.com/zcash/zips/pull/1275) finalised two changes to ZIP 2005's `pre_rcm` that this PR doesn't yet reflect:

1. **Drop `leadByte` from `pre_rcm`.** $\mathsf{H^{rcm,Orchard}}$ no longer takes `leadByte` as a parameter; the `leadByte` case-split for rcm derivation moved up to a `Derive_rcm^{Orchard}` wrapper at the call sites. The current spec has the domain separator as the single byte `[0x0B]`.
2. **Drop `AssetBase` from `pre_rcm`.** ZIP 2005 now commits to activating without ZSAs, so `AssetBase` is no longer in the preimage.

Net effect on `src/spec.rs::h_rcm_orchard_qr`:

```
pre_rcm = [0x0B] || g_d★ || pk_d★ || I2LEOSP_64(v) || ρ || I2LEOSP_256(ψ)
```

(137 bytes vs the current 170.) `ORCHARD_ASSET_BASE_BYTES` and the `lead_byte` parameter become unused; the `qr_asset_base_encoding_matches_value_commitment_generator` regression test becomes moot. The `note_qr_rcm` / `note_qr_cmx` vectors in zcash/zcash-test-vectors#129 will need regenerating against the no-leadByte / no-AssetBase preimage before this PR can land.

The `NoteVersion::V2`/`V3` discriminator on `Note` is still needed for plaintext-format handling on parse/serialize, but no longer for `Derive_rcm` dispatch — it's worth updating the rationale comment to reflect that.

🤖 Claude Opus 4.7 (1M context)

src/note.rs
daira: > * recovery protocol + qsk/qk spend authority impl (far into the future),
> * some kind of auto-migration tooling support (far into the future)

Not too far I hope. Ideally 2026-7.

ValarDragon: Dependency chain here is:
* this first phase in this PR allows wallets to serialize and read v3 notes (parser support at crate level), but not minting / emission support
* We make **hotkey-only wallets do QR recoverable change notes**
* We wait for parsing to land in Keystone firmware
* We enable QR change notes in Keystones for some period of time
* We set a QR send height, and wire sending through the librustzcash + wallet logic
* After that block height, we update all these constants.
* (far into future components)

TalDerei: yeah this directionally tracks my operating mental model!

src/note.rs
ValarDragon: FLAGGING: This is a discrepancy with our latest ZIP update. I believe we want the current code logic, and need to adapt the ZIP. 
image

src/note.rs
ValarDragon: Good idea, applied!

ValarDragon:
Updated according to the ZIP updates. Also updated the PCZT crate, in order to make this version work with Keystone's call to checking the Note Commitment

ValarDragon:
zcash_spec change prepared separately here: https://github.com/zcash/zcash_spec/pull/14

src/note.rs
ValarDragon: Should be correct since it re-computes the commitment. And the commitment depends on the note version

src/note.rs
```diff
@@ -263,6 +400,26 @@ impl Note {
         self.rho
     }
 
+    /// Returns the version of this note.
+    pub fn version(&self) -> NoteVersion {
+        self.version
+    }
+
+    /// Derives the note commitment trapdoor for this note.
+    pub(crate) fn rcm(&self) -> commitment::NoteCommitTrapdoor {
```
ValarDragon, line 412: It became more natural to add an rcm function here, so the witness generation logic does not need to do the match on note version

ValarDragon:
Latest commit updates the Prover witness generation. (Apologies, this was missed earlier, was now end to end testing this more)

TalDerei:
> Latest commit updates the Prover witness generation. (Apologies, this was missed earlier, was now end to end testing this more)

don't think this was missed necessarily, was assuming this was follow-up work per my https://github.com/zcash/orchard/pull/499#discussion_r3243001124:

"then we can wire up the orchard action machinery for constructing V3 notes (`from_action_context_unchecked` would need to dispatch on the note version when witnessing rcm (_branching in a dedicated rcm method works too as you mentioned in_ https://github.com/zcash/orchard/pull/499#discussion_r3261292519), PCZT would need to carry a version field, and dummy notes would need to become version aware)."

we should probably also make the dummy notes version aware while we're at it?

src/spec.rs
TalDerei: tracking in https://github.com/zcash/zcash_spec/pull/14.

src/note.rs
ValarDragon: Resolved in latest ZIP update of https://github.com/zcash/zips/pull/1276

Dummy notes are allowed to have any version.

ValarDragon:
> we should probably also make the dummy notes version aware while we're at it?

Isn't the dummy note version handled with the NoteVersion field in Spend & Output though?

src/note.rs
ValarDragon: They don't have to! Its fine for them to be either version, but sensibly they would just be V3. For v6 tx formats, they would need to be V3

TalDerei:
> Isn't the dummy note version handled with the NoteVersion field in Spend & Output though?

technically yes, but the observation in https://github.com/zcash/orchard/pull/499#discussion_r3216273046 still constitutes a hygiene gap. I think without addressing that piece, there's an asymmetry in the builder where it's like _partially_ version aware. 

PCZT now carries the version field, but it's always using the default note version because of `SpendInfo::dummy` / `OutputInfo::dummy` hard code to the default. the intention with that comment was just to point out that if we're already threading versioning through the PCZT layer, the dummies should probably be included too (otherwise the dummy constructors are the only version-unaware piece in the whole pipeline). though mixed bundles are fine today and we could defer this change to when the default flips.

daira:
I will review this on Monday, if not before.

PR #505, opened 2026-06-19
Integrate QR and TxId into Orchard

czarcas7ic:
This integrates QR into Orchard.

The change introduces an API break by making `NoteVersion` part of the `Note` struct and note construction flow.

The note version is derived from the selected bundle protocol: Orchard bundle protocols use V2 notes, while Ironwood uses V3 notes.

The protocol choice is also threaded through builder, PCZT, commitment, and proof paths so note version, flag encoding, commitment domains, and circuit selection stay tied to the same bundle protocol.

PR #509, opened 2026-06-23
Add QR note version support

ValarDragon:
This PR integrates the QR (quantum-recoverable) note-version cryptography into the Ironwood branch.

## rcm derivation and note versioning

We add the quantum-recoverable `rcm` derivation per [ZIP 2005] as `RandomSeed::rcm_v3` (the existing [ZIP 212] derivation is renamed `RandomSeed::rcm_v2`); both are gated behind `unstable-voting-circuits`. A `Note` now carries a `NoteVersion` — `V2` ([ZIP 212], lead byte `0x02`) or `V3` ([ZIP 2005], lead byte `0x03`) — which selects its `rcm` derivation. The `rcm_v3` derivation is pulled from #499.

Test vectors: we extend the existing key-component vectors with `note_qr_rcm` / `note_qr_cmx` fields, from https://github.com/zcash/zcash-test-vectors/pull/129.

## Note encryption domains

We introduce a second domain, `note_encryption::IronwoodDomain`, which accepts only `0x03` note plaintexts and refuses `0x02`; `OrchardDomain` correspondingly accepts only `0x02` and refuses `0x03`.

Rather than two large struct definitions, both are type aliases over a single generic `NoteEncryptionDomain<P>` whose marker type (`OrchardVersion` / `IronwoodVersion`) selects the accepted version via a sealed `DomainVersion` trait. This gives code reuse while keeping `OrchardDomain` and `IronwoodDomain` as distinct types. Encryption behavior is shared: the plaintext lead byte comes from `Note::version()`, and the domain governs only which versions are accepted during parsing/decryption.

A third, crate-internal domain, `BundleDomain`, carries the target `BundlePoolRestrictions` and enforces the matching note version *after* decryption. Accordingly, the public `Bundle::{decrypt_outputs_with_keys, decrypt_output_with_key, recover_outputs_with_ovks, recover_output_with_ovk}` helpers now take a `BundlePoolRestrictions`, which is what lets them discover `V3` Ironwood notes.

## Builder and PCZT

The builder derives the note version from `BundlePoolRestrictions::note_version` and stamps it onto outputs, change, dummy, and fabricated notes. Supplied outputs/change whose version does not match are rejected with the new `BuildError::InvalidNoteVersion` (spends are not version-checked). The low-level note/builder constructors (`Note::{from_parts, new}`, `OutputInfo::new`, `ChangeInfo::new`) now take an explicit `NoteVersion`.

PCZT `Spend` and `Output` gain a `note_version` field and getter, threaded through `parse`, the prover, and the verifier; `Bundle::parse` rejects an output whose note version mismatches the pool restrictions with `ParseError::InvalidNoteVersion`.

I _do not_ recommend reviewing this commit-by-commit, but a reviewer can choose to do so.

[ZIP 212]: https://zips.z.cash/zip-0212
[ZIP 2005]: https://zips.z.cash/zip-2005


src/note.rs
```diff
@@ -284,14 +409,18 @@ impl Note {
...
-            g_d.to_bytes(),
-            self.recipient.pk_d().to_bytes(),
+            g_d_bytes,
+            pk_d_bytes,
             self.value,
             self.rho.0,
-            self.rseed.psi(&self.rho),
-            self.rseed.rcm(&self.rho),
+            psi,
+            self.rcm(),
```
czarcas7ic, line 433: Super minor nit, this function `commitment_inner` calculates `g_d`, `pk_d`, and `psi` above, then calls `self.rcm()` here, which recalculates those values again.

czarcas7ic (approved):

src/note.rs
ValarDragon: @talderei had left the same comment in #499 as well actually

https://github.com/zcash/orchard/pull/499#discussion_r3241635739

My suggestion was to serialize this twice to reduce any risk of qr_rcm()'s arguments taking in wrong inputs, and only doing a refactor if its performance sensitive.

src/note.rs
```diff
@@ -134,6 +173,59 @@ impl RandomSeed {
...
+        // g_d: LEBS2OSP_256(repr_P(g_d)) — compressed Pallas point (32 bytes)
+        h.update(&g_d.to_bytes());
+        // pk_d: LEBS2OSP_256(repr_P(pk_d)) — compressed Pallas point (32 bytes)
+        h.update(&pk_d.to_bytes());
+        // v: I2LEOSP_64(v) — unsigned 64-bit little-endian (8 bytes)
+        h.update(&value.to_le_bytes());
+        // rho: LEBS2OSP_256(repr_P(rho)) — Pallas base field canonical repr (32 bytes)
+        h.update(&rho.0.to_repr());
+        // psi: LEBS2OSP_256(repr_P(psi)) — Pallas base field canonical repr (32 bytes)
+        h.update(&psi.to_repr());
```
ValarDragon, line 228: One point of feedback in #499 was to update zcash_spec: https://github.com/zcash/zcash_spec/pull/14

My suggestion would be for us to do that in a subsequent PR if desired, as there is no tests in that repo anyway, so the real review is happening here.

TalDerei:
will do a pass and orient the diff against my review of https://github.com/zcash/orchard/pull/499.

TalDerei:
@ValarDragon does this now completely superceed https://github.com/zcash/orchard/pull/505? making sure to avoid reviewing the wrong thing.

ValarDragon:
I'm doing a split of tx-id logic from #505 on-top of this, so #505 should become fully obsoleted.

CHANGELOG.md
```diff
@@ -65,6 +75,9 @@ the current behavior by selecting `BundlePoolRestrictions::OrchardNu6_2Only` (an
   transfers. `arb_flags` is unchanged and only generates flag sets with
   cross-address transfers enabled, which are representable under every pool
   restriction other than Orchard post-NU6.3.
+- `orchard::bundle::testing::arb_bundle_protocol_restriction` (under the
+  `test-dependencies` feature), a strategy that generates bundle pool
```
ebfull, line 87: I think we could consolidate the test-dependencies CHANGELOG updates in a future PR.

src/bundle.rs
```diff
@@ -120,6 +116,19 @@ impl BundlePoolRestrictions {
...
+    /// Orchard pools use V2 note plaintexts, and Ironwood pools use V3 note
+    /// plaintexts.
+    pub fn note_version(self) -> NoteVersion {
+        match self {
+            BundlePoolRestrictions::OrchardPreNu6_2 => NoteVersion::V2,
+            BundlePoolRestrictions::OrchardNu6_2Only => NoteVersion::V2,
+            BundlePoolRestrictions::OrchardNu6_3Onward => NoteVersion::V2,
+            BundlePoolRestrictions::IronwoodNu6_3Onward => NoteVersion::V3,
+        }
+    }
```
ebfull, line 129:
```suggestion
    pub fn note_version(self) -> NoteVersion {
        use BundlePoolRestrictions::*;
        match self {
            OrchardPreNu6_2 | OrchardNu6_2Only | OrchardNu6_3Onward => NoteVersion::V2,
            BundlePoolRestrictions::IronwoodNu6_3Onward => NoteVersion::V3,
        }
    }
```

ebfull (approved):

src/pczt/parse.rs
```diff
@@ -39,6 +41,16 @@ impl Bundle {
         let flags =
             Flags::from_byte(flags, pool_restrictions).ok_or(ParseError::UnexpectedFlagBitsSet)?;
 
+        let note_version = pool_restrictions.note_version();
+        for action in actions.iter() {
+            if *action.spend.note_version() != note_version {
+                return Err(ParseError::InvalidNoteVersion);
+            }
```
ebfull: Why would we care about the spend note version here?

CHANGELOG.md
```diff
@@ -150,6 +167,19 @@ the current behavior by selecting `BundlePoolRestrictions::OrchardNu6_2Only` (an
...
   property, returning `ProverError::DisallowedCrossAddressTransfer` (or
   `ProverError::MissingRecipient` if a `recipient` field is unset).
+- `orchard::pczt::{Spend, Output}::parse` now take the note plaintext version
+  for the parsed spend or output, and `orchard::pczt::Bundle::parse` rejects
+  actions whose spend or output note version does not match the
+  `BundlePoolRestrictions`.
+- `test-dependencies` note and bundle strategies now select note versions from
+  their callers or generated bundle pool restrictions:
+  - `orchard::note::testing::arb_note` now takes a note version.
+  - `orchard::action::testing::{arb_action, arb_unauthorized_action}` now take a
```
ebfull: These aren't in the public API, they're exported through a different path (`orchard::bundle::testing::{arb_action, arb_unauthorized_action}`).

src/note.rs
```diff
@@ -23,6 +27,41 @@ pub mod commitment;
 pub use self::commitment::NoteCommitTrapdoor;
 pub use self::commitment::{ExtractedNoteCommitment, NoteCommitment};
 
+/// Note plaintext version.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum NoteVersion {
```
ebfull, line 32: This API isn't documented in the changelog.

src/note.rs
```diff
@@ -263,6 +368,26 @@ impl Note {
         self.rho
     }
 
+    /// Returns the version of this note.
+    pub fn version(&self) -> NoteVersion {
```
ebfull, line 375: This isn't documented in the changelog.

src/note.rs
```diff
@@ -184,12 +278,14 @@ impl Note {
         value: NoteValue,
         rho: Rho,
         rseed: RandomSeed,
+        version: NoteVersion,
```
ebfull, line 284: This is a breaking change that should be documented in the changelog.

src/builder.rs
```diff
@@ -456,13 +463,15 @@ impl OutputInfo {
         ovk: Option<OutgoingViewingKey>,
         recipient: Address,
         value: NoteValue,
+        note_version: NoteVersion,
```
ebfull, line 468: This is a breaking change not documented in the changelog.

src/builder.rs
```diff
@@ -472,24 +481,25 @@ impl OutputInfo {
...
+            note_version,
             randomized_ciphertext: true,
         }
     }
 
     /// Defined in [Zcash Protocol Spec § 4.8.3: Dummy Notes (Orchard)][orcharddummynotes].
     ///
     /// [orcharddummynotes]: https://zips.z.cash/protocol/nu5.pdf#orcharddummynotes
-    pub fn dummy(rng: &mut impl RngCore) -> Self {
+    pub fn dummy(note_version: NoteVersion, rng: &mut impl RngCore) -> Self {
```
ebfull, line 500: Also needs to be documented.

ebfull (changes requested):
There are some major gaps in the changelog that should be fixed.

src/pczt/parse.rs
ValarDragon: I don't think we care, was indexing off of some comments in the past about wanting wallets to reject funds (e.g. don't consider funds accepted in Ironwood if its 0x02)

That should be handled at the note_encryption layer though. Will remove here

src/bundle.rs
TalDerei, line 1: there are existing public API helpers in this file (`decrypt_outputs_with_keys, decrypt_output_with_key, recover_outputs_with_ovks, recover_output_with_ovk`) that need to be threaded with `BundlePoolRestrictions`.

why? well, before [https://github.com/zcash/orchard/pull/499](https://github.com/zcash/orchard/%EE%80%80pull/499) there only existed an "orchard domain," so the plaintext parser unconditionally accepted both lead bytes:

```ts
let note_version = NoteVersion::from_lead_byte(plaintext[0])?;
```

but that's now stale against the changes in https://github.com/zcash/orchard/commit/a4f4c01e58a106f444d2d9e92cd6b4d4b2014912 that now make this enforcement pool specific.

so these helpers effectively hard code `OrchardDomain` and ironwood notes would technically be undiscovered for those callers.

so for completeness, suggesting these be contextually aware of `BundlePoolRestrictions`. we might as well fold in these changes in this PR as one consolidatable unit because it doesn't hurt.

CHANGELOG.md
```diff
@@ -150,6 +167,19 @@ the current behavior by selecting `BundlePoolRestrictions::OrchardNu6_2Only` (an
   `FixedPostNu6_2`), and checks the cross-address restriction's same-expanded-receiver
   property, returning `ProverError::DisallowedCrossAddressTransfer` (or
   `ProverError::MissingRecipient` if a `recipient` field is unset).
+- `orchard::pczt::{Spend, Output}::parse` now take the note plaintext version
+  for the parsed spend or output, and `orchard::pczt::Bundle::parse` rejects
+  actions whose spend or output note version does not match the
+  `BundlePoolRestrictions`.
```
TalDerei: changelog mentions that `BundlePoolRestrictions` determines `note_version`. but there isn't an explicit mention of direct constructor/signature changes. we should mention that `Note::from_parts / Note::new / OutputInfo::new / ChangeInfo::new` requires `NoteVersion` as a changelog entry:

```ts
   The builder derives new output, dummy, and fabricated note plaintext versions
    from `BundlePoolRestrictions::note_version`, and rejects spends whose note
    version does not match the builder's pool restrictions with
    `SpendError::InvalidNoteVersion`.
```

CHANGELOG.md
ValarDragon: Ok removing these details from here

src/builder.rs
TalDerei, line 525: a bit lower down, i noticed that encryptor still uses `OrchardNoteEncryption` despite `IronwoodNoteEncryption` now existing? should we select the encryptor by `note_version`, or document that the encryptor aliases only differ for parsing/decryption and encryption is separately driven by `Note::version()`? this just seems confusing without any kind of docs.

src/builder.rs
```diff
@@ -816,6 +833,9 @@ impl Builder {
         if !flags.spends_enabled() {
             return Err(SpendError::SpendsDisabled);
         }
+        if note.version() != self.note_version() {
+            return Err(SpendError::InvalidNoteVersion);
+        }
```
TalDerei: noticed `Builder::add_spend` docs still mentions that the input note is obtained by trial decrypting with `OrchardDomain`. there's probably other stale references like this.

CHANGELOG.md
ebfull: I'm going to revert your removal of this, since it was just a path adjustment that needed to happen.

src/pczt/parse.rs
ebfull: This was addressed in c445aba7a9fcefd2b8ee9cc90380a2b369623ff6.

TalDerei:
> Orchard Domain similarly rejects 0x03 note plaintexts and only accepts 0x03.

misstatement

ValarDragon:
Thank you, fixed PR description

src/note.rs
```diff
@@ -23,6 +27,41 @@ pub mod commitment;
...
+
+    /// Parses a note plaintext lead byte into the corresponding version,
+    /// returning `None` if the byte is not a recognized version.
+    pub(crate) fn from_lead_byte(b: u8) -> Option<Self> {
+        match b {
+            0x02 => Some(Self::V2),
+            0x03 => Some(Self::V3),
+            _ => None,
+        }
+    }
```
TalDerei, line 62: deadcode? don't see it used anywhere and it's marked `pub(crate)`.

src/pczt/parse.rs
```diff
@@ -342,6 +358,8 @@ pub enum ParseError {
     /// The provided `flags` field had unexpected bits set for the bundle's pool
     /// restrictions.
     UnexpectedFlagBitsSet,
+    /// An invalid `note_version` was provided.
+    InvalidNoteVersion,
```
TalDerei, line 359: nit; proposing we have some tests around `ParseError::InvalidNoteVersion` and `Bundle::parse` version checks. also, it'd be prudent to test PCZT serialization roundtripping for `note_version`.

src/circuit.rs
```diff
@@ -267,11 +267,11 @@ impl Circuit {
         let sender_address = spend.note.recipient();
         let rho_old = spend.note.rho();
         let psi_old = spend.note.rseed().psi(&rho_old);
-        let rcm_old = spend.note.rseed().rcm(&rho_old);
+        let rcm_old = spend.note.rcm();
```
ebfull, line 270: This reveals an issue, we should consider making the API defensive so that you cannot forget to call `note.rcm()` instead of `note.rseed.rcm()`.

ebfull (changes requested):

src/note_encryption.rs
```diff
@@ -178,116 +160,338 @@ impl Domain for OrchardDomain {
...
+#[derive(Debug)]
+pub struct IronwoodDomain {
+    shared: SharedDomain,
+}
 
-    fn enc_ciphertext(&self) -> &[u8; ENC_CIPHERTEXT_SIZE] {
-        &self.output().encrypted_note().enc_ciphertext
-    }
+impl_domain!(OrchardDomain, NoteVersion::V2);
+impl_domain!(IronwoodDomain, NoteVersion::V3);
```
TalDerei: not the way i would've necessarily done this, because really these domains only differ by their `NoteVersion` constant (so what `impl_domain` macro expands to is pure duplication); in that context, maybe instead of two wrapper structs + `impl_domain` macro, i would've considered a generic sealed marker impl:

```ts

  mod sealed { pub trait Sealed {} }
  pub trait DomainVersion: sealed::Sealed { const NOTE_VERSION: NoteVersion; }

  pub enum OrchardVersion {}   impl sealed::Sealed for OrchardVersion {}
  pub enum IronwoodVersion {}  impl sealed::Sealed for IronwoodVersion {}
  impl DomainVersion for OrchardVersion  { const NOTE_VERSION: NoteVersion = NoteVersion::V2; }
  impl DomainVersion for IronwoodVersion { const NOTE_VERSION: NoteVersion = NoteVersion::V3; }

  #[derive(Debug)]
  pub struct NoteEncryptionDomain<V: DomainVersion> { rho: Rho, _v: PhantomData<V> }
  pub type OrchardDomain  = NoteEncryptionDomain<OrchardVersion>;
  pub type IronwoodDomain = NoteEncryptionDomain<IronwoodVersion>;

  impl<V: DomainVersion> Domain for NoteEncryptionDomain<V> {}
```

which deletes both macros and `SharedDomain`.

src/note.rs
TalDerei: well actually now that i think about it, assuming this was for maintaining the status quo for trial decryption work where you'd only need to trial decrypt once? currently, doesn't your `SharedDomain` impl technically necessitate having to trial decrypt twice (call `try_note_decryption` once for each domain?).  we should decrypt once, then read the revealed byte and pick V2/V3 (the `from_lead_byte` path).

TalDerei: actually shouldn't https://github.com/zcash/orchard/pull/509/changes#r3469193011 transitively fix this issue by threading the proper domain?

src/bundle.rs
```diff
@@ -905,8 +914,19 @@ pub mod testing {
...
 
+    /// Create an arbitrary bundle protocol restriction.
+    pub fn arb_bundle_protocol_restriction() -> impl Strategy<Value = BundlePoolRestrictions> {
+        prop_oneof![
+            Just(BundlePoolRestrictions::OrchardPreNu6_2),
+            Just(BundlePoolRestrictions::OrchardNu6_2Only),
+            Just(BundlePoolRestrictions::OrchardNu6_3Onward),
+            Just(BundlePoolRestrictions::IronwoodNu6_3Onward),
+        ]
+    }
```
TalDerei: should be named "pool_restriction", and there's other callers with the same stale naming.

src/bundle.rs
```diff
@@ -120,6 +116,18 @@ impl BundlePoolRestrictions {
...
+    ///
+    /// Orchard pools use V2 note plaintexts, and Ironwood pools use V3 note
+    /// plaintexts.
+    pub fn note_version(self) -> NoteVersion {
+        use BundlePoolRestrictions::*;
+        match self {
+            OrchardPreNu6_2 | OrchardNu6_2Only | OrchardNu6_3Onward => NoteVersion::V2,
+            BundlePoolRestrictions::IronwoodNu6_3Onward => NoteVersion::V3,
+        }
+    }
```
TalDerei, line 129: noticed that something like `Builder::add_spend` correctly rejects a note whose `note.version` differs from `self.pool_restrictions.note_version`, but the enforcement feels a bit loose because the lower level public `builder::bundle` can just take supplied `SpendInfo, OutputInfo, and ChangeInfo`, and internally `build_bundle` doesn't validates their note versions.

TalDerei (commented):
some design observations from a first pass.

src/note_encryption.rs
```diff
@@ -178,116 +160,338 @@ impl Domain for OrchardDomain {
...
 }
 
-/// Implementation of in-band secret distribution for Orchard bundles.
-pub type OrchardNoteEncryption = zcash_note_encryption::NoteEncryption<OrchardDomain>;
+macro_rules! impl_domain {
+    ($domain:ty, $note_version:expr) => {
+        impl memuse::DynamicUsage for $domain {
+            fn dynamic_usage(&self) -> usize {
+                self.shared.dynamic_usage()
+            }
```
TalDerei: we should double check macro correctness.

ebfull: I am going to look at this again and try to just delete the macro...

ValarDragon: Sgtm! We could just take code reuse and write it twice.

was on the fence for if the macro is worth it

src/bundle.rs
ebfull: I agree they should be consistent, but this is another example where the pool restrictions _on spend_ shouldn't exist IMO.

src/note.rs
TalDerei: @ebfull @ValarDragon am i missing anything here? that seems to be the current behavior, and it doesn't look totally wired in.

ebfull: This has been addressed by your other suggestion indeed.

src/test_vectors/keys.rs
```diff
@@ -21,11 +21,13 @@ pub(crate) struct TestVector {
...
     pub(crate) note_rseed: [u8; 32],
     pub(crate) note_cmx: [u8; 32],
+    pub(crate) note_qr_rcm: [u8; 32],
+    pub(crate) note_qr_cmx: [u8; 32],
     pub(crate) note_nf: [u8; 32],
 }
 
 pub(crate) fn test_vectors() -> Vec<TestVector> {
-    // From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/orchard_key_components.py
+    // From https://github.com/zcash/zcash-test-vectors/blob/master/zcash_test_vectors/orchard/key_components.py
```
ebfull, line 30: Note: This hasn't been updated with https://github.com/zcash/zcash-test-vectors/pull/129 yet, but after that merges it should exist in this path.

**I have checked that these test vectors match.**

src/note.rs
```diff
@@ -263,6 +371,33 @@ impl Note {
         self.rho
     }
 
+    /// Returns the version of this note.
+    pub fn version(&self) -> NoteVersion {
+        self.version
+    }
+
+    /// Derives the ψ value for this note.
```
ebfull, line 379:
```suggestion
    /// Derives the $\psi$ value for this note.
```

ebfull (approved):

ValarDragon:
The sealed method for implementing Domain is much better than the macro! Clever idea to prevent no other impl's, hadn't seen this pattern in Rust before

Reviewed all follow-on commits, they LGTM

czarcas7ic (approved):
Re-reviewed the commits added since my last review (ef10e25ebca5008750a6a0588db74e62ab2abeef..0947ee269616ca5cad758c69ada20170aab40958), no further comments.

TalDerei (approved):
reviewed the recent commits, LGTM! a few notes:

- post-merge, we should run zkao against this PR for more coverage, 
- should we fold in @daira's OOB comments on `note_qr_nf` inclusion alongside `note_qr_rcm` and `note_qr_cmx`? or maybe as a follow up to this,
- @ValarDragon @czarcas7ic process note, since this came to mind from @ebfull's feedback. after https://github.com/zcash/orchard/pull/509/commits/5d349dce651ea9541053a96206989ffb6bc185f3, i was looking through some of the upstream librustzcash https://github.com/valargroup/librustzcash/pull/48 which now seems stale and needs a small change (removing the unused error variants in the spend path for PCZT note version checks) since those changes landed before this QR work! that was an obvious one, but wondering if there are any other assumptions or changes in the valar repo that assumed QR would land in its original form that are now stale / invalidated.

ebfull:
Thanks all!

* Let's do `zkao` and `v12` on the feature branch immediately after #510 merges as well.
* `note_qr_nf` stuff can be addressed separately IMO. See #511.
* Post-hoc reviews accepted! Any follow-ups can be dealt with in separate PRs; for example, I have a few orthogonal API improvements I'd like to toss into this release if there's time, but they would weigh too heavy on this branch right now.

This has received enough review feedback and ACKs to merge, and I'm satisfied with it.

TalDerei:
> * `note_qr_nf` stuff can be addressed separately IMO. Asked @TalDerei to open an issue about it.

filed https://github.com/zcash/orchard/issues/511

PR #510, opened 2026-06-24
Add in v6 and Ironwood TX-id / Auth-ID code

ValarDragon:
This PR brings the v6 transaction hashing (txid / authorizing digest) code for Orchard and Ironwood bundles into the branch. It originated as the V6 tx-hashing work from #505 layered on the QR note-version PR (#509).

Rather than introducing a single combined struct, commitment computation is selected by **two parameters** threaded through the public commitment API:

- `BundlePoolRestrictions` — the `(pool, era)` selector. It determines the **flag-byte encoding** (pre-NU6.3 rules where bit 2 is reserved, vs. NU6.3 rules where bit 2 is `enableCrossAddress`).
- `TxVersion` (`V5` / `V6`) — the transaction version the bundle is encoded in. It selects the **commitment personalization strings** and **where the anchor is committed**: a v5 transaction includes the anchor in the transaction-ID (effects) digest, while a v6 transaction includes it in the authorizing digest instead.

Both are now arguments to `Bundle::commitment`, `Bundle::<Authorized, V>::authorizing_commitment`, and the `hash_bundle_txid_empty` / `hash_bundle_auth_empty` helpers. This is required because the sighashing code definitively needs to know both the pool (for the flag byte) and the transaction version (for the personalizations and anchor placement) — neither alone is sufficient.

_Internally_, `(pool_restrictions, tx_version)` is resolved by `BundlePoolRestrictions::commitment_format(tx_version)` into a private `BundleCommitmentFormat` enum with three variants:

- `OrchardV5` — anchor in the txid digest; legacy personalizations.
- `OrchardV6` — anchor in the authorizing digest; reuses the v5 action-level personalizations and gains `_v6` top-level `bundle`/`auth` strings.
- `IronwoodV6` — anchor in the authorizing digest; fresh `_v6` personalizations throughout (Ironwood bundles exist only in v6 transactions).

Each format carries its set of BLAKE2b personalization strings. Because `Bundle::commitment` now depends on the pool restrictions, it returns `Result<BundleCommitment, CommitmentError>` and fails with `CommitmentError::UnrepresentableFlags` when a bundle's flags cannot be encoded under the given pool (e.g. a cross-address-restricted bundle under a pre-NU6.3 pool).

New personalization strings:

- Orchard v6: `ZTxIdOrchardH_v6`, `ZTxAuthOrchaH_v6`
- Ironwood v6: `ZTxIdIronwd_H_v6`, `ZTxIdIrnActCH_v6`, `ZTxIdIrnActMH_v6`, `ZTxIdIrnActNH_v6`, `ZTxAuthIrnwdH_v6`

All constants should match:
- [ZIP 229](https://zips.z.cash/zip-0229) (Ironwood transaction format)
- https://github.com/zcash/zcash-test-vectors/pull/130
  - Note: these test vectors are in librustzcash only, not in Orchard

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -199,14 +199,12 @@ the current behavior by selecting `BundlePoolRestrictions::OrchardNu6_2Only` (an
     action count.
 - `orchard::pczt::Bundle::create_proof` now builds the Action circuits for
   the provided `ProvingKey`'s circuit version (previously always
-  `FixedPostNu6_2`), and checks the cross-address restriction's
-  same-expanded-receiver property, returning
-  `ProverError::DisallowedCrossAddressTransfer` (or
+  `FixedPostNu6_2`), and checks the cross-address restriction's same-expanded-receiver
+  property, returning `ProverError::DisallowedCrossAddressTransfer` (or
   `ProverError::MissingRecipient` if a `recipient` field is unset).
 - `orchard::pczt::{Spend, Output}::parse` now take the note plaintext version
   for the parsed spend or output, and `orchard::pczt::Bundle::parse` rejects
-  actions whose spend or output note version does not match the
-  `BundlePoolRestrictions`.
+  actions whose spend or output note version does not match the `BundlePoolRestrictions`.
 - `orchard::note_encryption::{OrchardNoteEncryption, IronwoodNoteEncryption}`
   documentation now clarifies that encryption uses the note's own
   `NoteVersion`; the aliases differ in which note plaintext versions they
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,12 @@ the current behavior by selecting `BundlePoolRestrictions::OrchardNu6_2Only` (an
   Variants: `OrchardPreNu6_2`, `OrchardNu6_2Only`, `OrchardNu6_3Onward`, and
   `IronwoodNu6_3Onward` (which shares the post-NU6.3 circuit and uses V3 note
   plaintexts).
+- `orchard::bundle::TxVersion`, the transaction version (`V5` or `V6`) a bundle's
+  commitments are computed for. At NU6.3 an Orchard bundle may be encoded in a v5
+  or a v6 transaction; the two use different commitment personalizations and place
+  the anchor in different digests (v5 in the transaction-ID effects, v6 in the
+  authorizing data). It is passed to `Bundle::commitment`,
+  `Bundle::authorizing_commitment`, and the `hash_bundle_*_empty` helpers.
 - `orchard::note::NoteVersion`, the note plaintext version selector used by
   low-level note constructors and accessors. `NoteVersion::V2` is the ZIP 212
   Orchard note plaintext format, and `NoteVersion::V3` is the quantum-recoverable
@@ -193,12 +199,14 @@ the current behavior by selecting `BundlePoolRestrictions::OrchardNu6_2Only` (an
     action count.
 - `orchard::pczt::Bundle::create_proof` now builds the Action circuits for
   the provided `ProvingKey`'s circuit version (previously always
-  `FixedPostNu6_2`), and checks the cross-address restriction's same-expanded-receiver
-  property, returning `ProverError::DisallowedCrossAddressTransfer` (or
+  `FixedPostNu6_2`), and checks the cross-address restriction's
+  same-expanded-receiver property, returning
+  `ProverError::DisallowedCrossAddressTransfer` (or
   `ProverError::MissingRecipient` if a `recipient` field is unset).
 - `orchard::pczt::{Spend, Output}::parse` now take the note plaintext version
   for the parsed spend or output, and `orchard::pczt::Bundle::parse` rejects
-  actions whose output note version does not match the `BundlePoolRestrictions`.
+  actions whose spend or output note version does not match the
+  `BundlePoolRestrictions`.
 - `orchard::note_encryption::{OrchardNoteEncryption, IronwoodNoteEncryption}`
   documentation now clarifies that encryption uses the note's own
   `NoteVersion`; the aliases differ in which note plaintext versions they
@@ -213,23 +221,29 @@ the current behavior by selecting `BundlePoolRestrictions::OrchardNu6_2Only` (an
     note version through their re-exported path.
   - `orchard::bundle::testing::{arb_action_n, arb_unauthorized_action_n}` now
     take a note version.
-  - `orchard::bundle::testing::{arb_bundle, arb_unauthorized_bundle}` now
-    generate a bundle pool restriction internally to select the note version.
+  - `orchard::bundle::testing::{arb_bundle, arb_unauthorized_bundle}` no longer
+    take bundle pool restrictions, and instead select them internally.
 - `orchard::pczt::Bundle::finalize_io` verifies the cross-address restriction
   before modifying the bundle, returning
   `IoFinalizerError::CrossAddressRestriction` (wrapping the underlying
   `VerifyError`) and leaving the bundle unmodified if the PCZT is missing
   recipient data or violates the restriction.
-- `orchard::Bundle::commitment` now takes the `BundlePoolRestrictions` the bundle
-  follows, and hashes that era's flag byte (via `Flags::to_byte`). The ZIP-244
-  Orchard digest — and therefore the transaction ID and sighash — now depends on
-  the `BundlePoolRestrictions`: under `IronwoodNu6_3Onward` an unrestricted bundle's
-  flag byte sets bit 2. Callers computing transaction IDs or sighashes (e.g.
-  `zcash_primitives`, or the `pczt` crate via `Flags::to_byte`) must pass the
-  protocol matching the transaction. It now returns
+- `orchard::Bundle::{commitment, authorizing_commitment}` now take a
+  `orchard::bundle::BundlePoolRestrictions` (which selects the flag-byte encoding) and
+  an `orchard::bundle::TxVersion` (which selects the commitment personalization
+  strings and the anchor placement for the transaction the bundle is encoded in). The
+  ZIP-244 digest — and therefore the transaction ID and sighash — now depends on both:
+  under a NU6.3 protocol an unrestricted bundle's flag byte sets bit 2, and
+  `TxVersion::V6` uses the v6 personalization strings and commits the anchor in the
+  authorizing commitment instead of the effects commitment. Callers computing
+  transaction IDs or sighashes must pass the restrictions and version matching the
+  transaction. `Bundle::commitment` now returns
   `Result<BundleCommitment, CommitmentError>`, returning
   `Err(CommitmentError::UnrepresentableFlags)` if the flags are unrepresentable
-  under that protocol (cross-address transfers disabled under a pre-NU6.3 protocol).
+  under those restrictions (for example, cross-address transfers disabled under
+  pre-NU6.3 restrictions).
+- `orchard::bundle::commitments::{hash_bundle_txid_empty, hash_bundle_auth_empty}`
+  now take a `BundlePoolRestrictions` and a `TxVersion`.
 - Behind the `unstable-voting-circuits` feature, `RandomSeed::rcm` is renamed to
   `RandomSeed::rcm_v2`, marking it as the rcm derivation for V2 (ZIP 212) notes.
   These APIs are not covered by the crate's semver guarantees.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -204,7 +204,7 @@ the current behavior by selecting `BundlePoolRestrictions::OrchardNu6_2Only` (an
   `ProverError::MissingRecipient` if a `recipient` field is unset).
 - `orchard::pczt::{Spend, Output}::parse` now take the note plaintext version
   for the parsed spend or output, and `orchard::pczt::Bundle::parse` rejects
-  actions whose spend or output note version does not match the `BundlePoolRestrictions`.
+  actions whose output note version does not match the `BundlePoolRestrictions`.
 - `orchard::note_encryption::{OrchardNoteEncryption, IronwoodNoteEncryption}`
   documentation now clarifies that encryption uses the note's own
   `NoteVersion`; the aliases differ in which note plaintext versions they
@@ -219,8 +219,8 @@ the current behavior by selecting `BundlePoolRestrictions::OrchardNu6_2Only` (an
     note version through their re-exported path.
   - `orchard::bundle::testing::{arb_action_n, arb_unauthorized_action_n}` now
     take a note version.
-  - `orchard::bundle::testing::{arb_bundle, arb_unauthorized_bundle}` no longer
-    take bundle pool restrictions, and instead select them internally.
+  - `orchard::bundle::testing::{arb_bundle, arb_unauthorized_bundle}` now
+    generate a bundle pool restriction internally to select the note version.
 - `orchard::pczt::Bundle::finalize_io` verifies the cross-address restriction
   before modifying the bundle, returning
   `IoFinalizerError::CrossAddressRestriction` (wrapping the underlying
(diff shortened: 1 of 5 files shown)


src/bundle/commitments.rs
```diff
@@ -5,10 +5,111 @@ use blake2b_simd::{Hash as Blake2bHash, Params, State};
 use crate::bundle::{Authorization, Authorized, Bundle, BundlePoolRestrictions};
 
 const ZCASH_ORCHARD_HASH_PERSONALIZATION: &[u8; 16] = b"ZTxIdOrchardHash";
```
ebfull: Should rename this to have `V5` in it, forcing every usage site to carry a hunk in this diff, which propagates review coverage of the API changes.

src/bundle/commitments.rs
```diff
@@ -5,10 +5,111 @@ use blake2b_simd::{Hash as Blake2bHash, Params, State};
...
+/// protocol personalization strings, flag-byte encoding format, and anchor
+/// placement. Use the associated constants to select the domain required by the
+/// transaction format.
+#[derive(Clone, Copy, Debug)]
+pub struct BundleCommitmentDomain {
+    personalizations: BundleCommitmentPersonalizations,
+    pool_restrictions: BundlePoolRestrictions,
+    effects_anchor: AnchorCommitment,
+    auth_anchor: AnchorCommitment,
+}
```
ebfull: This API is kind of bizarre to me. I'd rather rename `BundlePoolRestrictions` to `BundlePoolParameters` (sorry lmao) and add these details as inherent methods on that enum. You can still keep concerns separated by file of course, just import what you need _as locally as possible_ in `bundle.rs` to expose these parameters.

The result will be cleaner and have less code. This arrangement requires you to roundtrip through constants and structures that don't need to exist, just to end up with something that should be defined for all `BundlePoolRestrictions` instances (hence the rename.)

ebfull (changes requested):
Let's make this API nicer.

ValarDragon:
Updated, agreed prior API was not good.

My recollection was some decisions were made in Orchard to not expose it to V5/V6 tx formats (which is not possible for the hashing update)

I think the new bundle pool parameters implementation is far cleaner.

src/bundle.rs
```diff
@@ -822,8 +837,11 @@ impl<V> Bundle<Authorized, V> {
     /// Computes a commitment to the authorizing data within for this bundle.
     ///
     /// This together with `Bundle::commitment` bind the entire bundle.
-    pub fn authorizing_commitment(&self) -> BundleAuthorizingCommitment {
-        BundleAuthorizingCommitment(hash_bundle_auth_data(self))
+    pub fn authorizing_commitment(
```
czarcas7ic, line 852: Add to changelog due to changing public method.

ValarDragon:
Addressed all comments and combined into one commit!

src/bundle.rs
```diff
@@ -83,61 +87,79 @@ pub enum BundlePoolRestrictions {
...
-    /// prohibited; Orchard actions are disallowed in coinbase. Notes use V2 plaintexts.
-    OrchardNu6_3Onward,
-    /// The Ironwood pool at NU6.3 and later.
+    /// Uses the post-NU6.3 circuit and the NU6.3 flag-byte format. For
+    /// transactional bundles `enableCrossAddress = 0` is required by consensus,
+    /// so cross-address transfers are prohibited; Orchard actions are
+    /// disallowed in coinbase. Notes use V2 plaintexts.
+    OrchardV5Nu6_3Onward,
+    /// The Orchard pool in v6 transactions at NU6.3 and later.
     ///
```
ebfull, line 93: Ah, I wasn't clear enough and didn't anticipate this.

`BundlePoolRestrictions` today is a `(pool, era)` selector, which we should want to keep. `BundlePoolParameters` is a good rename of this because it should gain the methods for expressing parameters, but those will need to accept arguments to distinguish transaction versions, rather than encoding it into the variants. Otherwise all of the code that touches the pool (as a concept) must touch the tx version (which is a distinction none of the current code need to handle, nor should it).

`commitment` / `authorizing_commitment` should get a transaction version parameter.

src/bundle.rs
```diff
@@ -822,8 +841,11 @@ impl<V> Bundle<Authorized, V> {
     /// Computes a commitment to the authorizing data within for this bundle.
     ///
     /// This together with `Bundle::commitment` bind the entire bundle.
-    pub fn authorizing_commitment(&self) -> BundleAuthorizingCommitment {
-        BundleAuthorizingCommitment(hash_bundle_auth_data(self))
+    pub fn authorizing_commitment(
```
ebfull, line 852: This is a latent issue which @ValarDragon warned me of... the actual tests for this stuff exist outside of this crate currently, so these changes are not properly exercised here. I don't want to deal with that now, and I can tolerate finding out later that it didn't work, but we need to fix that problem _here_ eventually.

src/bundle/commitments.rs
```diff
@@ -2,13 +2,94 @@
...
+    };
+
+const ORCHARD_V6_PERSONALIZATIONS: BundleCommitmentPersonalizations =
+    BundleCommitmentPersonalizations {
+        bundle: ZCASH_ORCHARD_V6_HASH_PERSONALIZATION,
+        actions_compact: ZCASH_ORCHARD_ACTIONS_COMPACT_HASH_PERSONALIZATION,
+        actions_memos: ZCASH_ORCHARD_ACTIONS_MEMOS_HASH_PERSONALIZATION,
+        actions_noncompact: ZCASH_ORCHARD_ACTIONS_NONCOMPACT_HASH_PERSONALIZATION,
+        auth: ZCASH_ORCHARD_V6_SIGS_HASH_PERSONALIZATION,
+    };
```
ebfull, line 50: Spec note... The v6 personalizations are deliberately being kept the same as v5 in here? I think that's fine because of the top-level personalization, though it makes me wonder what the point of freshening the ironwood ones is? Why aren't they using the same ones that Orchard uses, and merely the bundles are different? If there's a good reason, why doesn't it apply to Orchard on v6?

At least we should comment about this above here.

CHANGELOG.md
```diff
@@ -136,9 +136,9 @@ the current behavior by selecting `BundlePoolRestrictions::OrchardNu6_2Only` (an
 - `orchard::builder::BundleType::Transactional` no longer embeds a full `Flags`;
   it carries `{ spends_enabled, outputs_enabled, bundle_required }`, and
   `BundleType::flags` and `BundleType::num_actions` now take the
-  `BundlePoolRestrictions`. The builder chooses the cross-address bit as a prover-side
-  default — the least-restrictive value consensus permits: enabled, except under
-  `OrchardNu6_3Onward`, where consensus mandates the restriction. `BundlePoolRestrictions`
+  `BundlePoolParameters`. The builder chooses the cross-address bit as a prover-side
+  default — the least-restrictive value consensus permits: enabled in 
+  `OrchardPreNu6_2` and `IronwoodNu6_3Onward`. `BundlePoolParameters`
```
czarcas7ic: Don't we need to mention `OrchardNu6_2Only` here?

tests/builder.rs
```diff
@@ -189,15 +186,11 @@ fn builder_builds_for_insecure_circuit_version() {
...
-        BundlePoolRestrictions::OrchardPreNu6_2,
-        SHIELDING,
-        recipient,
-    );
+    let builder = output_only_builder(BundlePoolParameters::OrchardPreNu6_2, SHIELDING, recipient);
 
     let (unauthorized, _) = builder.build::<i64>(&mut rng).unwrap().unwrap();
     let sighash: [u8; 32] = unauthorized
-        .commitment(BundlePoolRestrictions::OrchardPreNu6_2)
+        .commitment(BundlePoolParameters::OrchardNu6_2Only)
```
czarcas7ic: We build with `OrchardPreNu6_2` but commits with `OrchardNu6_2Only`, this is the only test to do this, I think they are supposed to match here.

ebfull (changes requested):

src/bundle/commitments.rs
ebfull: ~Wait... why are the personalization strings specific to the version anyway?~ Rediscovered why this is unavoidable...

src/bundle/commitments.rs
```diff
@@ -26,21 +107,18 @@ fn hasher(personal: &[u8; 16]) -> State {
 /// Then, hash these together along with (flags, value_balance_orchard, anchor_orchard),
```
czarcas7ic: I don't think we always hash together with anchor_orchard, specifically for v6 formats, right?

src/bundle/commitments.rs
```diff
@@ -26,21 +107,18 @@ fn hasher(personal: &[u8; 16]) -> State {
...
 /// personalized with ZCASH_ORCHARD_ACTIONS_HASH_PERSONALIZATION
 ///
-/// # Errors
-///
-/// Returns `None` if `bundle`'s flags cannot be encoded under `pool_restrictions`
-/// (cross-address transfers disabled for a bundle targetted to pre-NU6.3 Orchard);
-/// such a bundle cannot appear in a pre-NU6.3 transaction.
+/// Returns `None` if the bundle flags cannot be encoded in the domain's bundle format.
```
czarcas7ic: Consider:

```
/// Then, hash these together with (flags, value_balance_orchard), plus
/// anchor_orchard for v5 formats only. v6/Ironwood omit the anchor here and commit to it
/// in the auth hash, personalized with the pool's bundle personalization.
///
/// Returns `None` if the bundle flags cannot be encoded under `pool_parameters`.
```

src/bundle.rs
```diff
@@ -300,29 +321,29 @@ impl Flags {
...
+        assert!(Self::from_byte(value, pool_parameters) == Some(*self));
         Some(value)
     }
 
     /// Parses flags from a single byte as defined in [Zcash Protocol Spec § 7.1:
     /// Transaction Encoding And Consensus][txencoding], according to the interpretation
-    /// implied by `pool_restrictions`. Returns `None` if unexpected bits are set in the flag byte.
+    /// implied by `pool_parameters`. Returns `None` if unexpected bits are set in the flag byte.
     ///
     /// The protocol specification and ZIPs 225 and TODO:draft-zodl-valargroup-ironwood-txformat
```
czarcas7ic: I think this was assigned ZIP 229 recently

src/bundle.rs
ebfull: I've decided the `BundlePoolParameters` rename is good, but I want to make the diff slightly smaller until the API surface is decided. We can punt it to another PR perhaps. For now, I want `BundlePoolRestrictions` to still take the role of determining personalizations (to unify) using a new `TxVersion` threaded through by the caller, but not to do the rename yet.

tests/builder.rs
ebfull: Removed it.

CHANGELOG.md
ebfull: Removed.

src/bundle/commitments.rs
```diff
@@ -2,13 +2,121 @@
...
+const ZCASH_IRONWOOD_ACTIONS_COMPACT_HASH_PERSONALIZATION: &[u8; 16] = b"ZTxIdIrnActCH_v6";
+const ZCASH_IRONWOOD_ACTIONS_MEMOS_HASH_PERSONALIZATION: &[u8; 16] = b"ZTxIdIrnActMH_v6";
+const ZCASH_IRONWOOD_ACTIONS_NONCOMPACT_HASH_PERSONALIZATION: &[u8; 16] = b"ZTxIdIrnActNH_v6";
+const ZCASH_IRONWOOD_SIGS_HASH_PERSONALIZATION: &[u8; 16] = b"ZTxAuthIrnwdH_v6";
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum AnchorCommitment {
+    Include,
+    Omit,
+}
```
ebfull: Why can't this be a `bool`?

src/bundle/commitments.rs
```diff
@@ -2,13 +2,121 @@
...
+    }
+
+    fn auth_anchor_commitment(self) -> AnchorCommitment {
+        match self {
+            BundleCommitmentFormat::OrchardV5 => AnchorCommitment::Omit,
+            BundleCommitmentFormat::OrchardV6 | BundleCommitmentFormat::IronwoodV6 => {
+                AnchorCommitment::Include
+            }
+        }
+    }
```
ebfull: These have confusing names because "effects" can be misinterpreted as "affects" which has a semantic overload, "anchor" has two possible definitions (one is the actual anchor, one is a verb being anchored again semantically different) etc.

ebfull (changes requested):

src/bundle/commitments.rs
```diff
@@ -2,13 +2,108 @@
...
+    fn includes_anchor_in_txid_digest(self) -> bool {
+        matches!(self, BundleCommitmentFormat::OrchardV5)
+    }
+
+    fn includes_anchor_in_authorizing_digest(self) -> bool {
+        matches!(
+            self,
+            BundleCommitmentFormat::OrchardV6 | BundleCommitmentFormat::IronwoodV6
+        )
+    }
```
ebfull, line 105: We _could_ just define these in terms of each other, but I'll let someone else insist on it first.

ValarDragon:
Changes look great to me! Looks way better with tx version being explicit

ebfull (approved):

src/builder.rs
```diff
@@ -167,7 +167,7 @@ impl BundleType {
     }
 }
 
-/// Decide, for a bundle targetting the consensus rules implied by `pool_restrictions`,
+/// Decide, for a bundle targeting the consensus rules implied by `pool_restrictions`,
```
ebfull, line 170: sorry daira

CHANGELOG.md
```diff
@@ -220,16 +226,22 @@ the current behavior by selecting `BundlePoolRestrictions::OrchardNu6_2Only` (an
...
+  under a NU6.3 protocol an unrestricted bundle's flag byte sets bit 2, and
+  `TxVersion::V6` uses the v6 personalization strings and commits the anchor in the
+  authorizing commitment instead of the effects commitment. Callers computing
+  transaction IDs or sighashes must pass the restrictions and version matching the
+  transaction. `Bundle::commitment` now returns
   `Result<BundleCommitment, CommitmentError>`, returning
   `Err(CommitmentError::UnrepresentableFlags)` if the flags are unrepresentable
-  under that protocol (cross-address transfers disabled under a pre-NU6.3 protocol).
+  under those restrictions (for example, cross-address transfers disabled under
+  pre-NU6.3 restrictions).
```
TalDerei, line 243: ACK – this API makes more sense to me than bundling `BundlePoolRestrictions` + `TxVersion` into a `BundleCommitmentDomain` (pool/era and the tx version are orthogonal concerns). maybe the changelog should be stricter in noting that there's no validation that catches them not matching.

src/bundle/commitments.rs
```diff
@@ -23,24 +118,25 @@ fn hasher(personal: &[u8; 16]) -> State {
...
+/// personalization string. In the v6 format the anchor is included by
+/// `hash_bundle_auth_data` instead.
 ///
-/// Returns `None` if `bundle`'s flags cannot be encoded under `pool_restrictions`
-/// (cross-address transfers disabled for a bundle targetted to pre-NU6.3 Orchard);
-/// such a bundle cannot appear in a pre-NU6.3 transaction.
+/// Returns `None` if the bundle flags cannot be encoded in the domain's bundle format.
 ///
 /// [zip244]: https://zips.z.cash/zip-0244
 pub(crate) fn hash_bundle_txid_data<A: Authorization, V: Copy + Into<i64>>(
```
TalDerei: "Orchard shielded action" doc comment is stale.

TalDerei (approved):
checked against [ZIP 229](https://github.com/zcash/zips/pull/1313) and generated some of my own local test vectors for redundancy against https://github.com/zcash/zcash-test-vectors/pull/130.

ebfull (approved):

PR #513, opened 2026-06-26
Reject Ironwood commitments for v5 transactions

ebfull:
Previously `commitment_format` ignored `tx_version` for the Ironwood pool, silently committing an Ironwood bundle under v6 personalization even when the caller requested `TxVersion::V5`. That should instead _fail_ because it's unrepresentable.

This makes `authorizing_commitment` and the `hash_bundle_*_empty` helpers fallible, mirroring `commitment`, and threads `CommitmentError` through the internal `hash_bundle_*_data` helpers. We add a `CommitmentError::InvalidTransactionVersion` variant and return it from the commitment APIs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,9 +46,11 @@ when building proving/verifying keys).
     introspection for whether a circuit version (or a key's circuit version)
     constrains the `disableCrossAddress` public input.
   - `orchard::circuit::VerifyingKey::circuit_version`
-- `orchard::bundle::CommitmentError`, with its `UnrepresentableFlags` variant,
-  returned by `orchard::Bundle::commitment` when a bundle's flags cannot be
-  represented under the requested `BundlePoolRestrictions`.
+- `orchard::bundle::CommitmentError`, with its `UnrepresentableFlags` and
+  `InvalidTransactionVersion` variants, returned by bundle commitment APIs when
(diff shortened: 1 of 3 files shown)


TalDerei (approved):
yeah the fallibility here seems good, no objection.

ValarDragon (approved):
This works for me!

As a caller I've seen no realistic risk of this though. The real risk we have in calling, is confusion over which V5 Orchard choice you do in commitments, because they are all equivalent:
- Pre NU6_2 , V5
- NU6_2, V5
- NU6_3 V5

ebfull:
This isn't a big deal, it's just something I wish we would have done in #510 in retrospect. I also expect it to be identified in audits; it was identified by `zkao`.

Technically, this is fixing a case where we actually return an incorrect value for an unrepresentable input. The cases you mention are all possible mistakes the caller can make, but at least the library cannot produce completely invalid results in those cases.

Probably the weirder examples are ones like @TalDerei suggested to me, such as `OrchardPreNu6_2` with `v6`, which aren't possible _in practice_ (due to separate consensus restrictions) but still have coherent and specified results in the digest computation.

TalDerei:
> it was identified by `zkao`

forgot to mention, v12 didn't surface anything new (only some duplicative informational findings that are outside our threat model).

PR #516, opened 2026-06-26
Add support for the Ironwood shielded pool.

ebfull:
**❗ Do not merge into `main` until `0.15.0` is released on `crates.io` and the commit is tagged as `0.15.0`.**

This is the release PR for `0.15.0`, which features NU6.3-related changes including the new circuit and pool APIs for Ironwood. This tracks the [`feat/ironwood`](https://github.com/zcash/orchard/tree/feat/ironwood) branch, which is currently open for PRs.

We are doing pre-release testing first. We'll be releasing `0.15.0-pre.N` from this branch until ZIPs, specs, audits and other internal reviews confirm we're ready for an `orchard` crate release.

PR #519, opened 2026-06-29
Distinguish value pool, protocol version, and bundle version

nuttycom:
Stacked on `feat/ironwood`.

Replaces `BundlePoolRestrictions` with `BundleVersion` — the `(ValuePool, ProtocolVersion)` of an Orchard bundle — and makes each `Bundle` carry its version as non-serialized context, so bundles are encodable and committable by construction.

## Summary
- New top-level `ValuePool` and `ProtocolVersion`; `BundleVersion` with safe-by-construction `const fn` constructors (`orchard_insecure_v0` / `orchard_v1` / `orchard_v2` / `ironwood_v2`).
- `Bundle` carries its `BundleVersion`; construction (`from_parts` / `try_from_parts`, builder, PCZT) validates flags against it (`BundleError::UnrepresentableFlags`); `from_parts` is now fallible.
- `commitment` / `authorizing_commitment` and the `decrypt_*` / `recover_*` helpers drop their version argument; new `flag_byte()` / `bundle_version()` accessors.
- `CommitmentError::UnrepresentableFlags` removed; proof-size enforcement derived from the version, `ProofSizeEnforcement` removed.
- CHANGELOG `[Unreleased]` updated to the final API.

Tests: 138 lib + 8 integration + 1 doctest, all passing.

---
Opened as a draft for review. Filed with Claude Code assistance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,26 +7,34 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
-This release introduces `orchard::bundle::BundlePoolRestrictions`, a `(pool, era)`
-selector now threaded through bundle construction, wire encoding, and commitments,
-alongside the post-NU 6.3 Action circuit that enforces the cross-address restriction.
-Existing callers keep the current behavior by selecting
-`BundlePoolRestrictions::OrchardNu6_2Only` (and `OrchardCircuitVersion::FixedPostNu6_2`
-when building proving/verifying keys).
+This release introduces `orchard::bundle::BundleVersion`, the `(value pool, protocol
+version)` of an Orchard bundle, built from the new `orchard::ValuePool` and
+`orchard::ProtocolVersion` types. Each `Bundle` now carries its `BundleVersion` as
+non-serialized context, so a bundle can be serialized and committed to without separately
+supplying a — possibly mismatching — version, and is encodable and committable by
+construction. The post-NU 6.3 Action circuit enforces the cross-address restriction.
+Existing callers keep the current behavior by constructing bundles with
+`BundleVersion::orchard_v1()` (and `OrchardCircuitVersion::FixedPostNu6_2` when building
+proving/verifying keys).
 
 ### Added
-- NU6.3 and Ironwood bundle-selection APIs:
-  - `orchard::bundle::BundlePoolRestrictions`, the `(pool, era)` selector for an
-    Orchard bundle. It determines the note plaintext version
-    (`BundlePoolRestrictions::note_version`), the circuit version
-    (`BundlePoolRestrictions::circuit_version`, when the `circuit` feature is
-    enabled), the flag-byte interpretation (pre-NU6.3 rules, where bit 2 is
-    reserved and cross-address transfers are implicitly enabled, vs NU6.3 rules,
-    where bit 2 is `enableCrossAddress`), and whether consensus mandates the
-    cross-address restriction (the builder then chooses the value within that
-    constraint). Variants: `OrchardPreNu6_2`, `OrchardNu6_2Only`,
-    `OrchardNu6_3Onward`, and `IronwoodNu6_3Onward` (which shares the post-NU6.3
-    circuit and uses V3 note plaintexts).
+- NU6.3 and Ironwood bundle-version APIs:
+  - `orchard::ValuePool`, the value pool an Orchard bundle belongs to (`Orchard` or
+    `Ironwood`), and `orchard::ProtocolVersion`, the Orchard protocol version
+    (`InsecureV0`, the historical pre-NU6.2 protocol that uses the unsound circuit; `V1`,
+    NU6.2; `V2`, NU6.3, which also instantiates the Ironwood pool).
+  - `orchard::bundle::BundleVersion`, the `(value pool, protocol version)` of an Orchard
+    bundle. Its `const fn` constructors `orchard_insecure_v0`, `orchard_v1`, `orchard_v2`,
+    and `ironwood_v2` make only the valid combinations representable. It determines the
+    note plaintext version (`BundleVersion::note_version`), the circuit version
+    (`BundleVersion::circuit_version`, when the `circuit` feature is enabled), the
+    flag-byte interpretation (pre-NU6.3 rules, where bit 2 is reserved and cross-address
+    transfers are implicitly enabled, vs NU6.3 rules, where bit 2 is `enableCrossAddress`),
+    and whether consensus mandates the cross-address restriction (the builder then chooses
+    the value within that constraint). `BundleVersion::value_pool` and
+    `BundleVersion::protocol_version` return the bundle's `ValuePool` and
+    `ProtocolVersion`; the Ironwood pool (`ironwood_v2`) shares the post-NU6.3 circuit and
+    uses V3 note plaintexts.
   - `orchard::bundle::TxVersion`, the transaction version (`V5` or `V6`) a
     bundle's commitments are computed for. At NU6.3 an Orchard bundle may be
     encoded in a v5 or a v6 transaction; the two use different commitment
@@ -46,11 +54,12 @@ when building proving/verifying keys).
     introspection for whether a circuit version (or a key's circuit version)
     constrains the `disableCrossAddress` public input.
   - `orchard::circuit::VerifyingKey::circuit_version`
-- `orchard::bundle::CommitmentError`, with its `UnrepresentableFlags` and
-  `InvalidTransactionVersion` variants, returned by bundle commitment APIs when
-  a bundle's flags cannot be represented under the requested
-  `BundlePoolRestrictions`, or when `IronwoodNu6_3Onward` is requested for a v5
-  transaction.
+- `orchard::bundle::CommitmentError`, with its `InvalidTransactionVersion` variant,
+  returned by the bundle commitment APIs when an Ironwood bundle's commitment is requested
+  for a `TxVersion::V5` transaction.
+- `orchard::Bundle::bundle_version`, returning the `BundleVersion` the bundle is encoded
+  under, and `orchard::Bundle::flag_byte`, the infallible byte encoding of the bundle's
+  flags under that version.
 - `orchard::bundle::BatchError` (requires the `circuit` feature), with its
   `RestrictionUnsupportedByKey` variant, returned by
   `orchard::bundle::BatchValidator::add_bundle` when a restricted bundle is added
@@ -81,6 +90,9 @@ when building proving/verifying keys).
 - PCZT note-version and cross-address APIs:
   - `orchard::pczt::{Spend, Output}::note_version`, the generated getters for the
     note plaintext version of a parsed spend or output.
+  - `orchard::pczt::Bundle::bundle_version`, the generated getter for the bundle's
+    `BundleVersion`, and `orchard::pczt::Bundle::flag_byte`, the infallible byte encoding of
+    its flags under that version.
   - `orchard::pczt::Bundle::verify_cross_address_restriction`, so that Signers can
     check the cross-address restriction's same-expanded-receiver structural
     property before signing. It is a no-op for bundles that permit cross-address
@@ -92,21 +104,21 @@ when building proving/verifying keys).
     underlying `orchard::pczt::VerifyError`.
 
 ### Changed
-- Bundle construction now requires explicit pool restrictions:
+- Bundle construction now requires an explicit `BundleVersion`:
   - `orchard::builder::Builder::new` now takes
-    `(BundlePoolRestrictions, BundleType, Anchor)`; the builder derives the circuit
-    version from the pool restrictions rather than from an explicit
+    `(BundleVersion, BundleType, Anchor)`; the builder derives the circuit
+    version from the bundle version rather than from an explicit
     `OrchardCircuitVersion`.
-  - `orchard::builder::bundle` now takes a `BundlePoolRestrictions` in place of
+  - `orchard::builder::bundle` now takes a `BundleVersion` in place of
     the circuit-version argument, and takes the wallet-controlled change outputs
     as a separate `changes: Vec<ChangeInfo>` argument (plain `outputs` and
     `changes` are distinct). It rejects supplied `OutputInfo`/`ChangeInfo` values
-    whose note version does not match the `BundlePoolRestrictions`, returning
+    whose note version does not match the `BundleVersion`, returning
     `BuildError::InvalidNoteVersion`.
   - `orchard::builder::BundleType::Transactional` no longer embeds a full
     `Flags`; it now carries `{ spends_enabled, outputs_enabled, bundle_required }`.
   - `orchard::builder::BundleType::{num_actions, flags}` now take a
-    `BundlePoolRestrictions`. For bundles that disable cross-address transfers,
+    `BundleVersion`. For bundles that disable cross-address transfers,
     `num_actions` counts `num_spends + num_outputs` requested actions (a requested
     spend and a requested output never share an action) rather than the maximum of
     the two, and `BundleMetadata` maps them to distinct actions; wallets
@@ -114,10 +126,10 @@ when building proving/verifying keys).
   - `orchard::builder::OutputInfo::{new, dummy}` now take an `orchard::NoteVersion`,
     so callers choose between V2 Orchard notes and V3 Ironwood notes;
     builder-created outputs use the note version associated with the selected
-    `BundlePoolRestrictions`.
+    `BundleVersion`.
   - `orchard::builder::BundleMetadata::output_action_index` now indexes the plain
     outputs first, followed by the wallet-controlled change outputs.
-- For `BundlePoolRestrictions::OrchardNu6_3Onward`, the builder constructs
+- For `BundleVersion::orchard_v2()`, the builder constructs
   withdrawal/change bundles that disable cross-address transfers: every action's
   output is addressed to the expanded receiver of the note it spends. The
   fabricated zero-value output paired with each real spend carries a randomized,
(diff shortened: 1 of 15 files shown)


src/bundle.rs
```diff
@@ -59,85 +59,114 @@ impl<T> Action<T> {
...
     ///
-    /// `OrchardNu6_3Onward` mandates this restriction; every other value of
-    /// `BundlePoolRestrictions` leaves it free. This is not necessarily the same
+    /// The [`ValuePool::Orchard`] pool under [`ProtocolVersion::V2`] mandates this restriction;
+    /// every other value of `BundleVersion` leaves it free. This is not necessarily the same
     /// cross-address-enabled decision the [`Builder`](crate::builder::Builder) makes;
     /// that is determined by builder policy chosen within this constraint.
-    pub(crate) fn requires_cross_address_restriction(self) -> bool {
-        matches!(self, BundlePoolRestrictions::OrchardNu6_3Onward)
+    pub(crate) fn requires_cross_address_restriction(&self) -> bool {
```
nuttycom: In looking at the use sites for this method, it would be better to invert its sense; this should be `bundle_version.permits_cross_address_transfers()`

src/bundle/commitments.rs
```diff
@@ -221,11 +218,12 @@ pub(crate) fn hash_bundle_auth_data<V>(
 ///
 /// [zip244]: https://zips.z.cash/zip-0244
 pub fn hash_bundle_auth_empty(
-    pool_restrictions: BundlePoolRestrictions,
+    bundle_version: BundleVersion,
```
nuttycom: This should now just take the value pool instead of the bundle version.

src/lib.rs
```diff
@@ -117,3 +117,42 @@ impl Proof {
         BASE + PER_ACTION * num_actions
     }
 }
+
+/// The set of value pools supported by the Orchard protocol.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ValuePool {
```
ebfull, line 123: This should be `#[non_exhaustive]` to match `BundleVersion`.

src/lib.rs
```diff
@@ -117,3 +117,42 @@ impl Proof {
...
+    /// The Orchard value pool.
+    Orchard,
+    /// The Ironwood value pool.
+    Ironwood,
+}
+
+/// The versions of the Orchard protocol.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ProtocolVersion {
```
ebfull, line 132: Ditto re: `#[non_exhaustive]`.

src/bundle.rs
```diff
@@ -55,89 +55,118 @@ impl<T> Action<T> {
...
+        Self {
+            value_pool: ValuePool::Ironwood,
+            protocol_version: ProtocolVersion::V2,
+        }
+    }
+
+    /// Returns the [`ValuePool`] to which these pool restrictions apply.
+    pub fn value_pool(&self) -> ValuePool {
+        self.value_pool
+    }
```
ebfull: Consider a `protocol_version` accessor.

src/bundle.rs
```diff
@@ -149,7 +178,7 @@ impl BundlePoolRestrictions {
 /// includes it in the authorizing digest. Ironwood bundles exist only in v6 transactions, so
 /// attempting to compute an Ironwood commitment for a v5 transaction returns an error.
 ///
-/// This is independent of the [`BundlePoolRestrictions`] that govern construction: the same
+/// This is independent of the [`BundleVersion`] that govern construction: the same
 /// Orchard bundle can be committed under either version, and the caller must pass the one
```
ebfull:
```suggestion
/// This is independent of the [`BundleVersion`] that governs construction: the same
/// Orchard bundle can be committed under either transaction version, and the caller must pass the one
```

src/bundle.rs
```diff
@@ -162,7 +191,11 @@ pub enum TxVersion {
     V6,
 }
 
-/// Orchard-specific flags.
+/// Flags denoting what operations may be performed by the Orchard actions
+/// in a bundle.
+///
+/// This type is intended to be treated as safe-by-construction for a given
+/// value pool and protocol version.
```
ebfull: `Flags` itself is version-agnostic, so it cannot be "safe-by-construction" in the literal sense described here.

src/bundle.rs
```diff
@@ -55,89 +55,118 @@ impl<T> Action<T> {
 /// (Orchard or Ironwood) and the epoch range (pre-NU6.2, NU6.2 only, NU6.3 onward)
 /// that it targets. The Ironwood pool is only allowed for NU6.3 onward.
```
ebfull: This is now stale.

src/builder.rs
ebfull: "restrictions" is stale here and in some other places. Here's a list of some of them:

Doc comments — reword "pool restriction(s)" → "bundle version":
- [ ] src/bundle.rs:115 — value_pool(): "Returns the [ValuePool] to which these pool restrictions apply." → "…this bundle version applies."
- [ ] src/bundle.rs:121 — circuit_version(): "…consistent with these pool restrictions." → "…with this bundle version."
- [ ] src/bundle.rs:136 — note_version(): "The [NoteVersion] associated with this bundle pool restriction." → "…with this bundle version."
- [ ] src/bundle.rs:232 — Flags::from_parts: "…prover-side default for the pool restrictions (see …)" → "…for the bundle version".
- [ ] src/builder.rs:71 — BundleType::Coinbase: "…enabled if possible for this bundle's pool restrictions." (already partly fixed at :57) → "…for this bundle's version."
- [ ] src/builder.rs:204 — BuildError::InvalidNoteVersion doc: "…inconsistent with the bundle pool restrictions." → "…with the bundle version."
- [ ] src/builder.rs:792 — Builder::note_version(): "…associated with this builder's pool restrictions." → "…this builder's bundle version."
- [ ] src/builder.rs:800 — add_spend: "…trial-decrypting an [Action] under the bundle's pool restrictions." → "…under the bundle's version."
- [ ] src/builder.rs:973 — Builder::build: "…circuit version consistent with the builder's pool restrictions." → "…with the builder's bundle version."
- [ ] src/pczt/prover.rs:26 and :30 — ProverError::DisallowedCrossAddressTransfer / ProofFailed docs: "…the bundle's pool restrictions disable cross-address transfers…" → "…the bundle disables cross-address transfers…"

User-facing error message strings (higher priority — emitted at runtime):
- [ ] src/builder.rs:240 — BuildError::InvalidNoteVersion Display: "…does not match the bundle pool restrictions." → "…the bundle version."
- [ ] src/pczt/prover.rs:191 — ProverError Display: "…the bundle's pool restrictions disable cross-address transfers"
- [ ] src/pczt/verify.rs:231 — VerifyError Display: "…the bundle's pool restriction disables cross-address transfers"
  - Note: these two also disagree (plural "restrictions disable" vs singular "restriction disables"), and both attribute the disable to the version when it's really a flags property — consider "…but the bundle disables cross-address transfers."

Plain comments:
- [ ] src/builder.rs:1852 — "…Pool restrictions leave the choice free everywhere except Orchard post-NU6.3…" → "The bundle version leaves the choice free…"
- [ ] src/builder.rs:1947 — "…as a non-coinbase transaction would for the same pool restrictions." → "…for the same bundle version."

arb_bundle_pool_restriction testing strategy (rename, not just reword — returns BundleVersion):
- [ ] src/bundle.rs:1050 — doc "Create an arbitrary bundle pool restriction." → "Create an arbitrary [BundleVersion]."
- [ ] src/bundle.rs:1051 — pub fn arb_bundle_pool_restriction() → arb_bundle_version
- [ ] src/bundle.rs:1166, src/bundle.rs:1196 — update the two call sites
- [ ] CHANGELOG.md:253 — update the strategy name reference to match

ebfull (commented):
Comments from an LLM review of this draft (044eec15c83e8e9e65a1f0edd75a73566e47903c).

CHANGELOG.md
```diff
@@ -245,10 +264,13 @@ when building proving/verifying keys).
     (ZIP 2005, Ironwood) notes.
 
 ### Removed
+- `orchard::bundle::ProofSizeEnforcement`; `Bundle::try_from_parts` now derives the
+  canonical proof-size check from the `BundleVersion` (enforced for every version except
+  `BundleVersion::orchard_insecure_v0`).
 - `orchard::builder::Builder::new_for_version`; use
-  `Builder::new(pool_restrictions, bundle_type, anchor)`.
+  `Builder::new(bundle_version, bundle_type, anchor)`.
```
nuttycom: Was this constructor introduced since the 0.14 release? If so, it shouldn't be listed as removed; the actual removed constructor should be listed.

ebfull: Yeah, `new_for_version` was indeed removed since 0.14.

src/builder.rs
```diff
@@ -37,55 +37,36 @@ use {
...
+/// to the builder (see [`Builder::new`] and [`BundleVersion::default_flags`]).
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum BundleType {
     /// A transactional bundle will be padded if necessary to contain at least 2 actions,
     /// irrespective of whether any genuine actions are required.
     Transactional {
-        /// Whether Orchard spends are enabled for the bundle.
-        spends_enabled: bool,
-        /// Whether Orchard outputs are enabled for the bundle.
-        outputs_enabled: bool,
```
nuttycom, line 48: This removal is one thing that I'm still unsure about - I don't like having the flags duplicated here, so I removed the duplication.

src/builder.rs
```diff
@@ -832,7 +801,7 @@ impl Builder {
         note: Note,
         merkle_path: MerklePath,
     ) -> Result<(), SpendError> {
-        let flags = self.bundle_type.flags(self.pool_restrictions);
+        let flags = self.flags;
         if !flags.spends_enabled() {
```
nuttycom: nit: 
```suggestion
        if !self.flags.spends_enabled() {
```
and so below

src/lib.rs
nuttycom: I like `#[non_exhaustive]` for error types, but since `BundleVersion` is no longer an enum, I don't think this applies?

src/lib.rs
nuttycom: Same: https://github.com/zcash/orchard/pull/519#discussion_r3493698355

nuttycom:
[force-pushed](https://github.com/zcash/orchard/compare/044eec1..cbeddd3520af3b3da85e2afe7d0b1b6424c37a53) to address comments from code review.

src/lib.rs
ebfull: Agreed (that review comment was from Claude) in my view this should _not_ be non_exhaustive, we want to break everything that matches on it if we change this someday.

src/builder.rs
```diff
@@ -37,55 +37,36 @@ use {
...
         bundle_required: false,
     };
 
-    /// The DISABLED bundle type does not permit any bundle to be produced, and when used in the
-    /// builder will prevent any spends or outputs from being added.
-    pub const DISABLED: BundleType = BundleType::Transactional {
-        spends_enabled: false,
-        outputs_enabled: false,
-        bundle_required: false,
-    };
```
ebfull, line 85: This was removed but no `Flags::DISABLED` exists. This also makes the change awkward to describe in the `CHANGELOG.md`, which also incorrectly omits the removal of this constant.

src/builder.rs
```diff
@@ -772,46 +722,64 @@ impl BundleMetadata {
...
+    /// Returns [`BuildError::UnrepresentableFlags`] if `flags` cannot be encoded under
+    /// `bundle_version`, or [`BuildError::CoinbaseSpendsEnabled`] if `bundle_type` is
+    /// [`BundleType::Coinbase`] but `flags` enable spends.
     pub fn new(
-        pool_restrictions: BundlePoolRestrictions,
         bundle_type: BundleType,
+        bundle_version: BundleVersion,
+        flags: Flags,
         anchor: Anchor,
-    ) -> Self {
```
ebfull, line 795: The ordering of args here is different from `fn bundle` below, not sure which ordering you wanted to do, but they should both be consistent.

ebfull (changes requested):
Should do a couple fixes including fixing doc links, but ACK otherwise.

src/builder.rs
nuttycom: This is the correct order, we'll fix the `fn bundle` below.

src/builder.rs
nuttycom: I don't understand this comment. What's wrong with the removal of this constant? Is it just that it needs to be documented in the CHANGELOG?

ebfull: Yeah, I wasn't sure why we had this to begin with, and removing it entirely is a good thing.

ebfull (approved):

ebfull:
Merging, the tests will rerun on #516 after the merge and before the next prerelease.

src/pczt/tx_extractor.rs
```diff
@@ -192,6 +198,10 @@ impl fmt::Display for TxExtractorError {
                 f,
                 "Orchard `zkproof` has non-canonical length {actual}; expected {expected} bytes",
             ),
+            TxExtractorError::UnrepresentableFlags => write!(
+                f,
+                "Orchard bundle flags are not representable under its value pool and protocol version",
```
daira, line 203: "Orchard" in this error message refers to the Orchard protocol, which could be misleading if the problem was for an Ironwood-pool bundle. But in practice Ironwood will allow all of the flag combinations, so that case won't happen. At some point I will do a full pass for accuracy of terminology, but it doesn't need to block anything.

src/builder.rs
daira: I think this is a good change.

src/builder.rs
```diff
@@ -1718,8 +1703,14 @@ pub mod testing {
         /// Create a bundle from the set of arbitrary bundle inputs.
         fn into_bundle<V: TryFrom<i64>>(mut self) -> Bundle<Authorized, V> {
             let fvk = FullViewingKey::from(&self.sk);
-            let pool_restrictions = BundlePoolRestrictions::OrchardNu6_2Only;
-            let mut builder = Builder::new(pool_restrictions, BundleType::DEFAULT, self.anchor);
+            let bundle_version = BundleVersion::orchard_v2();
```
daira, line 1706: This misses test coverage of the Ironwood pool, or of other Orchard versions.

src/builder.rs
```diff
@@ -1849,61 +1840,76 @@ mod tests {
...
-        // for every variant but that one.
-
-        // Orchard pre-NU6.3: free, so the default is enabled.
-        for pool_restrictions in [
-            BundlePoolRestrictions::OrchardPreNu6_2,
-            BundlePoolRestrictions::OrchardNu6_2Only,
+    fn default_flags_match_pool_policy() {
+        // The builder's default flags enable spends and outputs and leave cross-address transfers
+        // as permissive as consensus allows. The bundle version leaves the cross-address choice free
+        // everywhere except Orchard from NU6.3 onward, where it is mandatorily disabled.
```
daira, line 1851: I think this is incorrect; the cross-address choice is not free for Orchard pre-NU6.3. Cross-address must be allowed there. So the comment is wrong and the test coverage that would show it to be wrong is missing: we are only testing the default flags here, not whether the choice of flags is free.

src/bundle.rs
```diff
@@ -51,93 +51,139 @@ impl<T> Action<T> {
...
-    pub fn note_version(self) -> NoteVersion {
-        use BundlePoolRestrictions::*;
-        match self {
-            OrchardPreNu6_2 | OrchardNu6_2Only | OrchardNu6_3Onward => NoteVersion::V2,
-            BundlePoolRestrictions::IronwoodNu6_3Onward => NoteVersion::V3,
+    pub fn note_version(&self) -> NoteVersion {
+        match self.value_pool {
+            ValuePool::Orchard => NoteVersion::V2,
+            ValuePool::Ironwood => NoteVersion::V3,
         }
```
daira, line 149: Grump: I dislike note plaintext lead bytes being referred to as "versions"; the specs never do that. "Version" is far too overloaded (and we just added yet another kind of "version" for bundles).

Pretty please can we stop naming things completely differently than the specs? Even just naming the variants `Byte0x02` and `Byte0x03` would be better. My eyes are bleeding with all these `V...` that mean different things.

src/bundle.rs
```diff
@@ -51,93 +51,139 @@ impl<T> Action<T> {
...
+    /// Every version permits them except the [`ValuePool::Orchard`] pool under
+    /// [`ProtocolVersion::V3`], which mandates the cross-address restriction. This is not
+    /// necessarily the same cross-address-enabled decision the
+    /// [`Builder`](crate::builder::Builder) makes; that is builder policy chosen within this
+    /// constraint.
+    pub(crate) fn permits_cross_address_transfers(&self) -> bool {
+        !matches!(
+            (self.protocol_version, self.value_pool),
+            (ProtocolVersion::V3, ValuePool::Orchard)
+        )
```
daira, line 164: Nit: this could be written more clearly and concisely as `self != Self::orchard_v3()`.

src/bundle.rs
```diff
@@ -162,7 +208,13 @@ pub enum TxVersion {
     V6,
 }
 
-/// Orchard-specific flags.
+/// Flags denoting what operations may be performed by the Orchard actions
```
daira, line 211:
```suggestion
/// Flags denoting what operations may be performed by the Orchard-protocol actions
```

src/lib.rs
```diff
@@ -117,3 +117,42 @@ impl Proof {
...
+    /// The original version of the protocol, used in Zcash prior to NU6.2, only instantiated for
+    /// the Orchard value pool.
+    ///
+    /// Uses the historical unsound Orchard circuit. Cross-address transfers are permitted and
+    /// notes use the V2 plaintext format. Used to reconstruct the historical verifying key and to
```
daira, line 137:
```suggestion
    /// notes use the lead-byte 0x02 plaintext format. Used to reconstruct the historical verifying key and to
```

src/lib.rs
```diff
@@ -117,3 +117,42 @@ impl Proof {
...
+    /// parse/verify historical bundles, not to build new ones.
+    InsecureV1,
+    /// The version of the Orchard protocol used in Zcash for NU6.2, only instantiated for the
+    /// Orchard value pool.
+    ///
+    /// Uses the post-NU6.2 fixed Orchard circuit. Cross-address transfers are permitted and notes
+    /// use the V2 plaintext format.
```
daira, line 144:
```suggestion
    /// use the lead-byte 0x02 plaintext format.
```

src/lib.rs
```diff
@@ -117,3 +117,42 @@ impl Proof {
...
+    V2,
+    /// The version of the Orchard protocol used in Zcash NU6.3, instantiated for both the Orchard
+    /// and Ironwood value pools.
+    ///
+    /// Uses the post-NU6.3 circuit for both the Orchard and Ironwood value pools.
+    ///
+    /// For transactional bundles affecting the [`ValuePool::Orchard`] value pool,
+    /// `enableCrossAddress = 0` is required by consensus, so cross-address transfers are
+    /// prohibited and Orchard actions are disallowed in coinbase. Notes use V2 plaintexts.
```
daira, line 153:
```suggestion
    /// prohibited and Orchard actions are disallowed in coinbase. Notes use lead-byte 0x02 plaintexts.
```

src/lib.rs
```diff
@@ -117,3 +117,42 @@ impl Proof {
...
+    ///
+    /// For transactional bundles affecting the [`ValuePool::Ironwood`] value pool, cross-address
+    /// transfers are permitted and notes use V3 plaintexts.
```
daira, line 156:
```suggestion
    /// transfers are permitted and notes use lead-byte 0x03 plaintexts.
```

src/note_encryption.rs
```diff
@@ -196,24 +195,21 @@ pub type OrchardDomain = NoteEncryptionDomain<OrchardVersion>;
 /// [`NoteVersion::V3`] note plaintexts, which use lead byte `0x03`.
 pub type IronwoodDomain = NoteEncryptionDomain<IronwoodVersion>;
 
-/// Bundle-pool-restricted note encryption logic.
+/// Note encryption logic restricted to a single note plaintext version.
```
daira, line 198:
```suggestion
/// Note encryption logic restricted to a single note plaintext format.
```

src/note_encryption.rs
```diff
@@ -196,24 +195,21 @@ pub type OrchardDomain = NoteEncryptionDomain<OrchardVersion>;
...
 ///
-/// This domain is used by public bundle helpers that are given the target
-/// [`BundlePoolRestrictions`]. Trial decryption still happens once; after
-/// decryption succeeds, the revealed note plaintext lead byte selects the note
-/// version and the bundle pool restrictions are enforced against it.
+/// This domain is used by public bundle helpers that are given the bundle's
+/// [`NoteVersion`]. Trial decryption still happens once; after decryption
```
daira, line 201:
```suggestion
/// note plaintext format ([`NoteVersion`]). Trial decryption still happens once; after decryption
```

src/note_encryption.rs
```diff
@@ -196,24 +195,21 @@ pub type OrchardDomain = NoteEncryptionDomain<OrchardVersion>;
...
+/// succeeds, the revealed note plaintext lead byte selects the note version, which is
```
daira, line 202:
```suggestion
/// succeeds, the revealed note plaintext lead byte selects the note plaintext format, which is
```

src/bundle.rs
```diff
@@ -51,93 +51,139 @@ impl<T> Action<T> {
...
+            ProtocolVersion::InsecureV1 => OrchardCircuitVersion::InsecurePreNu6_2,
+            ProtocolVersion::V2 => OrchardCircuitVersion::FixedPostNu6_2,
+            ProtocolVersion::V3 => OrchardCircuitVersion::PostNu6_3,
         }
     }
 
-    /// The [`NoteVersion`] associated with this bundle pool restriction.
+    /// The [`NoteVersion`] associated with this bundle version.
     ///
     /// Orchard pools use V2 note plaintexts, and Ironwood pools use V3 note
```
daira, line 143:
```suggestion
    /// Orchard pools use lead-byte 0x02 note plaintexts, and Ironwood pools use lead-byte 0x03 note
```

src/bundle.rs
```diff
@@ -283,95 +335,91 @@ impl Flags {
...
         // https://p.z.cash/TCR:bad-txns-v5-reserved-bits-nonzero
         if value & FLAGS_ALWAYS_EXPECTED_UNSET != 0 {
             return None;
         }
-        let bit2 = value & FLAG_V6_CROSS_ADDRESS_ENABLED != 0;
 
-        // Bit 2 can only be 1 for an Ironwood bundle post-NU6.3 (necessarily a v6+ transaction).
-        // Otherwise it MUST be 0, independent of the tx version:
+        // Bit 2 can only be 1 for an Ironwood bundle. For Orchard it MUST be 0, independent of the
+        // tx version:
```
daira, line 408:
```suggestion
        // tx version and protocol version:
```

src/bundle.rs
```diff
@@ -439,26 +496,46 @@ pub(crate) fn validate_proof_size(proof: &Proof, num_actions: usize) -> Result<(
...
+    /// guarantee (it is debug-asserted here).
     pub(crate) fn from_parts_unchecked(
         actions: NonEmpty<Action<T::SpendAuth>>,
         flags: Flags,
         value_balance: V,
         anchor: Anchor,
         authorization: T,
+        bundle_version: BundleVersion,
     ) -> Self {
+        debug_assert!(flags.to_byte(bundle_version).is_some());
```
daira, line 531: We don't normally use `debug_assert!`. Why not `assert!`, and document the panic condition?

src/bundle.rs
```diff
@@ -491,6 +568,26 @@ impl<T: Authorization, V> Bundle<T, V> {
...
+    /// [`BundleVersion`].
+    ///
+    /// Unlike [`Flags::to_byte`], this is infallible: a `Bundle` is only ever constructed with
+    /// flags that are representable under its version.
+    ///
+    /// [txencoding]: https://zips.z.cash/protocol/protocol.pdf#txnencoding
+    pub fn flag_byte(&self) -> u8 {
+        self.flags
+            .to_byte(self.bundle_version)
+            .expect("flags are validated against the bundle version at construction")
```
daira, line 588: This being an `.expect` reinforces that the check on line 531 should be an `assert!`.

src/bundle.rs
```diff
@@ -965,16 +1071,29 @@ pub mod testing {
...
+            Just(BundleVersion::ironwood_v3()),
         ]
     }
 
+    /// Returns `flags` with its `cross_address_enabled` bit forced to the value representable under
+    /// `bundle_version`, leaving the spend/output bits untouched.
+    ///
+    /// The arbitrary-bundle strategies generate flags independently of the version; this pairs them
+    /// into a combination that a `Bundle` can actually be constructed from.
+    fn flags_for_version(bundle_version: BundleVersion, flags: Flags) -> Flags {
```
daira, line 1089: This arguably should be `arb_flags_for_version` and should return flags with `enableCrossAddress` either set or clear in the `IronwoodV3` case. Otherwise we're getting less test coverage for the (perfectly valid) case where `enableCrossAddress` is clear for an Ironwood-pool bundle.

src/bundle.rs
```diff
@@ -196,7 +248,7 @@ impl Flags {
     /// Construct a set of flags from its constituent parts, including the cross-address bit.
     ///
     /// Crate-internal: the builder supplies `cross_address_enabled` from its prover-side default
-    /// for the pool restrictions (see [`Builder`](crate::builder::Builder)).
+    /// for the bundle version (see [`Builder`](crate::builder::Builder)).
     pub(crate) const fn from_parts(
```
daira, line 252: Why does this need to be crate-internal? You can already create a `Flags` with any combination of the defined flags (via `Flags::from_byte(b, BundleVersion::IronwoodV3)`). This being crate-internal just obscures that fact.
```suggestion
    pub const fn from_parts(
```

src/bundle.rs
```diff
@@ -1200,12 +1320,9 @@ pub(crate) mod tests {
         // A byte with bit 2 clear parses as an unrestricted bundle for Orchard pre-NU6.3,
         // and a restricted bundle for Orchard or Ironwood post-NU6.3.
         for value in 0b000..=0b011 {
-            let pre_nu6_3_flags =
-                Flags::from_byte(value, BundlePoolRestrictions::OrchardNu6_2Only).unwrap();
-            let nu6_3_flags =
-                Flags::from_byte(value, BundlePoolRestrictions::OrchardNu6_3Onward).unwrap();
-            let ironwood_flags =
-                Flags::from_byte(value, BundlePoolRestrictions::IronwoodNu6_3Onward).unwrap();
+            let pre_nu6_3_flags = Flags::from_byte(value, BundleVersion::orchard_v2()).unwrap();
```
daira, line 1323: Nit: rename these as `orchard_v2_flags` etc.
Coverage: test `orchard_insecure_v1` here as well.

src/bundle.rs
```diff
@@ -1368,81 +1467,74 @@ pub(crate) mod tests {
...
+                    flags,
+                    0i64,
+                    anchor,
+                    authorization.clone(),
+                    bundle_version,
+                )
+            };
+
+            // Orchard pre-NU6.3 has cross-address implicitly enabled and Orchard NU6.3 has it
+            // disabled, but both encode to the same wire byte, so their commitments agree.
```
daira, line 1519:
```suggestion
            // disabled, but both encode to the same wire byte, so their commitments agree.
            // (The top-level txid or auth digest will still vary due to the consensus branch ID.)
```

src/bundle.rs
```diff
@@ -1475,31 +1573,37 @@ pub(crate) mod tests {
         #[test]
         fn anchor_placement_follows_tx_version(bundle in arb_bundle(3)) {
             // Orchard post-NU6.3 cannot encode cross-address transfers, so clear the bit to keep
-            // the flags representable in every format under test.
-            let mut flags = *bundle.flags();
-            flags.cross_address_enabled = false;
+            // the flags representable in every version under test.
```
daira, line 1576: This is one place `arb_flags_for_version` would help.

src/bundle.rs
```diff
@@ -1475,31 +1573,37 @@ pub(crate) mod tests {
...
-            let a = with_anchor(crate::Anchor::from_bytes([0u8; 32]).unwrap());
-            let b = with_anchor(crate::Anchor::from_bytes([6u8; 32]).unwrap());
+            let anchor_a = crate::Anchor::from_bytes([0u8; 32]).unwrap();
+            let anchor_b = crate::Anchor::from_bytes([6u8; 32]).unwrap();
 
-            for (pool_restrictions, tx, anchor_in_txid_digest) in [
-                (BundlePoolRestrictions::OrchardNu6_3Onward, TxVersion::V5, true),
-                (BundlePoolRestrictions::OrchardNu6_3Onward, TxVersion::V6, false),
-                (BundlePoolRestrictions::IronwoodNu6_3Onward, TxVersion::V6, false),
+            for (bundle_version, tx, anchor_in_txid_digest) in [
```
daira, line 1596: Consider adding `BundleVersion::supports_anchor_update()`.

PR #520, opened 2026-07-02
Add preverified PCZT signing parse path

czarcas7ic:
## Summary

Adds preverified PCZT signing parse entry points: `Bundle::parse_preverified_for_signing`, `Action::parse_preverified_for_signing`, and `Spend::parse_preverified_for_signing`. They behave identically to the existing `parse` methods except that the spend parse skips `FullViewingKey` derivation and does not validate or preserve the wire `fvk`.

The full `parse` methods are unchanged and remain the default for the Verifier, Prover, Updater, and any caller that needs to validate or retain `fvk`. The preverified signing path is only for a signing pass after the caller has already run the full Verifier checks over the exact same PCZT bytes. `Action::sign` reads only `alpha`, `rk`, and the seed-derived `ask`, never `fvk`, so a preverified-signing parsed action produces a byte-identical `spend_auth_sig`.

## Motivation

The consumer is an embedded hardware signer that signs Orchard PCZTs after a full verify pass, where `FullViewingKey::from_bytes` dominates signing-time parse and is redundant with the review pass that already derived and checked the FVK.

## Testing

- `cargo fmt --all -- --check`
- `cargo test preverified_parse_signs_identically_to_full_parse --locked`
- `cargo test --all-features preverified_parse_signs_identically_to_full_parse --locked`

The test builds a real Orchard action, re-parses its wire bytes through both the full and preverified signing paths, signs both with the same `ask`, `sighash`, and seeded RNG, and asserts the two `spend_auth_sig` values are byte-identical. It also verifies the preverified signing signature against the full action’s `rk`, and checks that a malformed on-wire `fvk` still parses as `None` on the preverified signing path after full parse rejects it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,10 @@ Existing callers keep the current behavior by constructing bundles with
   - `orchard::pczt::Bundle::bundle_version`, the generated getter for the bundle's
     `BundleVersion`, and `orchard::pczt::Bundle::flag_byte`, the infallible byte encoding of
     its flags under that version.
+  - `orchard::pczt::{Bundle, Action, Spend}::parse_preverified_for_signing`,
+    PCZT parse entry points for a preverified signing pass. They skip FVK
+    derivation and do not validate or preserve the wire `fvk`; callers must have
+    already run the full Verifier checks over the same PCZT bytes.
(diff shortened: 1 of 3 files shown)


czarcas7ic:
Superseded by #521, the same change reopened as a same-repo PR instead of from a fork.

PR #521, opened 2026-07-03
Add preverified PCZT signing parse path

czarcas7ic:
## Summary

Adds preverified PCZT signing parse entry points: `Bundle::parse_preverified_for_signing`, `Action::parse_preverified_for_signing`, and `Spend::parse_preverified_for_signing`. They behave identically to the existing `parse` methods except that the spend parse skips `FullViewingKey` derivation and does not validate or preserve the wire `fvk`.

The full `parse` methods are unchanged and remain the default for the Verifier, Prover, Updater, and any caller that needs to validate or retain `fvk`. The preverified signing path is only for a signing pass after the caller has already run the full Verifier checks over the exact same PCZT bytes. `Action::sign` reads only `alpha`, `rk`, and the seed-derived `ask`, never `fvk`, so a preverified-signing parsed action produces a byte-identical `spend_auth_sig`.

## Motivation

The consumer is a Keystone that signs Orchard PCZTs after a full verify pass, where `FullViewingKey::from_bytes` dominates signing-time parse and is redundant with the review pass that already derived and checked the FVK.

## Testing

- `cargo fmt --all -- --check`
- `cargo test preverified_parse_signs_identically_to_full_parse --locked`
- `cargo test --all-features preverified_parse_signs_identically_to_full_parse --locked`

The test builds a real Orchard action, re-parses its wire bytes through both the full and preverified signing paths, signs both with the same `ask`, `sighash`, and seeded RNG, and asserts the two `spend_auth_sig` values are byte-identical. It also verifies the preverified signing signature against the full action’s `rk`, and checks that a malformed on-wire `fvk` still parses as `None` on the preverified signing path after full parse rejects it.

### DIFF
--- a/src/pczt/parse.rs
+++ b/src/pczt/parse.rs
@@ -29,10 +29,6 @@ impl Bundle {
     /// See [`BundleVersion`] for the choice of `bundle_version`.
     ///
     /// `value_sum` is represented as `(magnitude, is_negative)`.
-    ///
-    /// The `actions` must have been parsed via [`Action::parse`] (the full parse). Use
-    /// this for the Verifier, Prover, Updater, or any caller that needs to validate or
-    /// preserve the wire `fvk`.
     pub fn parse(
         actions: Vec<Action>,
         flags: u8,
@@ -85,40 +81,10 @@ impl Bundle {
             bsk,
         })
     }
-
-    /// Parses a PCZT bundle for a preverified signing pass, from `actions` parsed via
-    /// [`Action::parse_preverified_for_signing`].
-    ///
-    /// Bundle-level fields are parsed identically to [`Bundle::parse`]; the only difference
-    /// is that each action's spend omits its [`FullViewingKey`]. See
-    /// [`Spend::parse_preverified_for_signing`] for the invariant and usage constraints.
-    pub fn parse_preverified_for_signing(
-        actions: Vec<Action>,
-        flags: u8,
-        bundle_version: BundleVersion,
-        value_sum: (u64, bool),
-        anchor: [u8; 32],
-        zkproof: Option<Vec<u8>>,
-        bsk: Option<[u8; 32]>,
-    ) -> Result<Self, ParseError> {
-        // Bundle-level parsing does no FVK work; the elision is entirely in the spend parse.
-        Self::parse(
-            actions,
-            flags,
-            bundle_version,
-            value_sum,
-            anchor,
-            zkproof,
-            bsk,
-        )
-    }
 }
 
 impl Action {
     /// Parses a PCZT action from its component parts.
-    ///
-    /// This performs the full parse: the `spend` must have been parsed via
-    /// [`Spend::parse`], which validates and preserves the wire `fvk` when present.
     pub fn parse(
         cv_net: [u8; 32],
         spend: Spend,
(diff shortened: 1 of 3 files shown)


src/pczt/parse.rs
```diff
@@ -81,10 +85,40 @@ impl Bundle {
...
+        Self::parse(
+            actions,
+            flags,
+            bundle_version,
+            value_sum,
+            anchor,
+            zkproof,
+            bsk,
+        )
+    }
```
nuttycom: Given that this is a straight pass-through, why do we need a separate method?

src/pczt/parse.rs
```diff
@@ -110,10 +144,35 @@ impl Action {
...
+    /// method for the invariant and usage constraints.
+    pub fn parse_preverified_for_signing(
+        cv_net: [u8; 32],
+        spend: Spend,
+        output: Output,
+        rcv: Option<[u8; 32]>,
+    ) -> Result<Self, ParseError> {
+        // The FVK skip happens inside the spend parse, not here, so reuse `parse`.
+        Self::parse(cv_net, spend, output, rcv)
+    }
```
nuttycom: Same question here. If this doesn't *enforce* any additional invariants, why do we need a separate method?

nuttycom (approved):
utACK modulo that I don't think the two top-level `parse_preverified` methods are needed.

src/pczt/parse.rs
```diff
@@ -130,6 +189,108 @@ impl Spend {
...
+        rho: Option<[u8; 32]>,
+        rseed: Option<[u8; 32]>,
+        fvk: Option<[u8; 96]>,
+        witness: Option<(u32, [[u8; 32]; NOTE_COMMITMENT_TREE_DEPTH])>,
+        alpha: Option<[u8; 32]>,
+        zip32_derivation: Option<Zip32Derivation>,
+        dummy_sk: Option<[u8; 32]>,
+        note_version: NoteVersion,
+        proprietary: BTreeMap<String, Vec<u8>>,
+        skip_fvk: bool,
```
ValarDragon: Lets just change this to a private enum, so its easier for the caller

src/pczt/parse.rs
ValarDragon: Agreed

czarcas7ic: https://github.com/zcash/orchard/pull/521/commits/de422992ac4a24fcd45cd60916570441a7d3f3cd

src/pczt/parse.rs
czarcas7ic: https://github.com/zcash/orchard/pull/521/commits/de422992ac4a24fcd45cd60916570441a7d3f3cd

src/pczt/parse.rs
czarcas7ic: https://github.com/zcash/orchard/pull/521/commits/ab7c5fae83d3c197ba045b6485fa0206d68dc6f4

ValarDragon (approved):

src/pczt/parse.rs
```diff
@@ -112,8 +112,25 @@ impl Action {
     }
 }
 
+/// Whether [`Spend::parse_inner`] derives the on-wire `fvk` (the full parse) or ignores it,
+/// leaving the parsed `fvk` as `None` (the preverified signing parse). See
+/// [`Spend::parse_preverified_for_signing`] for why skipping is sound.
+enum FvkHandling {
```
nuttycom, line 118: Yay no more boolean blindness!

nuttycom (approved):
utACK ab7c5fae83d3c197ba045b6485fa0206d68dc6f4

PR #523, opened 2026-07-03
Add PCZT recompute helpers and opt-in unpadded bundles

czarcas7ic:
## Summary

Two changes that shrink the PCZT a producer sends an air-gapped signer like Keystone, and for single-action Orchard-pool bundles shrink the transaction itself.

- **Recomputable derived fields:** new `orchard::pczt::recompute` helpers rebuild an action's 6 derived fields (`cv_net`, `nullifier`, `rk`, `cmx`, `ephemeral_key`, `enc_ciphertext`) from the note components the PCZT already carries, so a producer can omit them and the signer regenerates them byte-identically before parsing. `out_ciphertext` is excluded because it is RNG-derived from `ovk` and never recomputable. Adds a non-breaking `ParseError::Recompute` variant.
- **Opt-in unpadded transactional bundles:** `BundleType::Transactional` gains a `pad_to_minimum` flag. `BundleType::DEFAULT` keeps the padded 2-action-minimum behavior; the new `BundleType::DEFAULT_UNPADDED` builds exactly the requested actions, for transactions whose shape is already public (e.g. pool migrations). `Builder::bundle_type()` lets fee calculation read back the configured type.

Every recomputed field is asserted byte-identical to the value the builder itself produced, for Orchard V2 and Ironwood V3 bundles; a one-action Ironwood bundle also proves and verifies end-to-end.

### DIFF
--- a/src/pczt/verify.rs
+++ b/src/pczt/verify.rs
@@ -1,11 +1,167 @@
+use alloc::vec::Vec;
 use core::fmt;
+
+use ff::PrimeField;
+use pasta_curves::pallas;
 
 use crate::{
     keys::{FullViewingKey, SpendValidatingKey},
-    note::{ExtractedNoteCommitment, Rho},
-    value::ValueCommitment,
-    Note,
+    note::{ExtractedNoteCommitment, NoteVersion, Nullifier, RandomSeed, Rho},
+    note_encryption::OrchardNoteEncryption,
+    value::{NoteValue, ValueCommitTrapdoor, ValueCommitment},
+    Address, Note,
 };
+
+/// Byte-level recompute helpers for the derived fields of a PCZT action.
+///
+/// A PCZT producer may omit an action's derived fields (`cv_net`, `nullifier`, `rk`,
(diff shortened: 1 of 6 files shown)


src/pczt/verify.rs
```diff
@@ -1,12 +1,168 @@
...
+///
+/// A PCZT producer may omit an action's derived fields (`cv_net`, `nullifier`, `rk`,
+/// `cmx`, `ephemeral_key`, `enc_ciphertext`) to shrink the serialized encoding, leaving
+/// the receiver to recompute them from the note component fields it already holds. These
+/// functions perform the same derivations as the builder, directly on the PCZT wire byte
+/// encodings, so the receiver can repopulate an omitted field -- with a value
+/// byte-identical to the one the producer elided -- before the protocol structs are
+/// parsed, keeping every downstream consumer working with fully-populated actions.
+///
+/// Errors identify the note component that was missing or could not be decoded.
```
ValarDragon, line 25: I don't get why we need this at all. E.g.a Keystone is already doing each of these separately

ValarDragon: Except maybe the ciphertext - keystone may only decrypt there

PR #525, opened 2026-07-06
Randomize restricted dummy ciphertexts only for external spends

ValarDragon:
## Summary

This PR updates the randomization of output ciphertexts, to only occur for notes decryptable the external viewing key. We added the randomization to protect against a quantum adversary being able to identify a tx-splitting attack for Orchard -> Ironwood migration, but this is only needed for notes paid to one's _external viewing key_. The quantum risk does not apply for the internal case, as the adversary does not learn the user's internal ivk address.

## Validation

- cargo fmt --all -- --check
- cargo test -p orchard cross_address_disabled_spend_output_randomization_depends_on_scope --features circuit
- cargo test -p orchard cross_address_disabled --features circuit


nuttycom (approved):
utACK

ebfull (approved):

PR #527, opened 2026-07-06
Add opt-in unpadded transactional bundles

czarcas7ic:
Adds a `pad_to_minimum` flag to `BundleType::Transactional`. `BundleType::DEFAULT` keeps today's padded behavior; the new `BundleType::UNPADDED` builds exactly the requested actions, for transactions whose shape is already public (e.g. pool migrations). Also adds `Builder::bundle_type()` so fee calculation can read back the configured type. A 1-action Ironwood bundle proves and verifies end-to-end.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -34,53 +34,63 @@ use {
     nonempty::NonEmpty,
 };
 
-const MIN_ACTIONS: usize = 2;
+const DEFAULT_MIN_ACTIONS: u8 = 2;
 
 /// An enumeration of rules for Orchard bundle construction.
 ///
-/// This selects only the construction discipline; the bundle's [`Flags`] are supplied separately
-/// to the builder (see [`Builder::new`] and [`BundleVersion::default_flags`]).
+/// This selects only the construction discipline; the bundle's [`Flags`] are
+/// supplied separately to the builder (see [`Builder::new`] and
+/// [`BundleVersion::default_flags`]).
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum BundleType {
-    /// A transactional bundle is padded, when `pad_to_minimum` is set, to contain at least
-    /// 2 actions, irrespective of whether any genuine actions are required.
+    /// A transactional bundle is padded to contain at least the configured
+    /// `pad_to_minimum` actions, defaulting to 2 actions when
+    /// `pad_to_minimum` is [`None`].
     Transactional {
-        /// A flag that, when set to `true`, indicates that a bundle should be produced even if no
-        /// spends or outputs have been added to the bundle; in such a circumstance, all of the
-        /// actions in the resulting bundle will be dummies.
+        /// A flag that, when set to `true`, indicates that a bundle should be
+        /// produced even if no spends or outputs have been added to the bundle;
+        /// in such a circumstance, all of the actions in the resulting bundle
+        /// will be dummies.
         bundle_required: bool,
-        /// A flag that, when set to `false`, disables padding of the bundle to the 2-action
-        /// minimum: the bundle contains exactly the requested actions (at least one, if
+        /// The minimum number of actions to pad the transaction to.
+        ///
+        /// If this is [`None`], the default 2-action minimum is used. Set this
+        /// to `Some(1)` for transactions whose shape is already public: the
+        /// bundle then contains exactly the requested actions (at least one, if
         /// `bundle_required` is set). One-action bundles are consensus-valid
-        /// ([`BundleType::Coinbase`] bundles are never padded); the minimum exists to improve
-        /// indistinguishability, so disabling it lets the action count reveal the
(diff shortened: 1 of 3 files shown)


src/builder.rs
```diff
@@ -42,13 +42,20 @@ const MIN_ACTIONS: usize = 2;
...
         /// spends or outputs have been added to the bundle; in such a circumstance, all of the
         /// actions in the resulting bundle will be dummies.
         bundle_required: bool,
+        /// A flag that, when set to `false`, disables padding of the bundle to the 2-action
+        /// minimum: the bundle contains exactly the requested actions (at least one, if
+        /// `bundle_required` is set). One-action bundles are consensus-valid
+        /// ([`BundleType::Coinbase`] bundles are never padded); the minimum exists to improve
+        /// indistinguishability, so disabling it lets the action count reveal the
+        /// transaction shape.
+        pad_to_minimum: bool,
```
ValarDragon: From review w/ Kris
Make this an Option<u8>, that will then internally pad to that option's value

Default to None, which implies 2 internally.

ValarDragon (approved):

src/builder.rs
czarcas7ic: https://github.com/zcash/orchard/pull/527/commits/4c5034c21ee4e14a6f32a3a8ed839a0c953a7e12

tests/builder.rs
```diff
@@ -344,6 +344,41 @@ fn post_nu6_3_coinbase_bundle_proves_and_verifies() {
     verify_bundle(&bundle, &post_nu6_3_vk, TxVersion::V6);
 }
 
+// An explicitly unpadded transactional bundle builds exactly the requested single action
+// instead of padding to the 2-action minimum, and the result proves and verifies on the
+// post-NU6.3 circuit like any other bundle (coinbase bundles already demonstrate that
+// consensus accepts 1-action bundles).
+#[test]
+fn unpadded_ironwood_bundle_builds_single_action_and_verifies() {
+    let mut rng = OsRng;
```
dannywillems, line 353: Out-of-scope of this patch, as it seems to be more global.

Nit: for reproducibility, we could use either `proptest` (and therefore covers a larger set of inputs), or we could use something like (pseudo-code):
```
let seed = PRNG(OsRng);
println!("Seed: {seed}");
let mut rng = RNG(seed);
```

src/builder.rs
```diff
@@ -34,53 +34,63 @@ use {
...
 };
 
-const MIN_ACTIONS: usize = 2;
+const DEFAULT_MIN_ACTIONS: u8 = 2;
 
 /// An enumeration of rules for Orchard bundle construction.
 ///
-/// This selects only the construction discipline; the bundle's [`Flags`] are supplied separately
-/// to the builder (see [`Builder::new`] and [`BundleVersion::default_flags`]).
+/// This selects only the construction discipline; the bundle's [`Flags`] are
```
dannywillems, line 41: Nit: this doesn't seem to be any semantic change and related to the commit description `[Allow configurable minimum bundle padding](https://github.com/zcash/orchard/pull/527/changes/4c5034c21ee4e14a6f32a3a8ed839a0c953a7e12)`; this is wrapping at 80 chars (I guess). I support the changes but it would be better to have these cosmetic changes separately.

src/builder.rs
```diff
@@ -42,13 +42,20 @@ const MIN_ACTIONS: usize = 2;
...
+    /// A transactional bundle is padded, when `pad_to_minimum` is set, to contain at least
+    /// 2 actions, irrespective of whether any genuine actions are required.
     Transactional {
         /// A flag that, when set to `true`, indicates that a bundle should be produced even if no
         /// spends or outputs have been added to the bundle; in such a circumstance, all of the
         /// actions in the resulting bundle will be dummies.
         bundle_required: bool,
+        /// A flag that, when set to `false`, disables padding of the bundle to the 2-action
+        /// minimum: the bundle contains exactly the requested actions (at least one, if
+        /// `bundle_required` is set). One-action bundles are consensus-valid
```
dannywillems, line 54: Nit:
> One-action bundles are consensus-valid ([`BundleType::Coinbase`] bundles are never padded); the minimum exists to improve indistinguishability, so disabling it lets the action count reveal the
transaction shape.

Is an explanatory comment on the consensus design. This could be lifted up above in the enum definition to inform the user directly, without requiring to go through the field definitions, to understand the reasoning.

dannywillems (approved):
utACK c75c6e61e10c944e62c75c412fc20e7f314af37d.

The first commit and the second commits should be only one, and the doc wrapping could have been two separate commits.
The first commit uses a boolean where the second uses a count. I guess it changed over time for the API user, as more than two values (1 and 2) are would be accepted, by libraries.

dannywillems:
Merging this to update ref in librustzcash patch https://github.com/zcash/librustzcash/pull/2551/

PR #531, opened 2026-07-12
Add Post-NU6.3 verifier fingerprint fixtures

### LINKED ISSUE #52: Check how/where Zebra pins the PinnedVerificationKey
(and, for curiosity, how zcashd did pin it).

This is so that we can check that any fixture tests we have in `ironwood` (e.g. those added in #27, which use Lean-syntax fixtures captured by TalDerei/halo2#1) are checking the same things as `PinnedVerificationKey` checks in the deployed Rust code.

TalDerei:
**Human Summary:**

Orchard side PR for https://github.com/zcash/halo2/pull/914, and consumed in https://github.com/zcash/ironwood/pull/57. See the generated fixture in https://github.com/zcash/ironwood/blob/main/Zcash/Snark/Fixtures/MultiAction/Fixture.lean.

This adds Orchard capture tests for Post-NU6.3 verifier fixtures. The tests build real bundles, compare the generated verifying key with Orchard's `PinnedVerificationKey` (closes zcash/ironwood#52) , and then emits deterministic fixtures for Ironwood.

Concretely,
  - Generates real single/two-action Post-NU6.3 proofs,
  - Confirms their verifying key matches Orchard’s canonical pinned key,
  - Captures the verifier MSM and Fiat–Shamir transcript, then exports concrete Vesta Lean fixtures for Ironwood,
  - Tests tampered and malformed proofs,

Open to organizational / feature gating preferences.

----------

**AI Summary:**

This PR adds an opt-in Orchard `verifier-fingerprint` fixture module that builds deterministic real Post-NU6.3 bundles, pins their verifying key to Orchard's canonical checked-in key, and exports normal verifier runs as concrete Lean fixtures. It does not change proof acceptance.

## What Is Added

- `src/circuit/fingerprint/` contains the single- and two-action captures plus separate rejection cases, and is compiled only with Orchard's `verifier-fingerprint` feature.
- Each capture checks the exact `PinnedVerificationKey`, verifies normally, and exports the resulting transcript and assembled MSM.
- Fixed RNG seeds and separate opt-in output paths make regeneration byte-for-byte deterministic and race-free.
- Rejection tests cover a tampered advice evaluation, truncated multiopen data, and a missing permutation evaluation.
- The checkout patch pins the helper PR head from the canonical `zcash/halo2` repository so Orchard and `halo2_gadgets` use one `halo2_proofs` package; the capture API remains feature-gated.

## Remaining Floor

These fixtures compare Ironwood against concrete Rust verifier runs; they are not a verifier soundness proof. Regeneration still trusts Orchard's canonical-key comparison and Halo2's pinned-key serialization and Blake2b transcript derivation, and the exact cross-repository comparison is not yet automated in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,7 +116,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Generate coverage report
-        run: cargo tarpaulin --engine llvm --all-features --release --timeout 600 --out xml
+        # Enable every feature except `verifier-fingerprint`: it is opt-in capture tooling whose
+        # tests build the k=11 proving key and generate real proofs, which are far too heavy to run
+        # under coverage instrumentation on every PR. Run those manually per the docs in
+        # `src/circuit/fingerprint/mod.rs`.
+        run: cargo tarpaulin --engine llvm --features "dev-graph unstable-frost unstable-voting-circuits test-dependencies" --release --timeout 600 --out xml
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3.1.4
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,13 @@ jobs:
       - uses: actions/checkout@v4
       - name: Run tests
         run: cargo test --verbose
+      # verifier-fingerprint capture tests build the k=11 proving key and run real proofs, too
+      # heavy for every PR (regenerate the fixtures manually per the docs in
+      # src/circuit/fingerprint/mod.rs). Compile them so the cfg(test) module can't rot unnoticed:
+      # `cargo test` (default features) and `cargo build --all-features` (skips cfg(test)) never
+      # type-check it.
+      - name: Compile-check the verifier-fingerprint capture module
+        run: cargo test --features verifier-fingerprint --no-run --verbose
       - name: Verify working directory is clean
         run: git diff --exit-code
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,7 @@ criterion = "0.4" # 0.5 depends on clap 4 which has MSRV 1.70
 halo2_gadgets = { version = "0.5", features = ["test-dependencies"] }
 hex = "0.4"
 proptest = ">=1.0.0, <1.7.0"
+rand_chacha = "0.3"
 zcash_note_encryption = { version = "0.4", features = ["pre-zip-212"] }
 incrementalmerkletree = { version = "0.8.1", features = ["test-dependencies"] }
 shardtree = "0.6"
@@ -89,6 +90,7 @@ unstable-frost = []
 unstable-voting-circuits = []
 multicore = ["halo2_proofs?/multicore"]
 dev-graph = ["halo2_proofs?/dev-graph", "image", "plotters"]
+verifier-fingerprint = ["circuit", "halo2_proofs/verifier-fingerprint"]
 test-dependencies = ["proptest", "rand/std"]
 
 [[bench]]
@@ -110,3 +112,9 @@ debug = true
 
 [profile.bench]
 debug = true
+
+[patch.crates-io]
+# All builds in this checkout resolve Halo2 to the helper fork. Only Orchard's opt-in
+# `verifier-fingerprint` feature enables its capture tooling; downstream users do not inherit a
+# package's root-level patch.
+halo2_proofs = { git = "https://github.com/zcash/halo2", rev = "cd63cdf3826d2b274810b6c314950cccaa4139d3" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ keywords = ["zcash"]
 
 [package.metadata.docs.rs]
 # We intentionally do not use `all-features = true`: that would enable `verifier-fingerprint`, which
-# requires the `halo2_proofs/verifier-fingerprint` feature that lives only in the git fork pinned
+# requires the `halo2_proofs/unstable-verifier-fingerprint` feature that lives only in the git fork pinned
 # via the patch section below. On docs.rs the crate is built as published, where the patch is
 # stripped and `halo2_proofs` resolves from crates.io without that feature, so the docs build would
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -117,4 +124,13 @@ debug = true
...
+#
+# PRE-PUBLISH REQUIREMENT: this patch (and the git `rev` below) must be removed before any
+# `cargo publish` of Orchard, and `verifier-fingerprint` must resolve without it. The sequence is:
+#   1. merge https://github.com/zcash/halo2/pull/914,
+#   2. release `halo2_proofs` 0.3.3 with the `verifier-fingerprint` feature,
+#   3. drop this `[patch.crates-io]` section (the `version = "0.3"` dep then resolves 0.3.3),
+#   4. restore `all-features = true` under `[package.metadata.docs.rs]`.
+# The `rev` currently pins the *unmerged* head of PR #914; repin to the merged commit once it lands
+# so a rebase of that branch cannot orphan it.
 halo2_proofs = { git = "https://github.com/zcash/halo2", rev = "cd63cdf3826d2b274810b6c314950cccaa4139d3" }
```
TalDerei: this is a temporary measure for compilation purposes; whether it's the right convention for this repo still needs discussion.

.github/workflows/ci.yml
```diff
@@ -116,7 +116,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Generate coverage report
-        run: cargo tarpaulin --engine llvm --all-features --release --timeout 600 --out xml
+        # Enable every feature except `verifier-fingerprint`: it is opt-in capture tooling whose
+        # tests build the k=11 proving key and generate real proofs, which are far too heavy to run
+        # under coverage instrumentation on every PR. Run those manually per the docs in
+        # `src/circuit/fingerprint/mod.rs`.
+        run: cargo tarpaulin --engine llvm --features "dev-graph unstable-frost unstable-voting-circuits test-dependencies" --release --timeout 600 --out xml
```
ebfull, line 130: Excluding these from coverage makes sense, but nothing in CI compiles the module at all now: `cargo test` is default-features and `cargo build --all-features` skips `cfg(test)` code. A compile-only guard would catch rot cheaply, e.g. `cargo test --features verifier-fingerprint --no-run` in the test job (it type-checks in seconds).

src/circuit/fingerprint/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+//! challenges as a Lean fixture. This checks that Ironwood's Lean model reproduces one *concrete*
+//! Rust verifier run; it is not a verifier soundness proof.
+//!
+//! Two limitations are load-bearing for anyone consuming the exported fixtures:
+//!
+//! * **Only accepting runs are exported to Lean.** The negative captures in `rejected` confirm in
+//!   Rust that a tampered/short/desynchronized proof assembles a non-identity MSM and is rejected by
+//!   the deployed verifier, but they are never handed to Lean. The Lean `MsmMatch` theorem is only
+//!   ever checked against *identity* (accepting) runs, so a Lean `assemble` that trivially accepts
+//!   would still pass every exported fixture. Extending the cross-check to a rejecting run needs a
```
ebfull: A trivially-accepting `assemble` would fail `MsmMatch` here (the match pins every coefficient, not just acceptance); the unchecked part is the rejection paths:

```suggestion
//!   the deployed verifier, but they are never handed to Lean. `MsmMatch` pins the full coefficient
//!   structure of accepting runs, so the unchecked residual is narrower: a Lean `assemble` that
//!   agrees on accepting runs but mismodels the rejection paths would still pass every exported
//!   fixture. Extending the cross-check to a rejecting run needs a
```

Also, is the dumper variant filed anywhere?

Cargo.toml
```diff
@@ -19,7 +19,14 @@ categories = ["cryptography::cryptocurrencies"]
 keywords = ["zcash"]
 
 [package.metadata.docs.rs]
-all-features = true
+# We intentionally do not use `all-features = true`: that would enable `verifier-fingerprint`, which
+# requires the `halo2_proofs/verifier-fingerprint` feature that lives only in the git fork pinned
```
ebfull: nit: the halo2 feature is `unstable-verifier-fingerprint` (also in step 2 of the pre-publish list below):

```suggestion
# requires the `halo2_proofs/unstable-verifier-fingerprint` feature that lives only in the git fork pinned
```

src/circuit/fingerprint/rejected.rs
```diff
@@ -0,0 +1,230 @@
...
+    let n_permutation_set_evals = instances.len() * (3 * n_permutation_sets - 1);
+    let n_lookup_evals = instances.len() * n_lookups * 5;
+    let first_multiopen_u = n_instance_evals
+        + instances.len() * n_advice_queries
+        + n_fixed_evals
+        + 1
+        + n_permutation_common_evals
+        + n_permutation_set_evals
+        + n_lookup_evals;
+    let n_multiopen_u = transcript.scalars.len() - first_multiopen_u - 2;
```
ebfull: optional: `first_multiopen_u` and `n_multiopen_u` are derivable from `transcript.events`, since the u evals are exactly the ReadScalars between the $x_3$ and $x_4$ squeezes; that would drop most of the constants above.

ebfull (dismissed):
I think we can get halo2's PR merged before this one, so we can get rid of the patch before merging this, but it can also merge as-is.

TalDerei:
> I think we can get halo2's PR merged before this one, so we can get rid of the patch before merging this, but it can also merge as-is.

@ebfull opting to not merge with a patch (which currently points to an outdated commit anyways), so this will block on https://github.com/zcash/halo2/pull/914 and cutting release. I'll merge after that CI pass.

src/circuit/fingerprint/mod.rs
TalDerei: no it isn't. https://github.com/zcash/halo2/pull/914#issuecomment-5025729204 was a design i rejected, and instead we just fail fast on the rust side.

.github/workflows/ci.yml
TalDerei: it does indeed type check very quickly, added a compile guard.

ebfull (approved):

PR #532, opened 2026-07-12
Update to Rust 1.88 / edition 2024, rebuild CI, tighten lints

dannywillems:
Follows librustzcash, and mirrors the same change in [pasta_curves#90](https://github.com/zcash/pasta_curves/pull/90). MSRV goes 1.85.1 -> 1.88 and the crate moves to edition 2024.

## Edition 2024

Most of the diff is the 2024 style edition. The migration itself needed three real changes, all worth a reviewer's eye:

- `Builder::{spends,outputs,changes}` return `impl Trait`, which in edition 2024 captures every lifetime in scope by default. `+ use<>` keeps the 2021 capture behaviour, so the **public signatures do not change meaning**.
- `sample_authorized_bundle` binds the proptest value tree instead of leaving it a temporary of the tail expression, so its drop order relative to `runner` no longer depends on the edition. Neither type has a side-effecting `Drop`, so this is a readability fix, not a behaviour fix.
- The `if let` in `Bundle::extract` **keeps its shape**. `cargo fix` proposed rewriting it to a `match` because of the 2024 `if let` temporary-scope change, but the `else` branch never touches the scrutinee temporary, so the earlier drop is harmless and the `match` is strictly less readable.

The rest is warnings the newer compiler and clippy surface: inlined format args, closures replaced by the method they call, `u64 as i128` via `From`, iteration over references.

## Lints

Lints were crate attributes in `src/lib.rs`, which only covers the library target: tests and benches were linted with the defaults. They move to `[lints]`. `missing_docs`, `missing_debug_implementations` and the `dead_code` allow stay in `lib.rs` (the first two concern the public API, and denying them package-wide would also hit the benches). The `allow(dead_code)` is still genuinely earned, so it stays.

Two lints are **deliberately excluded**, with the reasons recorded next to the table:
- `needless_pass_by_value` (77 hits) would require changing public signatures to take references.
- `unused_qualifications` is per-feature-config, and its fix under `--all-features` **breaks the other configs**: `NoteCommitTrapdoor` is only re-exported behind `unstable-voting-circuits`, so with all features on it looks like a droppable qualification, but the crate then fails to compile without that feature. I hit this for real; see below.

## CI

- The Linux tests get a features matrix (default / all / none). Previously **only the default feature set was ever tested**.
- New `--no-default-features` clippy job. This is not redundant with `--all-features`: it is what caught the `NoteCommitTrapdoor` breakage above, which `--all-features` clippy happily suggested and could not see.
- Required jobs gated behind a single `required-checks` job to mark as required in branch protection; macOS/Windows, beta clippy and coverage stay advisory.
- Every action pinned by commit SHA, `permissions: {}` at the workflow level and minimum per job, `persist-credentials: false` on checkout, plus a zizmor workflow and a dependabot config (with cooldown) to keep it that way.
- mdBook installed via a pinned action rather than piping a `curl` release tarball into `tar`.

## Verified locally

On 1.88: `fmt --check`, `clippy -D warnings` across 8 feature combinations, the full test suite (153 tests), benches, `mdbook test`, and no-std builds for `wasm32-wasip1` and `thumbv7em-none-eabihf`.


dannywillems:
Closing as, after discussing with @nuttycom, we won't update to 1.88 until it is very necessary.

PR #534, opened 2026-07-21
builder: support deferring the bundle anchor to proving time (ZIP 374)

nuttycom:
The builder demanded an anchor at construction and a Merkle witness rooting to it for every non-dummy spend at add-spend time, even though for the V6 bundle formats neither the txid digest nor any signature commits to the anchor (ZIP 374 moved it to the authorizing-data digest), and witnesses are pure proving inputs. Flows that pre-sign transactions long before proving them — notably the ZIP 318 pool-migration transfers, whose proving anchor is drawn near each transfer's future broadcast height — therefore had to fabricate synthetic witnesses at build time and strip them before use.

Builder::new_with_anchor_deferred constructs a builder with no anchor at all (internally BuilderAnchor::Deferred, alongside the existing Fixed mode); add_spend_unwitnessed adds a spend as a bare (fvk, note) pair, still validating note ownership. It is refused for any bundle format whose txid digest commits the anchor (only the V6 formats defer), the witnessed and unwitnessed entry points reject each other's mode, and an in-memory build — which proves immediately — is refused outright; only build_for_pczt can build such a bundle. The witness-to-anchor consistency check that an anchored builder performs per spend is deferred to the PCZT Prover role, which performs it once the real anchor and witnesses are installed through the PCZT Updater role after signing.

The built pczt::Bundle reports the deferral through a new anchor_deferred accessor, and its real spends' witness fields are None (SpendInfo's Merkle path became Option internally; padding dummies keep their witnesses, which the prover requires and the consistency check exempts for zero values). The bundle's required — and publicly read — anchor field carries the empty-tree root purely as a placeholder, which a PCZT serializer must emit as ABSENT; parsed bundles always report a non-deferred anchor, wire-anchor absence being tracked separately by the serializer for re-emission.

All additions are API-additive: no existing signature changes, and the new error variants land on non-exhaustive enums.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- `orchard::builder::Builder::new_with_anchor_deferred`, constructing a builder whose bundle
+  anchor — and every real spend's Merkle witness — is deferred to proving time, per
+  [ZIP 374](https://zips.z.cash/zip-0374). Spends are added without witnesses via the new
+  `orchard::builder::Builder::add_spend_unwitnessed`; the built PCZT bundle reports the
+  deferral through the new `orchard::pczt::Bundle::anchor_deferred` accessor (its `anchor`
+  field then carries the empty-tree root purely as a placeholder that a PCZT serializer
+  must emit as absent, and the real spends' `witness` fields are `None`), and the real
+  anchor and witnesses are installed through the PCZT Updater role after signing. Only
+  supported for bundle formats whose txid digest — and hence every signature — excludes
+  the anchor (the V6 formats); only `Builder::build_for_pczt` can build such a bundle.
+- `orchard::builder::BuildError::{AnchorRequired, AnchorDeferralUnsupported}`
+- `orchard::builder::SpendError::{AnchorDeferred, WitnessRequired}`
+
 ## [0.15.0] 2026-07-09
 
 This release introduces `orchard::bundle::BundleVersion`, the `(value pool, protocol
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -12,7 +12,7 @@ use zcash_note_encryption::ENC_CIPHERTEXT_SIZE;
 
 use crate::{
     address::Address,
-    bundle::{Authorization, Authorized, Bundle, BundleVersion, Flags},
+    bundle::{Authorization, Authorized, Bundle, BundleVersion, Flags, TxVersion},
     keys::{
         FullViewingKey, OutgoingViewingKey, Scope, SpendAuthorizingKey, SpendValidatingKey,
         SpendingKey,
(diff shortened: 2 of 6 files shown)


src/pczt.rs
```diff
@@ -83,6 +83,17 @@ pub struct Bundle {
...
+    /// Whether this bundle was built with its anchor DEFERRED to proving time
+    /// ([ZIP 374]): [`anchor`](Self::anchor) then carries the empty-tree root purely as
+    /// a placeholder that a PCZT serializer must emit as ABSENT, and the real spends'
+    /// `witness` fields are `None`; the real anchor and witnesses are installed through
+    /// the PCZT Updater role before proving. Always `false` for a parsed bundle (a
+    /// serializer re-emitting a parsed bundle tracks the wire anchor's absence
+    /// separately).
+    ///
+    /// [ZIP 374]: https://zips.z.cash/zip-0374
+    pub(crate) anchor_deferred: bool,
```
nuttycom, line 100: We should be able to make the `anchor` field have type `Option<Anchor>` instead of carrying a separate boolean also.

src/pczt.rs
```diff
@@ -83,6 +83,22 @@ pub struct Bundle {
...
+    /// serializer re-emitting a parsed bundle tracks the wire anchor's absence
+    /// separately).
+    ///
+    /// [ZIP 374]: https://zips.z.cash/zip-0374
+    // TODO: Fold this flag into the `anchor` field as an `Option<Anchor>` (`None` =
+    // deferred, eliminating the placeholder) the next time an API-breaking change to
+    // this crate is needed for some other reason: it changes the derived public
+    // `anchor()` getter's return type, so it cannot ship in a semver-compatible
+    // release on its own.
+    pub(crate) anchor_deferred: bool,
```
TalDerei, line 100: native `pczt::Bundle::create_proof` afaict doesn't check this flag and passes `self.anchor` directly into every proof instance. so `create_proof` and `extract` paths still don't reject `anchor_deferred`. should we also check this flag in `pczt/prover.rs`?

`Option<Anchor>` per https://github.com/zcash/orchard/pull/534/changes#r3619520733 would alternatively sufficiently resolve this. 

```ts
let anchor = self.anchor.ok_or(ProverError::AnchorRequired)?;
```

CHANGELOG.md
```diff
@@ -5,6 +5,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+  [ZIP 374](https://zips.z.cash/zip-0374). Spends are added without witnesses via the new
+  `orchard::builder::Builder::add_spend_unwitnessed`; the built PCZT bundle reports the
+  deferral through the new `orchard::pczt::Bundle::anchor_deferred` accessor (its `anchor`
+  field then carries the empty-tree root purely as a placeholder that a PCZT serializer
+  must emit as absent, and the real spends' `witness` fields are `None`), and the real
+  anchor and witnesses are installed through the PCZT Updater role after signing. Only
+  supported for bundle formats whose txid digest — and hence every signature — excludes
+  the anchor (the V6 formats); only `Builder::build_for_pczt` can build such a bundle.
+- `orchard::builder::BuildError::{AnchorRequired, AnchorDeferralUnsupported}`
+- `orchard::builder::SpendError::{AnchorDeferred, WitnessRequired}`
```
TalDerei, line 22: "anchor and witnesses are installed through the PCZT Updater role after signing" where are the setters for this?

TalDerei (commented):
left some observations, think it's close.

src/pczt.rs
TalDerei: agreed, this feels like the right approach instead of laundering.

src/pczt/parse.rs
```diff
@@ -77,6 +77,10 @@ impl Bundle {
             bundle_version,
             value_sum,
             anchor,
+            // A parsed bundle always reports a non-deferred anchor: absence on the wire
```
dannywillems, line 80: Nit: We shouldn't leave comments above a parameter. If a parameter needs some comments, we can make a let-binding, and add the comment above the let-binding.

src/builder.rs
```diff
@@ -1899,6 +2053,138 @@ mod tests {
         (note, merkle_path, anchor)
     }
 
+    #[test]
+    fn deferred_anchor_builds_for_pczt_without_witnesses() {
```
dannywillems, line 2057: Nit: the built values could be coming from an `arb_` to be reused by other crates.

src/pczt.rs
dannywillems: Closing the thread as a TODO has been added since then in the code.

dannywillems (approved):
utACK, and open https://github.com/zcash/orchard/pull/535 to address comments and follow-up.

src/pczt.rs
nuttycom: Yeah, I want this to be Option<Anchor> but I didn't want to make an API-breaking change here.

PR #535, opened 2026-07-21
builder: reusable arb_spendable_note fixture + PCZT deferred-anchor Updater setters

dannywillems:
Builds on #534 (`feat/deferred-anchor-builder`), addressing three review comments on it.

## `arb_spendable_note` fixture (r3621743725)

The deferred-anchor test hand-built its `(note, merkle_path, anchor)` values. Extract that into a reusable proptest strategy `arb_spendable_note(value, note_version) -> (SpendingKey, Note, MerklePath, Anchor)` in `builder::testing` (behind `test-dependencies`), so other crates (e.g. the ZIP 318 pool migration) reuse it instead of re-deriving the fixture. `deferred_anchor_builds_for_pczt_without_witnesses` is rewritten as a `proptest!` that consumes it (plus a small `arb_deferred_anchor_builder`) and only asserts. (`note_with_path` is kept for the 6 other tests that need caller-chosen scope/version that this fixture does not model.)

## PCZT Updater setters (r3619909991)

#534's CHANGELOG says the real anchor and witnesses are "installed through the PCZT Updater role after signing", but no setter existed — the `ActionUpdater` had only zip32/proprietary/user-address setters, and `Spend.witness` / `Bundle.{anchor, anchor_deferred}` were `pub(crate)`, while the Prover already requires them. Add:
- `orchard::pczt::Updater::set_anchor(&mut self, Anchor)` — installs the real anchor and clears `anchor_deferred`.
- `orchard::pczt::ActionUpdater::set_spend_witness(&mut self, MerklePath)` — installs the deferred spend's witness.

So a deferred-anchor bundle can now be completed (built -> signed -> Updater installs anchor + witnesses -> proved).

## Prover guard (r3619690791)

The Prover read `self.anchor` directly, so a still-deferred bundle would silently prove against the empty-tree placeholder. Guard it: `create_proof` now returns the new `ProverError::AnchorDeferred` when `anchor_deferred` is still set. (The `tx_extractor` path is left for the `Option<Anchor>` refactor the code TODO proposes, which would cover both uniformly at the next breaking release; the extractor only runs after a successful proof, now blocked for deferred bundles.)

## Also

An `AGENTS.md` for the crate, aligned with librustzcash's conventions (a proptest strategy lives with the type it generates; canonical serialization on the type; semantic types).

## Tests

`cargo test deferred_anchor` -> 4 passed (incl. `deferred_anchor_updater_installs_anchor_and_witness` and `deferred_anchor_prover_rejects_uninstalled_anchor`); `cargo test --lib pczt` -> 21 passed; `--features test-dependencies` build, fmt, and clippy clean.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,60 @@
+# Orchard - Agent Guidelines
+
+Conventions for working in this crate. They mirror the shared conventions in the
+`librustzcash` repository's `AGENTS.md`; when a convention changes in one repo,
+keep the other aligned. Orchard is `#![no_std]` (the `std` feature is optional)
+and is a payment-protocol crate, so some librustzcash conventions about database
+/ storage layout do not apply, but the ones below do.
+
+## Code Conventions
+
+- **Never use magic numbers.** Do not inline a bare numeric (or string) literal
+  whose meaning is not obvious from context. Give it a `const` with a
+  doc-commented rationale, and reuse the protocol's own named constants rather
+  than re-deriving their values. This applies to production code, tests, and
+  fixtures alike.
+
+- **Public APIs use semantic types, never bare primitives.** A `pub` function,
+  trait method, field accessor, or constructor must not represent a domain
+  quantity as a bare integer, byte array, or scalar: note values are
+  `NoteValue`, value sums are `ValueSum`, tree roots are `Anchor`, commitments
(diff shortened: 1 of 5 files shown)


nuttycom (approved):
utACK, looks great!

PR #538, opened 2026-07-21
Derive instance commitments from public inputs in the fingerprint fixure

TalDerei:
Orchard side PR for https://github.com/zcash/halo2/pull/918, and consumed in https://github.com/zcash/ironwood/pull/85.

halo2's `dump_vesta_lean_fixture` now takes the verifier's public `instances` (and no longer a separate `num_proofs`), deriving each Lean instance commitment from the public inputs rather than exporting it as an opaque vk field.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -992,9 +992,8 @@ dependencies = [
 
 [[package]]
 name = "halo2_proofs"
-version = "0.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


TalDerei:
@ebfull patch will be removed after halo2 is tagged.

ebfull (approved):

TalDerei:
@ebfull auto-merge seems to have failed here.

ebfull (approved):

PR #539, opened 2026-07-22
orchard 0.15.4 release: Accelerate batched trial decryption with GLV endomorphism windows

nuttycom:
Supersedes #530: rebased onto `main` (0.15.3) and reworked to consume the GLV primitive from zcash/pasta_curves#93 (`pasta_curves::glv`) instead of a self-contained inline `endo` module. Same feature and the same byte-exact results — cleaner layering, with GLV living in `pasta_curves` and orchard as a thin consumer.

## What changes

The GLV primitive — batched odd-multiples windows built with a single shared inversion, per-scalar decomposition + wNAF recoding, and the shared-doubling Straus ladder — now lives in `pasta_curves::glv` (`Table` / `Decomposed` / `Table::mul_decomposed`). orchard consumes it in the batched pipeline:

- `PreparedEphemeralPublicKey::batch_tabled` builds every ephemeral key's window via `Table::batch` (one shared normalization across the whole batch).
- `batch_ka_agree_dec` decomposes each viewing key once via `Decomposed::new` and reuses it against every ephemeral key's window, through the `BatchDomain::batch_ka_agree_dec` hook added in `zcash_note_encryption` 0.4.2.

The per-output entry points (`try_note_decryption` etc.) are untouched. There are no public API changes; `src/endo.rs` is removed (its GLV moved to `pasta_curves`).

## Correctness

Shared secrets are byte-identical to the per-item path. `note_encryption`'s `batched_agreement_matches_per_item` and `batched_compact_decryption_matches_per_item_{orchard,ironwood}` pin this for both `OrchardDomain` and `IronwoodDomain` over hit / miss / undecodable lanes; the full existing suite passes. The GLV KATs and proptests now live in `pasta_curves::glv`.

## Stack

- **pasta_curves 0.5.2** (the `glv` module) 
- **zcash_note_encryption 0.4.2** — `batch_ka_agree_dec`

Original work by @LukasKorba (#530).

---
Filed with [Claude Code](https://claude.com/claude-code) assistance.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "orchard"
-version = "0.15.3"
+version = "0.15.4"
 authors = [
     "Sean Bowe <ewillbefull@gmail.com>",
     "Jack Grigg <thestr4d@gmail.com>",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ group = "0.13"
 hex = { version = "0.4", default-features = false, features = ["alloc"] }
 lazy_static = "1"
 memuse = { version = "0.2.2", default-features = false }
-pasta_curves = "0.5"
+pasta_curves = { version = "0.5.2", features = ["glv"] }
 proptest = { version = ">=1.0.0, <1.7.0", optional = true }
 rand = { version = "0.8", default-features = false }
 rand_core = { version = "0.6", default-features = false }
@@ -42,7 +42,7 @@ poseidon = { package = "halo2_poseidon", version = "0.1" }
 serde = { version = "1.0", default-features = false, features = ["derive"] }
 sinsemilla = "0.1"
 subtle = { version = "2.6", default-features = false }
-zcash_note_encryption = "0.4"
+zcash_note_encryption = "0.4.2"
 incrementalmerkletree = "0.8.1"
 zcash_spec = "0.2.1"
 zip32 = { version = "0.2.0", default-features = false }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -112,3 +112,9 @@ debug = true
 
 [profile.bench]
 debug = true
+
+# TEMPORARY: this PR is stacked on pasta_curves#91 (the `glv` module it consumes).
+# The patch lets it build until #91 lands in a pasta_curves release, at which point
+# the patch is removed and the `pasta_curves = "0.5.x"` requirement stands alone.
+[patch.crates-io]
+pasta_curves = { git = "https://github.com/LukasKorba/pasta_curves", rev = "357e1837c66e481c86676a7a9b6f9facd2fc2195" }
```
nuttycom: This PR will be kept in draft until https://github.com/zcash/pasta_curves/pull/93 is merged, so that we can depend on either a released `pasta_curves` crate or at least a rev that is on the upstream `main`.

src/keys.rs
```diff
@@ -852,17 +860,81 @@ impl EphemeralPublicKey {
     }
 }
 
+/// What a prepared ephemeral key carries.
+///
+/// Individually-prepared keys carry a `group::Wnaf` window table, consumed by
+/// the per-item multiplication. Batch-prepared keys carry a GLV odd-multiples
+/// window instead (see `pasta_curves::glv`), which is cheaper to build across a
```
ebfull, line 867: Could crosslink to the module here (and in some other places) instead of just mentioning it, especially to avoid bitrot. Alternatively, don't even discuss GLV in user-facing docs.

src/keys.rs
```diff
@@ -852,17 +860,81 @@ impl EphemeralPublicKey {
...
+/// batch (one shared normalization) and cheaper to multiply against (a
+/// shared-doubling ladder over the endomorphism split). Both produce
+/// identical shared secrets.
+#[derive(Clone, Debug)]
+enum PreparedEpkInner {
+    /// A `group::Wnaf` window table.
```
ebfull, line 873: Perhaps another opportunity to crosslink, which helps prevent bitrot.

ebfull (dismissed):
Details of the underlying mechanism and `pasta_curves` APIs don't really need to be exposed in the documentation, but I don't see anything wrong here!

ebfull:
~This PR title ideally wouldn't contain the version number of Orchard that we might speculate it would be merged into.~ Nevermind, we're releasing from this branch.

PR #540, opened 2026-07-22
Release orchard v0.15.3

ebfull:
Patch release: the `verifier-fingerprint` fixture export now derives each Lean instance commitment from the verifier's public inputs (via `halo2_proofs 0.3.4`).


nuttycom (approved):
ACK

PR #541, opened 2026-07-31
Add random match-only fingerprint captures (fabricate→replay)

ebfull:
The honest captures can only pin verifier behavior at honest proofs; a Lean/Rust divergence whose discrepancy vanishes on honest runs (e.g. sourcing a transcript-claimed evaluation from a recomputation instead of the proof string) is invisible to them. These drivers run the deployed verifier on a random proof string instead, exported match-only via halo2's dump_vesta_lean_fixture_match_only.

The proof string is produced in two passes so the read schedule is the deployed verifier's own: a RandomizingTranscript wrapping a real Blake2bWrite samples a fresh random canonical value at each read_point/read_scalar (points as generator·t, t ≠ 0), absorbs and buffers it, and forwards squeezes/common inputs to the inner Blake2b state; the finalized bytes are then replayed through ChallengeRecorder over Blake2bRead — identical to the honest pipeline — and the fixture is exported from the replay. The driver asserts (failing loudly, never re-seeding silently): both verifier passes ran to completion, replay challenges and transcript events equal fabrication's, the captured MSM is non-identity on both passes, and the pinned verifying-key description matches. Public inputs are random raw F_p vectors with the honest captures' shape; no proving key or prover run is involved.

Drivers: fingerprint_capture_random (seed 0x52 'R', 1 action), fingerprint_capture_random_two_actions (0x72 'r', 2), fingerprint_capture_random_three_actions (0x33 '3', 3), with ORCHARD_LEAN_{SINGLE,MULTI,TRIPLE}_RANDOM_{FIXTURE,PROOF}_OUT output env vars. Requires halo2_proofs with the match-only exporter patch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ visibility = "0.1.1"
 
 # Circuit
 halo2_gadgets = { version = "0.5", optional = true, default-features = false }
-halo2_proofs = { version = "0.3", optional = true, default-features = false, features = ["batch", "floor-planner-v1-legacy-pdqsort"] }
+halo2_proofs = { version = "0.3.5", optional = true, default-features = false, features = ["batch", "floor-planner-v1-legacy-pdqsort"] }
 
 # Boilerplate
 getset = "0.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "orchard"
-version = "0.15.4"
+version = "0.15.5"
 authors = [
     "Sean Bowe <ewillbefull@gmail.com>",
     "Jack Grigg <thestr4d@gmail.com>",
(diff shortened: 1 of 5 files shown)


src/circuit/fingerprint/random.rs
TalDerei, line 1: fabricate + replay is a neat design for creating arbitrary invalid proofs that still exercise the complete verifier.

TalDerei (dismissed):
ACK after pinning against halo2 release.

TalDerei:
CI failure observation in https://github.com/zcash/halo2/pull/924#issuecomment-5160222846 applies here too.

Cargo.toml
```diff
@@ -50,7 +50,7 @@ visibility = "0.1.1"
 
 # Circuit
 halo2_gadgets = { version = "0.5", optional = true, default-features = false }
-halo2_proofs = { version = "0.3", optional = true, default-features = false, features = ["batch", "floor-planner-v1-legacy-pdqsort"] }
+halo2_proofs = { version = "0.3.5", optional = true, default-features = false, features = ["batch", "floor-planner-v1-legacy-pdqsort"] }
```
ebfull, line 53: We should have bumped this in the other release, technically.

PR #544, opened 2026-08-26
Pass the consumed proof bytes to the fingerprint fixture exporters

TalDerei:
Related to https://github.com/zcash/halo2/pull/933 and consumed in https://github.com/zcash/ironwood/pull/215.

This passes those proof bytes from Orchard’s capture drivers into the new halo2 exporters and negative test coverage so there's parity with the coverage Lean claims Orchard has. Still needs to be tagged against https://github.com/zcash/halo2/pull/933.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,4 +115,4 @@ debug = true
 
 # Temporary until a halo2_proofs release includes zcash/halo2#933.
 [patch.crates-io]
-halo2_proofs = { git = "https://github.com/TalDerei/halo2.git", rev = "74a5c1fab777a59ab16bcce8cf42714212a6ecac" }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -112,3 +112,7 @@ debug = true
 
 [profile.bench]
 debug = true
+
+# Temporary until a halo2_proofs release includes zcash/halo2#933.
+[patch.crates-io]
+halo2_proofs = { git = "https://github.com/TalDerei/halo2.git", rev = "d9a8c7374192d299d11366ed159bc1b3d1a38902" }
```
TalDerei: temporary patch, still needs to be tagged against https://github.com/zcash/halo2/pull/933.

PR #545, opened 2026-08-26
Add ActionBytes and BundleBytes types which skip curve artihmetic for SerDe

elicbarbieri:
# A parse tier for Action descriptions

Adds `ActionBytes` / `BundleBytes`, holding an Action description with `cv_net`, `rk` and `epk` left
compressed. A node reading actions from the wire or from disk — and a wallet scanning them — does no
curve arithmetic until it calls `decompress`.

```rust
let bundle: BundleBytes<Authorized, Amount> = read_bundle(&mut reader)?;
let txid_part = bundle.commitment(TxVersion::V5)?;
let notes = bundle.decrypt_outputs_with_keys(&ivks);

let bundle: Bundle<Authorized, Amount> = bundle.decompress()?;  // 3 sqrt/action, once
bundle.verify_proof(&vk)?;
```

`Action` and `Bundle` are untouched and cannot represent an action that breaks a point rule:
`ActionBytes::decompress` recovers `cv_net` and `rk` and then delegates to `Action::from_parts`, so
the identity-`rk` and `epk` rules keep exactly one implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,30 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
-### Changed
-- MSRV is now 1.88
+### Added
+- A parse tier holding an Action description with `cv_net`, `rk` and `epk` left
+  compressed, so reading actions from the wire or from disk costs no curve
+  arithmetic:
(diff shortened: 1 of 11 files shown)


elicbarbieri:
Im taking a human-review pass over everything.  Will have the commit up shortly w/ a cleanup to the doc style

PR #550, opened 2026-09-02
Update MSRV to 1.88

dannywillems:


.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest, macOS-latest]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
```
nuttycom, line 16: We should follow up by adding a zizmor CI gate, and hash-pinning these revisions, as we do in the other repositories.

dannywillems: Agent spawned.

nuttycom (approved):
utACK fa028506bdd25f46c1d1e28d968d3b9f373726c4

PR #551, opened 2026-09-02
Setup Zizmor

dannywillems:
Follow up from https://github.com/zcash/orchard/pull/550#discussion_r3917172390


dannywillems (commented):
Self utACK f7e36021854a2caa8e6579a800bf174b65472a95

nuttycom (approved):

PR #560, opened 2026-09-18
pczt: add trial decryption and output recovery to Action

### LINKED ISSUE #559: pczt::Action could offer its own CompactAction view and OVK recovery
## Context

I maintain a host-side engine (not yet public) that streams a PCZT to a Ledger hardware wallet for signing. Before the first byte goes out, the host has to classify every Orchard action's output exactly the way the device will: trial-decrypt the compact note ciphertext under the account's internal IVK (that output is change, and the device keeps it off the review screen), and otherwise recover the output under the account's external OVK (the device displays it, and the recovered memo counts against its retention budget).

Doing that from outside `orchard` means reimplementing two things:

1. **Building a `CompactAction` from an `orchard::pczt::Action`** — every field comes from the action itself, so this is pure plumbing a caller shouldn't have to write.
2. **Choosing the domain from the action's note version** — `NoteVersion::V2` takes the Orchard domain and `NoteVersion::V3` the Ironwood one. That pairing is `orchard`'s knowledge, not a caller's.

The second is the one that worries me. A host that gets the version-to-domain pairing wrong does not fail loudly: trial decryption simply does not match, and the output is then misclassified — in this application, that is the difference between an output the hardware wallet displays for approval and one it treats as change. Every downstream that needs this will write the same code, and any of them can get it silently wrong.

## Request

Something on `orchard::pczt::Action` that keeps the pairing inside `orchard`, for example:

- a `CompactAction` view of the action (or a trial decryption on the action itself), and
- an output recovery under an outgoing viewing key that selects the domain from the action's own note version.

Happy to open a PR if you'd like it in a particular shape.

_Prepared with Claude Code assistance._

nullcopy:
Closes #559.

A PCZT consumer that classifies an Orchard action's output has to pair the action's note
plaintext version with the matching note encryption domain: `NoteVersion::V2` with
`OrchardDomain`, `NoteVersion::V3` with `IronwoodDomain`. That pairing is this crate's
knowledge, and a consumer that reproduces it wrongly does not fail loudly — trial
decryption simply does not match, and the output is then misclassified. This keeps the
pairing inside the crate.

## Added

All on `orchard::pczt::Action`, in a new private `pczt::note_encryption` module that holds
nothing but these impls. Each method matches exhaustively on the output's own
`note_version` to select the domain; no existing item changes behaviour, and nothing's
visibility is widened.

```rust
impl From<&pczt::Action> for CompactAction

impl pczt::Action {
    pub fn decrypt_output_with_ivk(&self, ivk: &PreparedIncomingViewingKey)
        -> Option<(Note, Address, [u8; 512])>;

    pub fn decrypt_compact_output_with_ivk(&self, ivk: &PreparedIncomingViewingKey)
        -> Option<(Note, Address)>;

    pub fn recover_output_with_ovk(&self, ovk: &OutgoingViewingKey)
        -> Option<(Note, Address, [u8; 512])>;
}
```

## Notes on the shape, in case you'd prefer a different one

- **Where it lives.** The domain needs the spend's nullifier for `rho`, the output's
  ciphertexts and `cmx`, and — for OVK recovery — the action's `cv_net`, so `Action` is the
  only type that has everything. The new module is private and declared alongside the
  other role modules in `src/pczt.rs`; the methods surface on `pczt::Action` itself.

- **Prepared IVK.** `Bundle::decrypt_output_with_key` takes an `IncomingViewingKey` and
  prepares it internally because it owns the loop over actions. A per-action method does
  not own that loop, so it takes a `PreparedIncomingViewingKey` and lets the caller hoist
  the preparation across the bundle. That is also why these are named `_with_ivk` rather
  than `_with_key`, to keep the difference visible at the call site.

- **The compact variant** is there because that is what a consumer which only ever sees the
  compact note ciphertext does — a light client, or a hardware wallet that is streamed one
  — and it is not equivalent to the full decryption: it is not authenticated by the AEAD
  tag, and is validated by re-deriving `cmx` instead.

- **The `CompactAction` conversion** follows the existing `impl<T> From<&Action<T>> for
  CompactAction`, and is built through the public `CompactAction::from_parts`, so
  `note_encryption.rs` is untouched.

- The docs point callers at `Output::verify_note_commitment` for the other half of the
  check: a successful decryption binds the recovered note to the action's `cmx`, but says
  nothing about the action's separately-transmitted (and redactable) `recipient` and
  `value` fields.

## Tests

Three tests in the new module, which need no proving key and so are gated on `cfg(test)`
alone rather than the `circuit` feature:

- `compact_action_view_matches_the_action` — the compact view is assembled from the
  action's own fields.
- `decryption_domain_follows_the_output_note_version` — over both an `orchard_v2` and an
  `ironwood_v3` single-output bundle: all three entry points open the output and agree on
  the note, address and memo; another account's keys open nothing; and, the load-bearing
  assertion, flipping the output's `note_version` makes every one of them fail against the
  very keys the output was encrypted to.
- `output_without_an_ovk_is_not_recoverable` — an output added under an OVK policy of
  "None" is still readable by its recipient and recoverable by nobody.

Verified locally on the 1.88 toolchain: `cargo fmt --all -- --check`, `cargo clippy
--all-features --all-targets -- -D warnings`, `cargo test --all-features`, `cargo build
--no-default-features`, and `cargo doc --all-features --document-private-items`.

https://claude.ai/code/session_017y3oJoP92CcwnRyuThLT4b


src/pczt/note_encryption.rs
```diff
@@ -0,0 +1,305 @@
...
+        let out_ciphertext = &self.output.encrypted_note.out_ciphertext;
+        match self.output.note_version {
+            NoteVersion::V2 => try_output_recovery_with_ovk(
+                &OrchardDomain::for_pczt_action(self),
+                ovk,
+                self,
+                &self.cv_net,
+                out_ciphertext,
+            ),
+            NoteVersion::V3 => try_output_recovery_with_ovk(
```
nuttycom: NoteVersion::V3 only *currently* implies Ironwood. This is very purposefully handled in the Orchard crate: the dependency goes the other way; `BundleVersion::note_version` gives the note version for a given bundle version, but to determine what domain to use you need both the note version *and* the value pool.

nuttycom (changes requested):
This needs to be made safe against future changes where the note version doesn't uniquely identify the domain.

src/pczt/note_encryption.rs
nullcopy: Reworked so the domain is never selected from a note version. All three helpers now take the `BundleVersion` of the bundle the action belongs to, and a single private `ActionDomain::for_action` maps `(value_pool, protocol_version)` to `OrchardDomain`/`IronwoodDomain`. Both enums are spelled out there with no wildcard arm, so adding a value pool or a protocol version is a compile error in that one place rather than a silent reuse of one of today's domains (checked both ways by temporarily adding a variant to each).

The action's own `note_version` is no longer an input at all. It restates `BundleVersion::note_version`, which `Bundle::parse` already rejects a bundle for disagreeing with, and declining to decrypt on a disagreement would turn a tampered field into exactly the silent misclassification these helpers exist to prevent; `Output::verify_note_commitment` binds it to the `cmx` and fails loudly instead. That choice is documented on the methods.

`decryption_domain_follows_the_output_note_version` is replaced by `decryption_domain_follows_the_bundle_value_pool`: for both of today's pools it asserts that every protocol version instantiated for the output's own pool opens it, that the other pool's `BundleVersion` opens nothing, and that rewriting the action's `note_version` moves neither result. Branch squashed to one commit.

src/pczt/note_encryption.rs
```diff
@@ -0,0 +1,407 @@
...
+                ProtocolVersion::InsecureV1 | ProtocolVersion::V2 | ProtocolVersion::V3,
+            ) => Self::Orchard(OrchardDomain::for_pczt_action(action)),
+            // The Ironwood pool uses `IronwoodDomain`. It is only defined from
+            // `ProtocolVersion::V3` onward, and `BundleVersion` has no constructor for an
+            // earlier Ironwood version, so the earlier protocol versions are unconstructable;
+            // they are named rather than wildcarded to keep the exhaustiveness check above,
+            // and for them the pool alone already settles the answer.
+            (
+                ValuePool::Ironwood,
+                ProtocolVersion::InsecureV1 | ProtocolVersion::V2 | ProtocolVersion::V3,
```
nullcopy: Why are v1 & v2 matched here, when Ironwood only supports v3?

nullcopy: Only to keep the match exhaustive without a wildcard arm, so that adding a `ValuePool` or `ProtocolVersion` variant is a compile error here rather than a silent reuse of an existing domain. The cost was that the unconstructable `(Ironwood, InsecureV1 | V2)` pairings got handed `IronwoodDomain`.

Those arms are now the error arm. `ActionDomain::for_action` returns `Result<Self, UnsupportedBundleVersion>`, and the three helpers return `Result<Option<_>, UnsupportedBundleVersion>`. `Ok(None)` keeps its meaning — "this output did not decrypt under this key" — and the error means this crate defines no note encryption domain for that bundle version, so nothing was attempted and nothing was learned; they are deliberately not collapsed, since a consumer reading the second as the first is the misclassification the module exists to prevent.

The fallible arm and the exhaustive match are both worth keeping, because they catch different things. The match only fails to compile when a *variant* is added to either enum, and that is not how a new pairing is most likely to arrive: `BundleVersion` is `#[non_exhaustive]` with private fields, so its values come from its constructors, and a fifth constructor pairing variants that already exist changes no enum and breaks no match. No constructor produces such a version today, so the error is unreachable through the public API as it stands; since its input cannot be built, the tests pin it the other way round, asserting that a domain is defined for every version the crate can construct.

CHANGELOG.md
```diff
@@ -7,6 +7,35 @@ and this project adheres to Rust's notion of
...
+  domain from that version's `ValuePool` and `ProtocolVersion` together — never from a note
+  plaintext version, which is a consequence of the pool rather than an identifier of it. The
+  action's own `note_version` is not an input: it restates `BundleVersion::note_version`,
+  which `orchard::pczt::Bundle::parse` already enforces, and
+  `orchard::pczt::Output::verify_note_commitment` binds it to the output's `cmx`.
+  Each returns `Result<Option<_>, orchard::pczt::UnsupportedBundleVersion>`, keeping the two
+  negative outcomes apart: `Ok(None)` is the ordinary result of trial decryption not matching
+  ("not yours"), while the error means this crate defines no note encryption domain for the
+  bundle version supplied, so nothing was attempted and nothing was learned. A consumer that
+  read the latter as the former would misclassify every output of such a bundle.
```
nuttycom: This is way too much exposition for an `Added` section - additions don't need rationale and shouldn't narrate the progression that led to the current design.

nuttycom (dismissed):
utACK 7c7c16cff75cbeb8c7a0a7f53f494112f0b5837b

CHANGELOG.md
nullcopy: Trimmed to the symbol list.

nuttycom (approved):
utACK 68ceabe6eadd9b653bd685ee3cde4dc3600ee951

